PR #5, opened 2018-11-27
Documentation changes

boikonur:


boikonur:
I have renamed the licence file, so that will be displayed in the github interface, also added readme.md that can be used for landing page for all of the links there.

boikonur:
Yeah, for the licence file can be also used txt extention, as there is no markdown inside it, and Github will treat it as licence file in that case as well

PR #7, opened 2019-03-06
ENABLE_POWER_FOR_ID support

adays1234:
uses a #define ENABLE_POWER_FOR_ID PowerPINS<pin1,pin2,...> supported config file to enable the power on the pins prior to checking the resistance for the blade ID.  Allows support for pogo-pin type connectors for blade IDs where the blade ID resistor is placed on the LED negative as opposed to GND.

### DIFF
--- a/lightsaber.ino
+++ b/lightsaber.ino
@@ -949,13 +949,10 @@ public:
   // Measure and return the blade identifier resistor.
   float id() {
 #ifdef ENABLE_POWER_FOR_ID
+    ENABLE_POWER_FOR_ID power_pins_to_toggle;
     STDOUT.println("Power for ID enabled. Turning on FETs");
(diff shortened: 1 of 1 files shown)


profezzorn:
Merging offline.

adays1234:
Really wish you'd NOT merge these offline. This totally messes up folks' ability to refresh merges from the parent fork. This change basically collided and duplicated in my fork and I had to fix the merge errors.

:|

profezzorn:
I did try to merge it online, but unfortunately it messed up my workflow.
(CVS->github mirroring stopped working.)
I will put it on my TODO list to stop using CVS, but I will need to refactor part of the release process to do that, so it's not quite as simple as just not doing it anymore I'm afraid.

PR #10, opened 2019-04-01
Initial lock transition support (Monophonic endlock only)

adays1234:
SB_EndLockup change to detect whether endlock exists and use that instead of clash as a transition sound back to hum.

Additional changes for polyphonic lockon / lockoff support and monophonic bgnlock to come later.


profezzorn:
LGTM, but I will merge offline.
Merge should be 100% identical, which should hopefully let automatic merge tools work properly.

adays1234:
Sounds good. Just probably will have a fork merge from the parent again after you do. But I can resolve.

adays1234:
Oh did you figure out merging?

PR #13, opened 2019-04-16
Sa 22 c smoothswing only

SA-22C:
Tried to fix formatting, can't get smooth_swing_v2.h to format properly.


lightsaber.ino
```diff
@@ -539,6 +539,7 @@ public:
     CONFIG_VARIABLE(Transition1Degrees, 45.0f);
     CONFIG_VARIABLE(Transition2Degrees, 160.0f);
     CONFIG_VARIABLE(MaxSwingVolume, 3.0f);
+    CONFIG_VARIABLE(AccentSwingThreshold, 450.0f);
```
profezzorn: AccentSwingThreshold -> AccentSwingSpeedThreshold?

sound/smooth_swing_v2.h
```diff
@@ -7,220 +7,243 @@
...
+        void rotate(float degrees) {
+            midpoint += degrees;
+        }
+        RefPtr<BufferedWavPlayer> player;
+        float midpoint = 0.0;
+        float width = 0.0;
+    };
+    Data A;
+    Data B;
+    BufferedWavPlayer D;
```
profezzorn: We don't want a whole new bufferedwavplayer here.
Besides, it won't do anything unless hooked up to the dynamic mixer.

We just need to do what HybridFont::PlayPolyphonic does.
Alternatively, we could just allow swing events go down to the underlying font, and let it make up it's own mind of when to play swing sounds.

profezzorn:
This doesn't seem to be going anywhere, so closing it.
Feel free to re-open or create new ones as needed.

PR #14, opened 2019-04-18
Dimitry Shtoks One Button Mod

boikonur:
Here is the 1 button usage that the community uses in the group. It will be handy to have it in the baseline.
At some point will try to rearrange the macros in more macro "friendly" way

### DIFF
--- a/lightsaber.ino
+++ b/lightsaber.ino
@@ -1329,7 +1329,11 @@ public:
         break;
 
 
+#if NUM_BUTTONS == 1
+      case EVENTID(BUTTON_POWER, EVENT_CLICK_LONG, MODE_OFF):
+#else
       case EVENTID(BUTTON_AUX, EVENT_CLICK_SHORT, MODE_OFF):
+#endif
 #ifdef DUAL_POWER_BUTTONS
         aux_on_ = true;
         On();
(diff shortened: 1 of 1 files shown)


lightsaber.ino
```diff
@@ -1329,7 +1329,11 @@ public:
         break;
 
 
+#if NUM_BUTTONS == 1
```
profezzorn: I don't think this is so much better that nobody will complain if we just change it. So I think it needs to be controlled by a different define than NUM_BUTTONS. Also, it's difficult to follow exactly what this does, so it needs to be documented somewhere. A comment in the code might be enough.

boikonur: I have other idea for approch. Not define the number of buttons but use something like BUTTON_PRESET 0 to N. And within this macro is stated the actual number of buttons, events that are used and etc. That way, the code can have preset for plecter buttons, Dimitrys one button mode and many more. Of cource every preset must be documented in a good and user freindly way what will show what actualy uses and supports as events.

profezzorn: maybe we should do it like we do with the config file:

#define BUTTON_PRESET "button/some_button_configuration.h"

And in lightsaber.ino, do something like:

#include BUTTON_PRESET

However, each button preset file may need additional defines, such as NUM_BUTTONS

boikonur: So if we only have #define  BUTTON_PRESET BUTTON_PRESET_NAME in the config name, this must set NUM_BUTTONS, and will actually Instantiate the button objects in a separate header, which is included in the main depending on the first macro. That way, one simple define in the config file opens a huge variety. Additionally separating the configs in different headers allows better readability of every single configuration itself

profezzorn: something kind of like BUTTON_PRESET has now been implemented, see config/td_pb1_config.h for an example

PR #16, opened 2019-04-24
Smooth Swing Accent changes

SA-22C-smoothswing:
I couldn't get a clean pull from my current fork, so made a new one.  white space is still a mess on smoothswingv2.h but it's better than I was able to get on my original attempt.

Changes are, in brief

- added two new variables for smoothswing.ini (threshold and volume)
- added one new Data object in smooth_swing_v2.h to play the accent swing.  I gave up trying to pass the sounds up to the hybrid font.  This way works well enough
- I check for the presence of swng files
- select random swng file
- set volume of swing to match config file
- play swing file

It all seems to work well, I've tested on a couple fonts with good results.

### DIFF
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -70,9 +70,7 @@ public:
   // This is really just for sound fonts.
   virtual void SetHumVolume(float volume) {}
   virtual void StartSwing() {}
-  virtual float SetSwingVolume(float swing_strength, float AccentSwingVolumeSharpness,
-  float MaxAccentSwingVolume, float MaxAccentSwingDucking, float mixhum) {}
-  virtual bool IsSwingPlaying() {}
+  virtual float SetSwingVolume(float swing_strength, float mixhum) {}
   
 #define SABERFUN(NAME, TYPED_ARGS, ARGS)                        \
 public:                                                         \
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -6,7 +6,6 @@
 // has a corresponding SaberBase::Do* function which invokes that function
 // on all active SaberBases.
 class SaberBase;
-class Effect;
 extern SaberBase* saberbases;
 
 class SaberBase {
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -69,7 +69,7 @@ public:
   // This is really just for sound fonts.
   virtual void SetHumVolume(float volume) {}
   virtual void StartSwing() {}
-  virtual float SetSwingVolume(float swing_strength, float mixhum) {}
+  virtual void SetSwingVolume(float swing_strength) {}
   
 #define SABERFUN(NAME, TYPED_ARGS, ARGS)                        \
 public:                                                         \
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -69,7 +69,7 @@ public:
   // 1.0 = kDefaultVolume
   // This is really just for sound fonts.
   virtual void SetHumVolume(float volume) {}
-  virtual void StartSwing(Effect* monophonic, Effect* polyphonic) {}
+  virtual void StartSwing() {}
   virtual float SetSwingVolume(float swing_strength, float AccentSwingVolumeSharpness,
   float MaxAccentSwingVolume, float MaxAccentSwingDucking, float mixhum) {}
   virtual bool IsSwingPlaying() {}
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -69,7 +69,7 @@ public:
   // This is really just for sound fonts.
   virtual void SetHumVolume(float volume) {}
   virtual void StartSwing() {}
-  virtual void SetSwingVolume(float swing_strength) {}
+  virtual float SetSwingVolume(float swing_strength, float mixhum) {}
   
 #define SABERFUN(NAME, TYPED_ARGS, ARGS)                        \
 public:                                                         \
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -68,7 +68,7 @@ public:
   // 1.0 = kDefaultVolume
   // This is really just for sound fonts.
   virtual void SetHumVolume(float volume) {}
-  virtual void StartSwing(bool AccentSwing) {}
+  virtual void StartSwing() {}
   virtual void SetSwingVolume(float swing_strength) {}
   
 #define SABERFUN(NAME, TYPED_ARGS, ARGS)                        \
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -68,7 +68,7 @@ public:
   // 1.0 = kDefaultVolume
   // This is really just for sound fonts.
   virtual void SetHumVolume(float volume) {}
-  virtual void StartSwing() {}
+  virtual void StartSwing(bool AccentSwing) {}
   virtual void SetSwingVolume(float swing_strength) {}
   
 #define SABERFUN(NAME, TYPED_ARGS, ARGS)                        \
--- a/lightsaber.ino
+++ b/lightsaber.ino
@@ -183,7 +183,7 @@ SnoozeTouch snooze_touch;
 SnoozeBlock snooze_config(snooze_touch, snooze_digital, snooze_timer);
 #endif
 
-const char version[] = "$Id$";
+const char version[] = "$Id: 0c3909b34a755d918ac8d7077ea1142902bbd0db $";
 
 #include "common/state_machine.h"
 #include "common/monitoring.h"
@@ -541,9 +541,6 @@ public:
     CONFIG_VARIABLE(Transition2Degrees, 160.0f);
     CONFIG_VARIABLE(MaxSwingVolume, 3.0f);
     CONFIG_VARIABLE(AccentSwingSpeedThreshold, 0.0f);
-    CONFIG_VARIABLE(AccentSwingVolumeSharpness, 1.0f);
-    CONFIG_VARIABLE(MaxAccentSwingVolume, 3.0f);
-    CONFIG_VARIABLE(MaxAccentSwingDucking, 0.5f);
   };
 
   int  Version;
@@ -555,9 +552,6 @@ public:
   float Transition2Degrees;
   float MaxSwingVolume;
   float AccentSwingSpeedThreshold;
-  float AccentSwingVolumeSharpness;
-  float MaxAccentSwingVolume;
-  float MaxAccentSwingDucking;
 };
 
 SmoothSwingConfigFile smooth_swing_config;
@@ -3235,4 +3229,3 @@ void loop() {
     last_activity = millis();
   }
 }
-
--- a/lightsaber.ino
+++ b/lightsaber.ino
@@ -183,7 +183,7 @@ SnoozeTouch snooze_touch;
 SnoozeBlock snooze_config(snooze_touch, snooze_digital, snooze_timer);
 #endif
 
-const char version[] = "$Id:$";
+const char version[] = "$Id$";
 
 #include "common/state_machine.h"
 #include "common/monitoring.h"
--- a/lightsaber.ino
+++ b/lightsaber.ino
@@ -183,7 +183,7 @@ SnoozeTouch snooze_touch;
 SnoozeBlock snooze_config(snooze_touch, snooze_digital, snooze_timer);
 #endif
 
-const char version[] = "$Id: 0c3909b34a755d918ac8d7077ea1142902bbd0db $";
+const char version[] = "$Id:$";
 
 #include "common/state_machine.h"
 #include "common/monitoring.h"
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -139,11 +139,8 @@ public:
         // avoid overlapping swings, based on value set in ProffieOSSwingOverlap.  Value is
         // between 0 (no overlap) and 1.0 (full overlap)
         if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
-          RefPtr<BufferedWavPlayer> overlap_swing = swing_player_;
+          swing_player_.Free();
           swing_player_ = PlayPolyphonic(&swng);
-          overlap_swing->set_fade_time(overlap_swing->length() - overlap_swing->pos());
-          overlap_swing->FadeAndStop();
-          overlap_swing.Free();
         }
       }
       else if (!swing_player_) {
@@ -160,7 +157,7 @@ public:
       if (swing_player_->isPlaying()) {
         float accent_volume = powf(swing_strength, config_.ProffieOSSwingVolumeSharpness) * config_.ProffieOSMaxSwingVolume;
         swing_player_->set_volume(accent_volume);
-        mixhum = mixhum - mixhum * (config_.ProffieOSSmoothSwingDucking * swing_strength);
+        mixhum = mixhum - mixhum * (config_.ProffieOSSmoothSwingDucking * accent_volume);
       }
       else {
         swing_player_.Free();
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -39,23 +39,23 @@ public:
     guess_monophonic_ = false;
     if (monophonic_hum_) {
       if ((clash.files_found() || blaster.files_found() || swing.files_found())) {
-	guess_monophonic_ = true;
-	STDOUT.print("monophonic");
+        guess_monophonic_ = true;
+        STDOUT.print("monophonic");
       } else {
-	guess_monophonic_ = false;
-	STDOUT.print("hybrid");
+        guess_monophonic_ = false;
+        STDOUT.print("hybrid");
       }
     } else {
       guess_monophonic_ = false;
       STDOUT.print("polyphonic");
     }
-	
+    
     STDOUT.println(" font.");
     SaberBase::Link(this);
     SetHumVolume(1.0);
     state_ = STATE_OFF;
   }
-
+  
   enum State {
     STATE_OFF,
     STATE_OUT,
@@ -63,7 +63,7 @@ public:
     STATE_HUM_ON,
     STATE_HUM_FADE_OUT,
   };
-
+  
   void Deactivate() {
     lock_player_.Free();
     hum_player_.Free();
@@ -72,7 +72,7 @@ public:
     second_swing_player_.Free();
     SaberBase::Unlink(this);
   }
-
+  
   RefPtr<BufferedWavPlayer> hum_player_;
   RefPtr<BufferedWavPlayer> next_hum_player_;
   RefPtr<BufferedWavPlayer> swing_player_;
@@ -84,8 +84,8 @@ public:
     if (!next_hum_player_) {
       next_hum_player_ = GetFreeWavPlayer();
       if (!next_hum_player_) {
-	STDOUT.println("Out of WAV players!");
-	return;
+        STDOUT.println("Out of WAV players!");
+        return;
       }
     }
     if (hum_player_) {
@@ -104,7 +104,7 @@ public:
     hum_player_->PlayOnce(f);
     if (loop) hum_player_->PlayLoop(loop);
   }
-							   
+  
   RefPtr<BufferedWavPlayer> PlayPolyphonic(Effect* f)  {
     EnableAmplifier();
     RefPtr<BufferedWavPlayer> player = GetFreeWavPlayer();
@@ -114,7 +114,7 @@ public:
     }
     return player;
   }
-
+  
   void Play(Effect* monophonic, Effect* polyphonic) {
     if (polyphonic->files_found()) {
       PlayPolyphonic(polyphonic);
@@ -122,7 +122,7 @@ public:
       PlayMonophonic(monophonic, &hum);
     }
   }
-
+  
   void PlayCommon(Effect* effect) {
     if (guess_monophonic_) {
       PlayMonophonic(effect, &hum);
@@ -136,44 +136,30 @@ public:
         swing_player_ = PlayPolyphonic(&swng);
       } else {
         if (!second_swing_player_) {
-          second_swing_player_ = PlayPolyphonic(&swng);
+          second_swing_player_ = swing_player_;
+          swing_player_ = PlayPolyphonic(&swng);
+          second_swing_player_->set_fade_time(0.2);
+          second_swing_player_->FadeAndStop();
+          second_swing_player_.Free();
         }
       }
     } else {
       PlayMonophonic(&swing, &hum);
     }
   }
-
+  
   float SetSwingVolume(float swing_strength, float mixhum) {
-    if (IsSwingPlaying()) {
-      float accent_volume = powf(swing_strength, config_.ProffieOSSwingVolumeSharpness) * config_.ProffieOSMaxSwingVolume;
-      swing_player_->set_volume(accent_volume);
-      if (second_swing_player_) {
-        second_swing_player_->set_volume(accent_volume);
+    if(swing_player_) {
+      if (swing_player_->isPlaying()) {
+        float accent_volume = powf(swing_strength, config_.ProffieOSSwingVolumeSharpness) * config_.ProffieOSMaxSwingVolume;
+        swing_player_->set_volume(accent_volume);
+        mixhum = mixhum * config_.ProffieOSMaxSwingDucking;
       }
-      mixhum = mixhum * config_.ProffieOSMaxSwingDucking;
+      else {
+        swing_player_.Free();
+      }
     }
     return mixhum;
-  }
-
-  bool IsSwingPlaying() {
-    if (swing_player_) {
-      if (second_swing_player_) {
-        if(second_swing_player_->isPlaying() || swing_player_->isPlaying()) {
-          return true;
-        }
-      } else {
-        if (!swing_player_->isPlaying()) {
-          swing_player_.Free();
-        }
-        if (!second_swing_player_->isPlaying()) {
-          second_swing_player_.Free();
-        }
-        return false;
-      }
-    } else {
-      return false;
-    }
   }
   void SB_On() override {
     if (monophonic_hum_) {
@@ -183,31 +169,31 @@ public:
       state_ = STATE_OUT;
       hum_player_ = GetFreeWavPlayer();
       if (hum_player_) {
-	hum_player_->set_volume_now(0);
-	hum_player_->PlayOnce(&hum);
-	hum_player_->PlayLoop(&hum);
-	hum_start_ = millis();
+        hum_player_->set_volume_now(0);
+        hum_player_->PlayOnce(&hum);
+        hum_player_->PlayLoop(&hum);
+        hum_start_ = millis();
       }
       RefPtr<BufferedWavPlayer> tmp = PlayPolyphonic(&out);
       if (config_.humStart && tmp) {
-	int delay_ms = 1000 * tmp->length() - config_.humStart;
-	if (delay_ms > 0 && delay_ms < 30000) {
-	  hum_start_ += delay_ms;
-	}
+        int delay_ms = 1000 * tmp->length() - config_.humStart;
+        if (delay_ms > 0 && delay_ms < 30000) {
+          hum_start_ += delay_ms;
+        }
       }
     }
   }
-
+  
   void SB_Off() override {
     if (monophonic_hum_) {
       size_t total = poweroff.files_found() + pwroff.files_found();
       if (total) {
-	state_ = STATE_OFF;
-	if ((rand() % total) < poweroff.files_found()) {
-	  PlayMonophonic(&poweroff, NULL);
-	} else {
-	  PlayMonophonic(&pwroff, NULL);
-	}
+        state_ = STATE_OFF;
+        if ((rand() % total) < poweroff.files_found()) {
+          PlayMonophonic(&poweroff, NULL);
+        } else {
+          PlayMonophonic(&pwroff, NULL);
+        }
       }
     } else {
       state_ = STATE_HUM_FADE_OUT;
@@ -220,24 +206,24 @@ public:
   void SB_Blast() override { Play(&blaster, &blst); }
   void SB_Boot() override { PlayPolyphonic(&boot); }
   void SB_NewFont() override { PlayPolyphonic(&font); }
-
+  
   void SB_BeginLockup() override {
     if (lockup.files_found()) {
       if (SaberBase::Lockup() == SaberBase::LOCKUP_DRAG &&
-	  drag.files_found()) {
-	PlayMonophonic(&drag, &drag);
+          drag.files_found()) {
+        PlayMonophonic(&drag, &drag);
       } else if (lockup.files_found()) {
         if (bgnlock.files_found()) {
           PlayMonophonic(&bgnlock, &lockup);
         } else {
           PlayMonophonic(&lockup, &lockup);
-	}
+        }
       }
     } else {
       Effect* e = &lock;
       if (SaberBase::Lockup() == SaberBase::LOCKUP_DRAG &&
-	  drag.files_found()) {
-	e = &drag;
+          drag.files_found()) {
+        e = &drag;
       }
       if (!lock_player_) {
         if (bgnlock.files_found()) {
@@ -245,26 +231,26 @@ public:
         } else {
           lock_player_ = PlayPolyphonic(e);
         }
-
-	if (lock_player_) {
-	  lock_player_->PlayLoop(e);
-	}
+        
+        if (lock_player_) {
+          lock_player_->PlayLoop(e);
+        }
       }
     }
   }
-
+  
   void SB_EndLockup() override {
     if (lock_player_) {
       // Polyphonic case
       lock_player_->set_fade_time(0.3);
-
+      
       if (endlock.files_found()) { // polyphonic end lock
         if (PlayPolyphonic(&endlock)) {
           // if playing an end lock fade the lockup faster
           lock_player_->set_fade_time(0.003);
-	}
+        }
       }
-
+      
       lock_player_->FadeAndStop();
       lock_player_.Free();
       return;
@@ -278,53 +264,53 @@ public:
       }
     }
   }
-
+  
   void SetHumVolume(float vol) override {
     if (!monophonic_hum_) {
       if (state_ != STATE_OFF && !hum_player_) {
-	hum_player_ = GetFreeWavPlayer();
-	if (hum_player_) {
-	  hum_player_->set_volume_now(0);
-	  hum_player_->PlayOnce(&hum);
-	  hum_player_->PlayLoop(&hum);
-	  hum_start_ = millis();
-	}
+        hum_player_ = GetFreeWavPlayer();
+        if (hum_player_) {
+          hum_player_->set_volume_now(0);
+          hum_player_->PlayOnce(&hum);
+          hum_player_->PlayLoop(&hum);
+          hum_start_ = millis();
+        }
       }
       if (!hum_player_) return;
       uint32_t m = micros();
       switch (state_) {
-	case STATE_OFF:
-	  volume_ = 0.0f;
-	  return;
-	case STATE_OUT:
-	  volume_ = 0.0f;
-	  if (millis() - hum_start_ < 0x7fffffffUL) {
-	    state_ = STATE_HUM_FADE_IN;
-	  }
-	  break;
-	case STATE_HUM_FADE_IN: {
-	  uint32_t delta = m - last_micros_;
-	  volume_ += (delta / 1000000.0) / 0.2; // 0.2 seconds
-	  if (volume_ >= 1.0f) {
-	    volume_ = 1.0f;
-	    state_ = STATE_HUM_ON;
-	  }
-	  break;
-	}
-	case STATE_HUM_ON:
-	  break;
-	case STATE_HUM_FADE_OUT: {
-	  SaberBase::RequestMotion();
-	  uint32_t delta = m - last_micros_;
-	  volume_ -= (delta / 1000000.0) / 0.2; // 0.2 seconds
-	  if (volume_ <= 0.0f) {
-	    volume_ = 0.0f;
-	    state_ = STATE_OFF;
-	    hum_player_->FadeAndStop();
-	    hum_player_.Free();
-	  }
-	  break;
-	}
+        case STATE_OFF:
+          volume_ = 0.0f;
+          return;
+        case STATE_OUT:
+          volume_ = 0.0f;
+          if (millis() - hum_start_ < 0x7fffffffUL) {
+            state_ = STATE_HUM_FADE_IN;
+          }
+          break;
+        case STATE_HUM_FADE_IN: {
+          uint32_t delta = m - last_micros_;
+          volume_ += (delta / 1000000.0) / 0.2; // 0.2 seconds
+          if (volume_ >= 1.0f) {
+            volume_ = 1.0f;
+            state_ = STATE_HUM_ON;
+          }
+          break;
+        }
+        case STATE_HUM_ON:
+          break;
+        case STATE_HUM_FADE_OUT: {
+          SaberBase::RequestMotion();
+          uint32_t delta = m - last_micros_;
+          volume_ -= (delta / 1000000.0) / 0.2; // 0.2 seconds
+          if (volume_ <= 0.0f) {
+            volume_ = 0.0f;
+            state_ = STATE_OFF;
+            hum_player_->FadeAndStop();
+            hum_player_.Free();
+          }
+          break;
+        }
       }
       last_micros_ = m;
       vol *= volume_;
@@ -338,7 +324,7 @@ public:
     float speed = sqrtf(gyro.z * gyro.z + gyro.y * gyro.y);
     if (speed > 250.0) {
       if (!swinging_ && state_ != STATE_OFF &&
-	  !(lockup.files_found() && SaberBase::Lockup())) {
+          !(lockup.files_found() && SaberBase::Lockup())) {
         swinging_ = true;
         StartSwing();
       }
@@ -351,8 +337,8 @@ public:
     }
     SetHumVolume(vol);
   }
-
- private:
+  
+private:
   uint32_t last_micros_;
   uint32_t hum_start_;
   bool monophonic_hum_;
@@ -363,3 +349,4 @@ public:
 };
 
 #endif
+
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -69,14 +69,12 @@ public:
     hum_player_.Free();
     next_hum_player_.Free();
     swing_player_.Free();
-    second_swing_player_.Free();
     SaberBase::Unlink(this);
   }
 
   RefPtr<BufferedWavPlayer> hum_player_;
   RefPtr<BufferedWavPlayer> next_hum_player_;
   RefPtr<BufferedWavPlayer> swing_player_;
-  RefPtr<BufferedWavPlayer> second_swing_player_;
   RefPtr<BufferedWavPlayer> lock_player_;
   
   void PlayMonophonic(Effect* f, Effect* loop)  {
@@ -135,13 +133,11 @@ public:
       if (!swing_player_) {
         swing_player_ = PlayPolyphonic(&swng);
       } else {
-        if (!second_swing_player_) {
-          second_swing_player_ = swing_player_;
-          swing_player_ = PlayPolyphonic(&swng);
-          second_swing_player_->set_fade_time(0.2);
-          second_swing_player_->FadeAndStop();
-          second_swing_player_.Free();
-        }
+        RefPtr<BufferedWavPlayer> second_swing_player_ = swing_player_;
+        swing_player_ = PlayPolyphonic(&swng);
+        second_swing_player_->set_fade_time(0.2);
+        second_swing_player_->FadeAndStop();
+        second_swing_player_.Free();
       }
     } else {
       PlayMonophonic(&swing, &hum);
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -343,7 +343,7 @@ public:
         StartSwing();
       }
       float swing_strength = std::min<float>(1.0, speed / config_.ProffieOSSwingSpeedThreshold);
-      SetSwingVolume(swing_strength);
+      SetSwingVolume(swing_strength, 1.0);
     } else if (swinging_ && speed <= config_.ProffieOSSwingSpeedThreshold * 0.8) {
       swinging_ = false;
     }
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -135,7 +135,9 @@ public:
       if (!swing_player_) {
         swing_player_ = PlayPolyphonic(&swng);
       } else {
-        second_swing_player_ = PlayPolyphonic(&swng);
+        if (!second_swing_player_) {
+          second_swing_player_ = PlayPolyphonic(&swng);
+        }
       }
     } else {
       PlayMonophonic(&swing, &hum);
@@ -161,10 +163,10 @@ public:
           return true;
         }
       } else {
-        if (!swing_player_) {
+        if (!swing_player_->isPlaying()) {
           swing_player_.Free();
         }
-        if (!second_swing_player_) {
+        if (!second_swing_player_->isPlaying()) {
           second_swing_player_.Free();
         }
         return false;
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -8,9 +8,9 @@ public:
     CONFIG_VARIABLE(volHum, 15);
     CONFIG_VARIABLE(volEff, 16);
     CONFIG_VARIABLE(ProffieOSSwingSpeedThreshold, 250.0f);
-    CONFIG_VARIABLE(ProffieOSSwingVolumeSharpness, 0.5f);
+    CONFIG_VARIABLE(ProffieOSSwingVolumeSharpness, 1.0f);
     CONFIG_VARIABLE(ProffieOSMaxSwingVolume, 3.0f);
-    CONFIG_VARIABLE(ProffieOSSwingOverlap, 0.6f);
+    CONFIG_VARIABLE(ProffieOSSwingOverlap, 0.5f);
     CONFIG_VARIABLE(ProffieOSSmoothSwingDucking, 0.25f);
   }
   int humStart;
@@ -137,7 +137,7 @@ public:
     if (!guess_monophonic_) {
       if (swing_player_) {
         // avoid overlapping swings, based on value set in ProffieOSSwingOverlap.  Value is
-        // between 0 (no overlap) and 1.0 (full overlap)
+        // between 0 (full overlap) and 1.0 (no overlap)
         if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
           swing_player_.Free();
           swing_player_ = PlayPolyphonic(&swng);
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -129,24 +129,21 @@ public:
     }
   }
   
-  void StartSwing(bool AccentSwing) override {
-      if (AccentSwing) {
-        if (!guess_monophonic_) {
-          if (!swing_player_) {
-            swing_player_ = PlayPolyphonic(&swng);
-          }
-          if (swing_player_->length() - swing_player_->pos() <= 0.1) {
-            RefPtr<BufferedWavPlayer> overlap_swing = swing_player_;
-            overlap_swing->set_fade_time(0.1);
-            overlap_swing->FadeAndStop();
-            overlap_swing.Free();
-            swing_player_ = PlayPolyphonic(&swng);
-          }
-        } else {
-          PlayMonophonic(&swing, &hum);
-        }
-      } else {
-        Play(&swing, &swng);
+  void StartSwing() override {
+    if (!guess_monophonic_) {
+      if (!swing_player_) {
+        swing_player_ = PlayPolyphonic(&swng);
+      }
+      // avoid overlapping swings unless at least a quarter of the swing has played.  Stops double swing events.
+      if ((swing_player_->length() - swing_player_->pos()) / swing_player_->length() <= 0.75) {
+        RefPtr<BufferedWavPlayer> overlap_swing = swing_player_;
+        swing_player_ = PlayPolyphonic(&swng);
+        overlap_swing->set_fade_time(0.1);
+        overlap_swing->FadeAndStop();
+        overlap_swing.Free();
+      }
+    } else {
+      PlayMonophonic(&swing, &hum);
     }
   }
 
@@ -329,6 +326,8 @@ public:
         swinging_ = true;
         StartSwing();
       }
+      float swing_strength = std::min<float>(1.0, speed / config_.ProffieOSSwingSpeedThreshold);
+      SetSwingVolume(swing_strength);
     } else {
       swinging_ = false;
     }
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -96,7 +96,7 @@ public:
     hum_player_->PlayOnce(f);
     if (loop) hum_player_->PlayLoop(loop);
   }
-					  		 
+				  	  		 
   RefPtr<BufferedWavPlayer> PlayPolyphonic(Effect* f)  {
     EnableAmplifier();
     RefPtr<BufferedWavPlayer> player = GetFreeWavPlayer();
@@ -208,7 +208,7 @@ public:
         PlayMonophonic(&drag, &drag);
       } else if (lockup.files_found()) {
         if (bgnlock.files_found()) {
-         PlayMonophonic(&bgnlock, &lockup);
+          PlayMonophonic(&bgnlock, &lockup);
         } else {
           PlayMonophonic(&lockup, &lockup);
       }
@@ -236,13 +236,14 @@ public:
     if (lock_player_) {
       // Polyphonic case
       lock_player_->set_fade_time(0.3);
-      
+
       if (endlock.files_found()) { // polyphonic end lock
         if (PlayPolyphonic(&endlock)) {
           // if playing an end lock fade the lockup faster
           lock_player_->set_fade_time(0.003);
-	}
+	      }
       }
+      
       lock_player_->FadeAndStop();
       lock_player_.Free();
       return;
@@ -257,16 +258,16 @@ public:
       PlayMonophonic(&clash, &hum);
     }
   }
-  
+
   void SetHumVolume(float vol) override {
     if (!monophonic_hum_) {
       if (state_ != STATE_OFF && !hum_player_) {
-        hum_player_ = GetFreeWavPlayer();
-        if (hum_player_) {
-	 hum_player_->set_volume_now(0);
-	 hum_player_->PlayOnce(&hum);
-	 hum_player_->PlayLoop(&hum);
-	 hum_start_ = millis();
+       hum_player_ = GetFreeWavPlayer();
+       if (hum_player_) {
+	      hum_player_->set_volume_now(0);
+	      hum_player_->PlayOnce(&hum);
+	      hum_player_->PlayLoop(&hum);
+	      hum_start_ = millis();
         }
       }
       if (!hum_player_) return;
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -159,7 +159,7 @@ public:
     if(swing_player_) {
       if (swing_player_->isPlaying()) {
         float accent_volume = powf(swing_strength, config_.ProffieOSSwingVolumeSharpness) * config_.ProffieOSMaxSwingVolume;
-        swing_player_->set_fade_time(0.02);
+        swing_player_->set_fade_time(0.04);
         swing_player_->set_volume(accent_volume);
         mixhum = mixhum - mixhum * (config_.ProffieOSSmoothSwingDucking * accent_volume);
       } else {
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -144,7 +144,9 @@ public:
           overlap_swing.Free();
         }
       }
-      swing_player_ = PlayPolyphonic(&swng);
+      else if (!swing_player_) {
+        swing_player_ = PlayPolyphonic(&swng);
+      }
     } else {
       PlayMonophonic(&swing, &hum);
     }
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -124,7 +124,7 @@ public:
   }
   void StartSwing(Effect* monophonic, Effect* polyphonic) override {
     if (polyphonic->files_found()) {
-        swing_player_ = PlayPolyphonic(polyphonic);
+      swing_player_ = PlayPolyphonic(polyphonic);
     } else {
       PlayMonophonic(monophonic, &hum);
     }
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -220,16 +220,14 @@ public:
       }
       if (!lock_player_) {
         if (bgnlock.files_found()) {
-        lock_player_ = PlayPolyphonic(&bgnlock);
-      } else {
+          lock_player_ = PlayPolyphonic(&bgnlock);
+        } else {
           lock_player_ = PlayPolyphonic(e);
         }
-      
-      if (lock_player_) {
-       if (bgnlock.files_found()) {
-        lock_player_ = PlayPolyphonic(&bgnlock);
-        
-        lock_player_->PlayLoop(e);
+	      
+	if (lock_player_) {
+	  lock_player_->PlayLoop(e);
+	}
       }
     }
   }
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -146,6 +146,7 @@ public:
       }
       else if (!swing_player_) {
         swing_player_ = PlayPolyphonic(&swng);
+        swing_player_->set_volume_now(0);
       }
     } else {
       PlayMonophonic(&swing, &hum);
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -124,9 +124,9 @@ public:
   }
   void StartSwing() override {
     if (!guess_monophonic_) {
-      swing_player_ = PlayPolyphonic(swng);
+      swing_player_ = PlayPolyphonic(&swng);
     } else {
-      PlayMonophonic(swing, &hum);
+      PlayMonophonic(&swing, &hum);
     }
   }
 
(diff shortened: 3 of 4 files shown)


sound/smooth_swing_v2.h
```diff
@@ -8,36 +8,53 @@
...
+  
   void Activate(SaberBase* base_font) {
     STDOUT.println("Activating SmoothSwing V2");
     SetDelegate(base_font);
     if (swingl.files_found() != swingh.files_found()) {
       STDOUT.println("Warning, swingl and swingh should have the same number of files.");
     }
     swings_ = std::min<size_t>(swingl.files_found(), swingh.files_found());
+    if (swng.files_found() > 0)
+    {
```
profezzorn: Please put the { at the end of the previous line, like the rest of this code does.

sound/smooth_swing_v2.h
```diff
@@ -8,36 +8,53 @@
...
+      aswings_ = size_t(swng.files_found());
```
profezzorn: the size_t cast is not needed.

sound/smooth_swing_v2.h
```diff
@@ -214,13 +260,21 @@ class SmoothSwingV2 : public SaberBasePassThrough {
   };
   Data A;
   Data B;
-
+  Data D;
```
profezzorn: We should not use a 'Data' here.
Instead we should use something like:

 RefPtr < BufferedWavPlayer > accent_player_;

instead.

sound/smooth_swing_v2.h
```diff
@@ -8,36 +8,53 @@
...
+  
   // Should only be done when the volume is near zero.
   void PickRandomSwing() {
     if (!on_) return;
     uint32_t m = millis();
     // No point in picking a new random so soon after picking one.
     if (A.player && m - last_random_ < 1000) return;
     last_random_ = m;
-    int swing = random(swings_);
+    while (swing == randomizer)
```
profezzorn: This will hang forever if there is only one swing sound.
Also, we shouldn't pick the sound, we should just play the effect and let it select which one to play.

sound/smooth_swing_v2.h
```diff
@@ -94,94 +111,123 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+           STDOUT.println(speed);
           }
 #endif
           break;
         }
         state_ = SwingState::ON;
         
       case SwingState::ON:
+        //check for AccentSwingThreshold, presence of accent swings and if the accent player is stopped (this prevents clipping)
+        if (speed >=smooth_swing_config.AccentSwingSpeedThreshold && accent_swings_present && !D.isPlaying() && (A.isPlaying() || B.isPlaying()))
```
profezzorn: I don't think we need to check if A or B is playing since we're in SwingState::ON

sound/smooth_swing_v2.h
```diff
@@ -94,94 +111,123 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
       case SwingState::OUT:
         if (!A.isOff() || !B.isOff()) {
           if (monitor.ShouldPrint(Monitoring::MonitorSwings)) {
-            Serial.println("Waiting for volume = 0");
+           Serial.println("Waiting for volume = 0");
           }
         }
         PickRandomSwing();
         state_ = SwingState::OFF;
+        accent_played = false;
```
profezzorn: This is set here, but never used?

sound/smooth_swing_v2.h
```diff
@@ -94,94 +111,123 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
       case SwingState::ON:
+        //check for AccentSwingThreshold, presence of accent swings and if the accent player is stopped (this prevents clipping)
+        if (speed >=smooth_swing_config.AccentSwingSpeedThreshold && accent_swings_present && !D.isPlaying() && (A.isPlaying() || B.isPlaying()))
+        {
+          while (accentswing == accentrandomizer)
+          {
+           accentswing = random(aswings_);
+          }
+          accentrandomizer = accentswing;
+          STDOUT.print("Accent SWING!");
```
profezzorn: Please remove.

sound/smooth_swing_v2.h
```diff
@@ -94,94 +111,123 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+          D.set_volume(smooth_swing_config.AccentSwingVolume);
```
profezzorn: I think we need to make the volume of the accent dynamic, and change as the swing progresses.
If the swing suddenly stops, the volume of the accent swing should also go down to zero I think.

sound/smooth_swing_v2.h
SA-22C-smoothswing: I thought so to, but I had accent swings playing while the saber was off until I added that check...

sound/smooth_swing_v2.h
SA-22C-smoothswing: right, one swing would cause an issue there.  What I was trying to do was prevent the same swing being played twice.

sound/smooth_swing_v2.h
SA-22C-smoothswing: That's a good idea, not quite sure right now how to accomplish that but I'll dig in.

sound/smooth_swing_v2.h
SA-22C-smoothswing: leftover from old version, should be removed.

sound/smooth_swing_v2.h
SA-22C-smoothswing: Done.

sound/smooth_swing_v2.h
SA-22C-smoothswing: I'm going to leave this check in for safety.  I don't think it's hurting anything and I don't get the accent swings when the saber is off anymore.

sound/smooth_swing_v2.h
SA-22C-smoothswing: Done.  changed up the config variables.
AccentSwingSpeedThreshold - same, deg/sec needed to trigger accent
AccentSwingVolumeSharpness - new, same idea as SwingSharpness.  Used to govern the fade in/out of the swing sound
MaxAccentSwingVolume - renamed from AccentSwingVolume, combines with AccentSwingVolumeSharpness to govern accent swing volume and fade in/out

SA-22C-smoothswing:
ok, I've made all the requested changes and it works really well.  Hopefully these changes meet with your approval as well.  Thanks.

lightsaber.ino
```diff
@@ -183,7 +183,7 @@ SnoozeTouch snooze_touch;
 SnoozeBlock snooze_config(snooze_touch, snooze_digital, snooze_timer);
 #endif
 
-const char version[] = "$Id$";
+const char version[] = "$Id: 3210a1ba122c79758d6fd90b84af5f57d46b1d2b $";
```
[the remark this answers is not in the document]
boikonur: If regular git commiting is used, it wont generate that Id

adays1234:
I notice a few of us are running into whitespace changes.

Clearly you are using whitespace as spaces and 2-char indent levels. Are there other things you are doing for a formatting? Are you using an IDE with a formatter / code cleanup type template? If so, can you post that Prof? Maybe add it to the repo? That way we can all be using the same formatter if we are using IDEs

profezzorn:
> I notice a few of us are running into whitespace changes.
> 
> Clearly you are using whitespace as spaces and 2-char indent levels. Are there other things you are doing for a formatting? Are you using an IDE with a formatter / code cleanup type template? If so, can you post that Prof? Maybe add it to the repo? That way we can all be using the same formatter if we are using IDEs

I'm not using any form of IDE. I use emacs, which does some code formatting, but mostly I do it by hand.
PS: I don't recommend emacs to anyone, it's extremely user hostile.
I roughly follow the google style guide for c++: https://google.github.io/styleguide/cppguide.html

profezzorn:
I thought about this pull request last night, and I'm wondering if it would be better to let the underlying (wrapped) font play the swing sounds. This has the advantage that it would also work for plecter style swing sounds.

adays1234:
> > I notice a few of us are running into whitespace changes.
> > Clearly you are using whitespace as spaces and 2-char indent levels. Are there other things you are doing for a formatting? Are you using an IDE with a formatter / code cleanup type template? If so, can you post that Prof? Maybe add it to the repo? That way we can all be using the same formatter if we are using IDEs
> 
> I'm not using any form of IDE. I use emacs, which does some code formatting, but mostly I do it by hand.
> PS: I don't recommend emacs to anyone, it's extremely user hostile.
> I roughly follow the google style guide for c++: https://google.github.io/styleguide/cppguide.html

Oh I know emacs. Though I always preferred vi or vim. 

But ok. If you are following the basic cpp guide then a complimentary formatter template is pretty easy. 👍

adays1234:
> I thought about this pull request last night, and I'm wondering if it would be better to let the underlying (wrapped) font play the swing sounds. This has the advantage that it would also work for plecter style swing sounds.

That would simplify things, but would it give the fine-tuned control that sa22c is looking for?

profezzorn:
> > I thought about this pull request last night, and I'm wondering if it would be better to let the underlying (wrapped) font play the swing sounds. This has the advantage that it would also work for plecter style swing sounds.
> 
> That would simplify things, but would it give the fine-tuned control that sa22c is looking for?

It certainly could, because there is nothing stopping us from improving/changing/refactoring how hybrid_font.h plays swings.

adays1234:
> > > I thought about this pull request last night, and I'm wondering if it would be better to let the underlying (wrapped) font play the swing sounds. This has the advantage that it would also work for plecter style swing sounds.
> > 
> > 
> > That would simplify things, but would it give the fine-tuned control that sa22c is looking for?
> 
> It certainly could, because there is nothing stopping us from improving/changing/refactoring how hybrid_font.h plays swings.

I actually like that. It simplifies things a lot if smoothswing is simply overlaid into the normal swing players. Presuming that it is configurable to turn on/off and the same tuning can be achieved. It does make the overall structure more simple since it is layers instead of a decision tree in essence.

sound/smooth_swing_v2.h
```diff
@@ -8,39 +8,55 @@
...
+  
   void Activate(SaberBase* base_font) {
     STDOUT.println("Activating SmoothSwing V2");
     SetDelegate(base_font);
     if (swingl.files_found() != swingh.files_found()) {
       STDOUT.println("Warning, swingl and swingh should have the same number of files.");
     }
     swings_ = std::min<size_t>(swingl.files_found(), swingh.files_found());
+    //check for swngxx files to use as accent swings
+    if (swng.files_found() > 0){
```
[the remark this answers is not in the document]
SA-22C-smoothswing: done

sound/smooth_swing_v2.h
```diff
@@ -8,39 +8,55 @@
...
+      aswings_ = swng.files_found();
+      STDOUT.print("Accent Swings Detected: ");
+      STDOUT.println(swng.files_found());
+      accent_swings_present = true;
+      //allocate player
+      if(!accent_player_){
```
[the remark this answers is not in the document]
SA-22C-smoothswing: moved to line 121 after the logical check to trigger an accent swing.

sound/smooth_swing_v2.h
```diff
@@ -8,39 +8,55 @@
...
   void Activate(SaberBase* base_font) {
     STDOUT.println("Activating SmoothSwing V2");
     SetDelegate(base_font);
     if (swingl.files_found() != swingh.files_found()) {
       STDOUT.println("Warning, swingl and swingh should have the same number of files.");
     }
     swings_ = std::min<size_t>(swingl.files_found(), swingh.files_found());
+    //check for swngxx files to use as accent swings
+    if (swng.files_found() > 0){
+      aswings_ = swng.files_found();
```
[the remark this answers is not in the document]
SA-22C-smoothswing: we don't if the select function does the randomization for us.

sound/smooth_swing_v2.h
```diff
@@ -8,39 +8,55 @@
...
+      STDOUT.print("Accent Swings Detected: ");
+      STDOUT.println(swng.files_found());
+      accent_swings_present = true;
+      //allocate player
+      if(!accent_player_){
+          accent_player_ = GetFreeWavPlayer();
+      }
+    }
```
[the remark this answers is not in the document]
SA-22C-smoothswing: moved to 121 and resolved.

sound/smooth_swing_v2.h
```diff
@@ -8,39 +8,55 @@
...
     if (!on_) return;
     uint32_t m = millis();
     // No point in picking a new random so soon after picking one.
     if (A.player && m - last_random_ < 1000) return;
     last_random_ = m;
-    int swing = random(swings_);
     float start = m / 1000.0;
     A.Stop();
     B.Stop();
+    int swing = random(swings_);
```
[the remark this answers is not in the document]
SA-22C-smoothswing: removed.

sound/smooth_swing_v2.h
```diff
@@ -214,11 +261,13 @@ class SmoothSwingV2 : public SaberBasePassThrough {
   };
   Data A;
   Data B;
-
+  RefPtr<BufferedWavPlayer> accent_player_;
   uint32_t last_random_ = 0;
   bool on_ = false;;
   BoxFilter<Vec3, 3> gyro_filter_;
   int swings_;
+  int aswings_;
```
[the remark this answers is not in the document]
SA-22C-smoothswing: removed

sound/smooth_swing_v2.h
```diff
@@ -94,71 +110,87 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
         state_ = SwingState::ON;
         
       case SwingState::ON:
+        //check for AccentSwingThreshold, presence of accent swings and if the accent player is stopped (this prevents clipping)
+        if (speed >=smooth_swing_config.AccentSwingSpeedThreshold && accent_swings_present && !accent_player_->isPlaying() && (A.isPlaying() || B.isPlaying()))
+        {
+          accent_player_->PlayOnce(&swng);
+          //select new random swng accent
+          int accentswing = random(aswings_);
+          swng.Select(accentswing);
```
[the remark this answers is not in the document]
SA-22C-smoothswing: removed

sound/smooth_swing_v2.h
```diff
@@ -167,21 +199,36 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
-	player = GetFreeWavPlayer();
-	if (!player) return;
+        player = GetFreeWavPlayer();
+        if (!player) return;
       }
       player->set_volume(0.0);
       player->PlayOnce(effect, start);
       player->PlayLoop(effect);
     }
+    void PlayAccent(Effect* effect){
```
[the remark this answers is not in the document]
SA-22C-smoothswing: removed

sound/smooth_swing_v2.h
```diff
@@ -94,71 +110,87 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
-            mixab = clamp(- A.begin() / A.width, 0.0, 1.0);
-
+           mixab = clamp(- A.begin() / A.width, 0.0, 1.0);
+          
           float mixhum =
-            powf(swing_strength, smooth_swing_config.SwingSharpness);
-
+          powf(swing_strength, smooth_swing_config.SwingSharpness);
+          
+          float accent_volume =
```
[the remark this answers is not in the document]
SA-22C-smoothswing: moved volume calculations into if statement to check if accent_player is active.

sound/smooth_swing_v2.h
```diff
@@ -94,71 +110,87 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
           float mixab = 0.0;
           if (A.begin() < 0.0)
-            mixab = clamp(- A.begin() / A.width, 0.0, 1.0);
-
+           mixab = clamp(- A.begin() / A.width, 0.0, 1.0);
+          
           float mixhum =
-            powf(swing_strength, smooth_swing_config.SwingSharpness);
-
+          powf(swing_strength, smooth_swing_config.SwingSharpness);
```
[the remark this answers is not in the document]
SA-22C-smoothswing: fixed.

SA-22C-smoothswing:
ok, I've made all the requested changes.

lightsaber.ino
```diff
@@ -540,6 +540,9 @@ public:
     CONFIG_VARIABLE(Transition1Degrees, 45.0f);
     CONFIG_VARIABLE(Transition2Degrees, 160.0f);
     CONFIG_VARIABLE(MaxSwingVolume, 3.0f);
+    CONFIG_VARIABLE(AccentSwingSpeedThreshold, 500.0f);
```
profezzorn: I think this needs to default to zero.

sound/smooth_swing_v2.h
```diff
@@ -8,28 +8,39 @@
...
+  
   void Activate(SaberBase* base_font) {
     STDOUT.println("Activating SmoothSwing V2");
     SetDelegate(base_font);
     if (swingl.files_found() != swingh.files_found()) {
       STDOUT.println("Warning, swingl and swingh should have the same number of files.");
     }
     swings_ = std::min<size_t>(swingl.files_found(), swingh.files_found());
+    //check for swngxx files to use as accent swings
+    if (swng.files_found() > 0) {
```
profezzorn: && config_.AccentSwingSpeedThreshold > 0.0

sound/smooth_swing_v2.h
```diff
@@ -8,28 +8,39 @@
...
+      STDOUT.print("Accent Swings Detected: ");
+      STDOUT.println(swng.files_found());
+      accent_swings_present = true;
+    }
```
profezzorn: } else {

sound/smooth_swing_v2.h
```diff
@@ -8,28 +8,39 @@
...
   }
-
+  
   void Deactivate() {
     SetDelegate(NULL);
     A.Free();
     B.Free();
   }
-
+
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -8,28 +8,39 @@
...
   }
-
+  
   void Swap() {
     Data C = A;
     A = B;
     B = C;
   }
-
+
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -49,15 +60,15 @@ class SmoothSwingV2 : public SaberBasePassThrough {
     float t1_offset = random(1000) / 1000.0 * 50 + 10;
     A.SetTransition(t1_offset, smooth_swing_config.Transition1Degrees);
     B.SetTransition(t1_offset + 180.0,
-      smooth_swing_config.Transition2Degrees);
+           smooth_swing_config.Transition2Degrees);
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -49,15 +60,15 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
   }
-
+
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -49,15 +60,15 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
   }
-
+  
   void SB_On() override {
     on_ = true;
     // Starts hum, etc.
     delegate_->SB_On();
     PickRandomSwing();
-    if (!A.player || !B.player) {
+    if (!A.player || !B.player ) {
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -67,13 +78,13 @@ class SmoothSwingV2 : public SaberBasePassThrough {
     B.Off();
     delegate_->SB_Off();
   }
-
+
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -67,13 +78,13 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
   }
-
+  
   enum class SwingState {
     OFF, // waiting for swing to start
     ON,  // swinging
     OUT, // Waiting for sound to fade out
   };
-
+
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -94,38 +105,66 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+           STDOUT.println(speed);
           }
 #endif
           break;
         }
         state_ = SwingState::ON;
         
       case SwingState::ON:
+        //check for AccentSwingThreshold, presence of accent swings and if the accent player is stopped (this prevents clipping)
+        if (speed >=smooth_swing_config.AccentSwingSpeedThreshold
```
profezzorn: Please put the && at the end of the lines, and make sure that the new lines line up under each other.

sound/smooth_swing_v2.h
```diff
@@ -94,38 +105,66 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+          && accent_swings_present
+            && (A.player->isPlaying() || B.player->isPlaying())
+        {
```
profezzorn: move { to previous line

sound/smooth_swing_v2.h
```diff
@@ -94,38 +105,66 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+          //allocate player
+          if(!accent_player_) {
```
profezzorn: space after if

sound/smooth_swing_v2.h
```diff
@@ -94,38 +105,66 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+            accent_player_ = GetFreeWavPlayer();
+          }
+          else {
```
profezzorn: This shouldn't be an else.
Instead we need to check if (accent_player_) {
As it is, the first time we get here, we'll allocate the accent player, but not start playing.

sound/smooth_swing_v2.h
```diff
@@ -94,38 +105,66 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+            if (!accent_player_->isPlaying()) {
+              accent_player_->PlayOnce(&swng);
+            }
+          }
+        }
         if (speed >= smooth_swing_config.SwingStrengthThreshold * 0.9) {
           float swing_strength =
-            std::min<float>(1.0, speed / smooth_swing_config.SwingSensitivity);
+          std::min<float>(1.0, speed / smooth_swing_config.SwingSensitivity);
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -94,38 +105,66 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
           A.rotate(-speed * delta / 1000000.0);
           // If the current transition is done, switch A & B,
           // and set the next transition to be 180 degrees from the one
           // that is done.
           while (A.end() < 0.0) {
-            B.midpoint = A.midpoint + 180.0;
-	    Swap();
+           B.midpoint = A.midpoint + 180.0;
```
profezzorn: please undo indentation changes here and below

sound/smooth_swing_v2.h
```diff
@@ -94,38 +105,66 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
           mixhum *= smooth_swing_config.MaxSwingVolume;
-
+          
+          if (on_) {
+            // We need to stop setting the volume when off, or playback may never stop.
+            A.set_volume(mixhum * mixab);
+            B.set_volume(mixhum * (1.0 - mixab));
+            //This volume will scale with swing speed but is modulated by AccentSwingVolumeSharpness.
+            if (accent_player_) {
+              accent_volume =
```
profezzorn: need to check if accent player is playing.
If it is not, we don't need to calculate and set the volume.
In fact, if it is not, we should free it.

sound/smooth_swing_v2.h
```diff
@@ -142,41 +181,38 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
 private:
   struct Data {
     void set_volume(float v) {
       if (player) player->set_volume(v);
     }
     void Play(Effect* effect, float start = 0.0) {
       if (!player) {
-	player = GetFreeWavPlayer();
-	if (!player) return;
+        player = GetFreeWavPlayer();
```
profezzorn: please undo

lightsaber.ino
SA-22C-smoothswing: set

sound/smooth_swing_v2.h
SA-22C-smoothswing: added.

sound/smooth_swing_v2.h
SA-22C-smoothswing: ok, resolved.

SA-22C-smoothswing:
Provided that this smooth_swing_v2.h is clean and there are no further functional changes, I'd be happy to take a crack at moving these events into the hybrid_font.h file.

smooth_swing_v2.h
```diff
@@ -0,0 +1,259 @@
+#ifndef SOUND_SMOOTH_SWING_V2_H
```
profezzorn: Seems like this file has the wrong name?

sound/smooth_swing_v2.h
```diff
@@ -8,28 +8,34 @@
 class SmoothSwingV2 : public SaberBasePassThrough {
 public:
   SmoothSwingV2() : SaberBasePassThrough() {}
-
+
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -94,15 +97,29 @@ class SmoothSwingV2 : public SaberBasePassThrough {
         if (speed < smooth_swing_config.SwingStrengthThreshold) {
 #if 1
           if (monitor.ShouldPrint(Monitoring::MonitorSwings)) {
-            STDOUT.print("speed: ");
-            STDOUT.println(speed);
+           STDOUT.print("speed: ");
```
profezzorn: please undo (and next line)

sound/smooth_swing_v2.h
```diff
@@ -94,15 +97,29 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
           }
 #endif
           break;
         }
         state_ = SwingState::ON;
         
       case SwingState::ON:
+        //check for AccentSwingThreshold, presence of accent swings and if the accent player is stopped (this prevents clipping)
+        if (speed >=smooth_swing_config.AccentSwingSpeedThreshold && 
+        accent_swings_present &&
```
profezzorn: the first character (a) needs to go under the "s" in the previous line.
(same with next line)

sound/smooth_swing_v2.h
```diff
@@ -94,15 +97,29 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
 #endif
           break;
         }
         state_ = SwingState::ON;
         
       case SwingState::ON:
+        //check for AccentSwingThreshold, presence of accent swings and if the accent player is stopped (this prevents clipping)
+        if (speed >=smooth_swing_config.AccentSwingSpeedThreshold && 
+        accent_swings_present && 
+        (A.player->isPlaying() || B.player->isPlaying())){
```
profezzorn: add space before {

sound/smooth_swing_v2.h
```diff
@@ -94,15 +97,29 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+          //allocate player
+          if (!accent_player_) {
+            accent_player_ = GetFreeWavPlayer();
+          }
+          if (accent_player){
```
profezzorn: space before {

sound/smooth_swing_v2.h
```diff
@@ -112,20 +129,36 @@ class SmoothSwingV2 : public SaberBasePassThrough {
           // that is done.
           while (A.end() < 0.0) {
             B.midpoint = A.midpoint + 180.0;
-	    Swap();
+          Swap();
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -112,20 +129,36 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+          
           mixhum *= smooth_swing_config.MaxSwingVolume;
-
+          
+          if (on_) {
+            // We need to stop setting the volume when off, or playback may never stop.
+            A.set_volume(mixhum * mixab);
+            B.set_volume(mixhum * (1.0 - mixab));
+            //This volume will scale with swing speed but is modulated by AccentSwingVolumeSharpness.
+            if (accent_player_->isPlaying()) {
```
profezzorn: this will crash if there is currently no accent_player_

sound/smooth_swing_v2.h
```diff
@@ -112,20 +129,36 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+              accent_volume =
+              powf(swing_strength, smooth_swing_config.AccentSwingVolumeSharpness);
+              accent_volume *= smooth_swing_config.MaxAccentSwingVolume;
+              accent_player_->set_volume(accent_volume);
+            }
```
profezzorn: } else {

sound/smooth_swing_v2.h
```diff
@@ -142,41 +175,38 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
   struct Data {
     void set_volume(float v) {
       if (player) player->set_volume(v);
     }
     void Play(Effect* effect, float start = 0.0) {
       if (!player) {
-	player = GetFreeWavPlayer();
-	if (!player) return;
+        player = GetFreeWavPlayer();
+      if (!player) return;
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -142,41 +175,38 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
 private:
   struct Data {
     void set_volume(float v) {
       if (player) player->set_volume(v);
     }
     void Play(Effect* effect, float start = 0.0) {
       if (!player) {
-	player = GetFreeWavPlayer();
-	if (!player) return;
+        player = GetFreeWavPlayer();
```
profezzorn: please undo

sound/smooth_swing_v2.h
```diff
@@ -142,41 +175,38 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
         }
         PickRandomSwing();
+        accent_player_.Free();
         state_ = SwingState::OFF;
     }
     // Must always set hum volume, or fade-out doesn't work.
     delegate_->SetHumVolume(hum_volume);
   }
-
+
```
profezzorn: please undo

profezzorn:
PS: I highly recommend checking how github shows your changes. (essentially reviewing your own code)

smooth_swing_v2.h
SA-22C-smoothswing: uploaded into the wrong directory

sound/smooth_swing_v2.h
SA-22C-smoothswing: wrapped it in an if(accent_player) {} check

SA-22C-smoothswing:
Ok, fixed the crash issue and the formatting, I think.  I also uploaded the smoothswing file to the root rather than sound directory.  That has been fixed too.

SA-22C-smoothswing:
Ok, got the swings moved to hybrid_font.h

I'm stuck on how to mange volume fading on monophonic swings.  Because a monophonic sound also includes fading out hums, playing the sound and fading hums back in, I'm not sure how to proceed on getting the swing sound broken out so it can be manipulated by smooth_swing_v2.h to perform the volume fades.

profezzorn:
Monophonic swing sounds can't fade out in the same way as polyphonic ones can.
We will still adjust the volume slightly based on the speed, which will add some dynamic feeling.
(This would already be handled by SetHumVolume())

sound/hybrid_font.h
```diff
@@ -120,6 +121,36 @@ class HybridFont : public SaberBase {
       PlayPolyphonic(effect);
     }
   }
+  
+  void CommonSwing(Effect* monophonic, Effect* polyphonic) override {
```
profezzorn: Why pass in the effect here?
The hybrid font already knows what effect to use.
Also, I think "StartSwing()" would be a better name.

sound/hybrid_font.h
```diff
@@ -120,6 +121,36 @@ class HybridFont : public SaberBase {
...
+    } else {
+      PlayMonophonic(monophonic, &hum);
+    }
+  }
+  
+  void SetSwingVolume(float volume) override {
+    if (IsSwingPlaying()) {
+      swing_player_->set_volume(volume);
+    }
+
```
profezzorn: extra empty line

sound/hybrid_font.h
```diff
@@ -173,11 +204,11 @@ class HybridFont : public SaberBase {
 	  drag.files_found()) {
 	PlayMonophonic(&drag, &drag);
       } else if (lockup.files_found()) {
-        if (bgnlock.files_found()) {
-          PlayMonophonic(&bgnlock, &lockup);
-        } else {
+	if (bgnlock.files_found()) {
```
profezzorn: why change indentation here?
(and lots of places below)

sound/hybrid_font.h
```diff
@@ -120,6 +121,36 @@ class HybridFont : public SaberBase {
...
+    if (polyphonic->files_found()) {
+      if (!IsSwingPlaying()) {
+        swing_player_ = PlayPolyphonic(polyphonic);
+      }
+    } else {
+      PlayMonophonic(monophonic, &hum);
+    }
+  }
+  
+  void SetSwingVolume(float volume) override {
```
profezzorn: I wonder if it would be better if this function just took the current speed, that way it could (optionally) be used by HybridFont::SB_Motion() as well.

Or maybe a speed and a multiplier?

sound/smooth_swing_v2.h
```diff
@@ -103,6 +108,12 @@ class SmoothSwingV2 : public SaberBasePassThrough {
         state_ = SwingState::ON;
         
       case SwingState::ON:
+        //check for AccentSwingThreshold, presence of accent swings and if the accent player is stopped (this prevents clipping)
```
profezzorn: This line is too long.
Also, "clipping" is when the sound is too loud to fit in - 32768 - 32768.
I think you mean "prevents cutting off the sound"
Also, we don't actually check if the accent player is stopped here anymore

sound/smooth_swing_v2.h
```diff
@@ -112,20 +123,34 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
           mixhum *= smooth_swing_config.MaxSwingVolume;
-
+          
+          if (on_) {
+            // We need to stop setting the volume when off, or playback may never stop.
+            A.set_volume(mixhum * mixab);
+            B.set_volume(mixhum * (1.0 - mixab));
+            //This volume will scale with swing speed but is modulated by AccentSwingVolumeSharpness.
+            if (delegate_->IsSwingPlaying()) {
+              accent_volume =
```
profezzorn: I kind of think we should do these volume calculations in hybrid_font.h if possible.

sound/smooth_swing_v2.h
```diff
@@ -142,41 +167,37 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
         state_ = SwingState::OUT;
-
       case SwingState::OUT:
         if (!A.isOff() || !B.isOff()) {
           if (monitor.ShouldPrint(Monitoring::MonitorSwings)) {
             Serial.println("Waiting for volume = 0");
           }
         }
         PickRandomSwing();
+        accent_player_.Free();
```
profezzorn: there are still a lot of whitespace changes in this file

sound/smooth_swing_v2.h
```diff
@@ -8,28 +8,36 @@
...
     swings_ = std::min<size_t>(swingl.files_found(), swingh.files_found());
+    //check for swngxx files to use as accent swings
+    if ((swng.files_found() || swing.files_found()) > 0 && smooth_swing_config.AccentSwingSpeedThreshold > 0.0) {
+      STDOUT.println("Accent Swings Enabled.");
+      STDOUT.print("Polyphonic swings: ");
+      STDOUT.println(swng.files_found());
+      STDOUT.print("Monophonic swings: ");
+      STDOUT.println(swing.files_found());
+      accent_swings_present = true;
+    }else{
```
profezzorn: spaces around else

sound/hybrid_font.h
SA-22C-smoothswing: did this so I could assign a polyphonic_player object and manipulate volume.
I'll see if I can do that without passing the effects and change the function name

sound/hybrid_font.h
SA-22C-smoothswing: still fighting whitespace issues.  Thought I had them all resolved.

sound/hybrid_font.h
SA-22C-smoothswing: I'll play with that.

sound/smooth_swing_v2.h
SA-22C-smoothswing: ok, will update.

sound/smooth_swing_v2.h
SA-22C-smoothswing: ok, I can pass in the relevant variables and do it there.

sound/hybrid_font.h
SA-22C-smoothswing: I think I need to pass the effects in, but I'll change the function name.

sound/hybrid_font.h
SA-22C-smoothswing: tried my best, still getting weird whitespace issues.

sound/smooth_swing_v2.h
SA-22C-smoothswing: ok, moved all the math to hybrid_font.h.  It's pretty specialized for accent swings, but we could change it for other purposes down the road as required.

sound/smooth_swing_v2.h
SA-22C-smoothswing: done.

sound/smooth_swing_v2.h
SA-22C-smoothswing: Apologies, tried hand-ending to fix, but still showing up as 'changes' in the diff even though I can't see differences when looking at the files.

SA-22C-smoothswing:
Ok, fixed whitespaces.  I'm seeing some odd behaviour now that I'm testing more.

It seems that accent swings are only triggering half the time I expect them.  For example, if I swing the saber to the right at a speed that is faster than the threshold, I get an accent.  But if I then reverse the swing to the left, at the same (or greater speed) I don't get an accent swing.  Then if I swing again to the right, the accent triggers.

Any idea why that might be?

The other issue I'm having that I'm not sure how to resolve is detecting spins (this might be related to the swing issues listed above).  In reviewing Thexter's original algorithm compared to this implementation, I don't see a way of detecting spins.

SA-22C-smoothswing:
I'm also getting audio glitches (pops, clipping) on accent swings when volume is set low (ie: under 1000)

When volume is set to 1000 and up, I don't get the glitches.  Am I setting the volume for the accent swings incorrectly?

common/saber_base.h
```diff
@@ -68,7 +69,11 @@ class SaberBase {
   // 1.0 = kDefaultVolume
   // This is really just for sound fonts.
   virtual void SetHumVolume(float volume) {}
-
+  virtual void StartSwing(Effect* monophonic, Effect* polyphonic) {}
```
profezzorn: This function does not need to send in effect pointers.
HybridFont already knows what effects to use to start a swing.

lightsaber.ino
```diff
@@ -540,6 +540,10 @@ public:
     CONFIG_VARIABLE(Transition1Degrees, 45.0f);
     CONFIG_VARIABLE(Transition2Degrees, 160.0f);
     CONFIG_VARIABLE(MaxSwingVolume, 3.0f);
+    CONFIG_VARIABLE(AccentSwingSpeedThreshold, 0.0f);
+    CONFIG_VARIABLE(AccentSwingVolumeSharpness, 1.0f);
```
profezzorn: I think we should move these variables (except for AccentSwingSpeedThreshold) into config.ini, which can be used directly from hybrid font. That way they can be used to adjust swings for non-smoothswing fonts as well.

sound/hybrid_font.h
```diff
@@ -120,7 +122,37 @@ class HybridFont : public SaberBase {
       PlayPolyphonic(effect);
     }
   }
+  void StartSwing(Effect* monophonic, Effect* polyphonic) override {
+    if (polyphonic->files_found()) {
+      swing_player_ = PlayPolyphonic(polyphonic);
```
profezzorn: This can cause some weirdness if swing_player_ is already set.
Not sure what we want to happen in that case exactly.

sound/smooth_swing_v2.h
```diff
@@ -148,6 +166,12 @@ class SmoothSwingV2 : public SaberBasePassThrough {
 	    // We need to stop setting the volume when off, or playback may never stop.
 	    A.set_volume(mixhum * mixab);
 	    B.set_volume(mixhum * (1.0 - mixab));
+      if (delegate_->IsSwingPlaying()) {
```
profezzorn: Let's call IsSwingPlaying inside SetSwingVolume instead.
Then this class will not need to call IsSwingPlaying and it can be removed from SaberBase.

sound/hybrid_font.h
```diff
@@ -104,23 +99,51 @@ class HybridFont : public SaberBase {
...
-
   void PlayCommon(Effect* effect) {
     if (guess_monophonic_) {
       PlayMonophonic(effect, &hum);
     } else {
       PlayPolyphonic(effect);
     }
   }
-
+  void StartSwing(Effect* monophonic, Effect* polyphonic) override {
```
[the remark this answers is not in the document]
SA-22C-smoothswing: fixed.  changed to not pass in effects.  Assigns swing_player_ to polyphonic effect, plays monophonic effect normally.

common/saber_base.h
SA-22C-smoothswing: fixed.

lightsaber.ino
SA-22C-smoothswing: ok, i'll drop 'accent' from the parameter names and move them to config.ini  That way we're adjusting 'swing' parameters in config.ini and only calling 'accent' swings from smoothsw.ini

sound/hybrid_font.h
SA-22C-smoothswing: I did this so, when spinning a saber, the sound plays instantly.  Having a check first caused a delay that made swings not play fast enough when spinning.

lightsaber.ino
profezzorn: config.ini is the same file that Igniter uses, so we need to use some sort of naming that makes it clear that these variables are intended for ProffieOS. Simplest way is probably to just use replace "Accent" with "ProffieOS". Hopefully Igniter ignores variables it doesn't recognize.

SA-22C-smoothswing:
Ah, ok. That make sense.

Matthew McGeary
Senior Advisor, Datacenter, IT
matthew.mcgeary@nutrien.com
306.933.8921
www.nutrien.com

________________________________

sound/hybrid_font.h
profezzorn: Oh, I'm not saying that we should not start a swing if one is already playing.
The problem is simply that first swing sound will suddenly not have it's volume adjusted anymore.
I'm not sure if that's a problem, or not, but I thought it would be worth mentioning at least.
If we want to fix it, we would need to possibly keep track of more than one playing swing and adjust all their volumes in SetSwingVolume().

SA-22C-smoothswing:
Oh, that makes sense now. I’ll put in some logic to check and assign a second player.

Matthew McGeary
Senior Advisor, Datacenter, IT
matthew.mcgeary@nutrien.com
306.933.8921
www.nutrien.com

________________________________

sound/hybrid_font.h
SA-22C-smoothswing: fixed.  Added a second swing player to handle the case where two swings might play.

lightsaber.ino
SA-22C-smoothswing: done, moved variables to config.ini

SA-22C-smoothswing:
Ok, changes made as requested.

sound/hybrid_font.h
```diff
@@ -62,11 +68,15 @@ class HybridFont : public SaberBase {
...
     hum_player_.Free();
     next_hum_player_.Free();
+    swing_player_.Free();
+    second_swing_player_.Free();
     SaberBase::Unlink(this);
   }
 
   RefPtr<BufferedWavPlayer> hum_player_;
   RefPtr<BufferedWavPlayer> next_hum_player_;
+  RefPtr<BufferedWavPlayer> swing_player_;
```
profezzorn: maybe just make an array:

RefPtr<BufferedWavPlayer> swing_players_[2];

?

sound/hybrid_font.h
```diff
@@ -120,7 +130,49 @@ class HybridFont : public SaberBase {
...
+  }
 
+  bool IsSwingPlaying() {
+    if (swing_player_) {
+      if (second_swing_player_) {
+        if(second_swing_player_->isPlaying() || swing_player_->isPlaying()) {
+          return true;
+        }
+      } else {
+        if (!swing_player_) {
```
profezzorn: This is wrong.
We should free the swing_player_ if it is true but not playing.
Same for the second_swing_player_.

sound/hybrid_font.h
```diff
@@ -120,7 +130,49 @@ class HybridFont : public SaberBase {
...
+        swing_player_ = PlayPolyphonic(&swng);
+      } else {
+        second_swing_player_ = PlayPolyphonic(&swng);
+      }
+    } else {
+      PlayMonophonic(&swing, &hum);
+    }
+  }
+
+  float SetSwingVolume(float swing_strength, float mixhum) {
```
profezzorn: In order for non-smoothswing fonts to use this, hybrid font has to call this function somewhere.
However, doing so will change font behaviors in ways which I'm not sure are always better.
If we think that updating the swing volume is almost always better than not doing it, that's fine, but otherwise we need the ProffieOSSwingVolumeSharpness to default to 0.0 and ProffiOSMaxSwingVolume to default to whatever the swing defaults to normally? (I don't remember if that is 1.0 or 0.5)

sound/hybrid_font.h
```diff
@@ -120,7 +130,49 @@ class HybridFont : public SaberBase {
...
+    if (IsSwingPlaying()) {
+      float accent_volume = powf(swing_strength, config_.ProffieOSSwingVolumeSharpness) * config_.ProffieOSMaxSwingVolume;
```
profezzorn: ProffieOSMaxSwingVolume seems to be misnamed, as it is not a maximum.

sound/hybrid_font.h
```diff
@@ -120,7 +130,49 @@ class HybridFont : public SaberBase {
...
+      swing_player_->set_volume(accent_volume);
+      if (second_swing_player_) {
+        second_swing_player_->set_volume(accent_volume);
+      }
+      mixhum = mixhum * config_.ProffieOSMaxSwingDucking;
```
profezzorn: We really need to make the ducking larger when the accent_volume is larger.
Unconditionally ducking by a fixed amount while the swing is playing is not going to sound good.
It might be better to just not duck at all and let the dynamic mixer sort it out, but at least that is configurable right now.

sound/smooth_swing_v2.h
```diff
@@ -148,6 +166,7 @@ class SmoothSwingV2 : public SaberBasePassThrough {
 	    // We need to stop setting the volume when off, or playback may never stop.
 	    A.set_volume(mixhum * mixab);
 	    B.set_volume(mixhum * (1.0 - mixab));
+      mixhum = delegate_->SetSwingVolume(swing_strength, mixhum);
```
profezzorn: fix indentation?

profezzorn: are you sure you don't want to do this *before* A.set_volume()
That will duck all the hum volumes, not just the "main" hum.

SA-22C-smoothswing: good point, I've moved it.

sound/hybrid_font.h
```diff
@@ -120,7 +130,49 @@ class HybridFont : public SaberBase {
       PlayPolyphonic(effect);
     }
   }
+  void StartSwing() override {
+    if (!guess_monophonic_) {
+      if (!swing_player_) {
```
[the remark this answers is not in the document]
SA-22C-smoothswing: added a check to prevent allocating second_swing_player_ if it's already in use.

sound/hybrid_font.h
SA-22C-smoothswing: Default swing volume is effectively 1.0.  I'll test some mono/poly fonts with volume changing in place and see how it works.

sound/hybrid_font.h
SA-22C-smoothswing: fixed.

sound/hybrid_font.h
SA-22C-smoothswing: decided to have the 'second' swing player not persistent.

sound/smooth_swing_v2.h
SA-22C-smoothswing: removed the mix hum setting.  the ducking wasn't required.

sound/hybrid_font.h
SA-22C-smoothswing: removed ducking, not required.

sound/hybrid_font.h
SA-22C-smoothswing: its the maximum multiple for the swing sound over the hum, much like the smooth swing pairs.

sound/hybrid_font.h
SA-22C-smoothswing: Setting swing volume seems fine in the polyphonic NEC fonts I've tested.

sound/hybrid_font.h
SA-22C-smoothswing: removed second_swing_player

common/saber_base.h
```diff
@@ -6,6 +6,7 @@
 // has a corresponding SaberBase::Do* function which invokes that function
 // on all active SaberBases.
 class SaberBase;
+class Effect;
```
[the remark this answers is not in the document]
SA-22C-smoothswing: removed.

SA-22C-smoothswing:
Ok, fixed up some things.  

Removed the concept of a persistent second_swing_player_ and inserted a check in the StartSwing function to ensure that overlapping swings don't occur by checking if at least a quarter of the swing sound has played before fading out the current swing and playing a new one.  This actually improves the saber behaviour with polyphonic NEC fonts as those used to false trigger swings concurrently during a swing, but now that doesn't occur.

Added a parameter to config.ini for swing speed threshold, in order to make saber response to swings tuneable.  Default set to 250, which was the previously hardcoded swing speed trigger.

NEC fonts now get volume fade in/out just like accent swings.  No problems with it so far as I can tell in my testing.

Hopefully we're close to fixing all the issues. 👍

sound/hybrid_font.h
```diff
@@ -120,7 +128,37 @@ class HybridFont : public SaberBase {
...
     }
   }
+  
+  void StartSwing() override {
+    if (!guess_monophonic_) {
+      if (!swing_player_) {
+        swing_player_ = PlayPolyphonic(&swng);
+      }
+      // avoid overlapping swings unless at least a quarter of the swing has played.  Stops double swing events.
+      if ((swing_player_->length() - swing_player_->pos()) / swing_player_->length() <= 0.75) {
```
profezzorn: I do like this, although I think we should make the constant configurable.
Also, it seems like pos / length >= 0.25 would be simpler and more clear.

SB_motion() has a different ways to avoid double swings, which is that swing starts once speed goes over a constant, but you can't trigger another swing until speed drops below that constant, and then goes above it again.

Although, we probably need a little hysteresis, so that speed has to drop below 200 or something before swinging_ is set to false again.

sound/hybrid_font.h
```diff
@@ -120,7 +128,37 @@ class HybridFont : public SaberBase {
...
+        RefPtr<BufferedWavPlayer> overlap_swing = swing_player_;
+        swing_player_ = PlayPolyphonic(&swng);
```
profezzorn: how about move this out and after the if statement, then it can be combined with line 135 I think.

sound/hybrid_font.h
```diff
@@ -120,7 +128,37 @@ class HybridFont : public SaberBase {
...
+        overlap_swing->set_fade_time(0.1);
```
profezzorn: I'm not sure this is what we want to do.
Previously, swing sounds were just left to play out by themselves.

sound/hybrid_font.h
```diff
@@ -158,7 +196,7 @@ class HybridFont : public SaberBase {
     } else {
       state_ = STATE_HUM_FADE_OUT;
       PlayPolyphonic(&in);
-    }
+    }
```
profezzorn: extra whitespace after }

sound/smooth_swing_v2.h
```diff
@@ -112,7 +130,7 @@ class SmoothSwingV2 : public SaberBasePassThrough {
           // that is done.
           while (A.end() < 0.0) {
             B.midpoint = A.midpoint + 180.0;
-	    Swap();
+            Swap();
```
profezzorn: whitespace change?

sound/smooth_swing_v2.h
```diff
@@ -103,6 +115,12 @@ class SmoothSwingV2 : public SaberBasePassThrough {
         state_ = SwingState::ON;
         
       case SwingState::ON:
+        //trigger accent swing
```
profezzorn: space after //

sound/smooth_swing_v2.h
```diff
@@ -175,8 +194,8 @@ class SmoothSwingV2 : public SaberBasePassThrough {
     }
     void Play(Effect* effect, float start = 0.0) {
       if (!player) {
-	player = GetFreeWavPlayer();
-	if (!player) return;
+        player = GetFreeWavPlayer();
```
profezzorn: whitespace change

sound/smooth_swing_v2.h
```diff
@@ -175,8 +194,8 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
+        if (!player) return;
```
profezzorn: whitespace change

sound/hybrid_font.h
SA-22C: This only applies for rapid swings playing over each other, like spins. The idea is to allow rapid accent swing sounds to play over each other without having to worry about each swing sound and setting volume for each. For normal saber swinging, this doesn’t really factor in.

sound/hybrid_font.h
SA-22C: That’s a good point about the threshold. I’ll play with that. 

I agree about the overlap constant. I’ll make it a config variable.

sound/hybrid_font.h
SA-22C-smoothswing: I see what you mean.  I've removed the redundant swing_player_ assignment and moved the single call to below the if statement regarding playback position/overlay_swing

sound/hybrid_font.h
SA-22C-smoothswing: ProffieOSSwingOverlay defined.  What should the default be?

sound/hybrid_font.h
SA-22C-smoothswing: we could dynamically set the fade time to the remaining duration of the swing.  That way all we're really doing is a cross fade.  I'll give that a try.

SA-22C-smoothswing: does overlap_swing->set_fade_time(overlap_swing->length() - overlap_swing->pos()); look good for a dynamic fade time, based on the remaining play time of an effect?

SA-22C-smoothswing:
> I think we're getting near the end.
> I know this is has been a long and difficult process, but I'm happy that you're putting up with all my ideas, suggestions and nitpicks.

I started doing this mostly because I haven't done any serious programming since school.  It's been fun to go through reviews and try and re-learn some of this stuff.  I appreciate all the feedback and rigorous adherence to code style.  I've got the whitespaces fixed in smooth_swing_v2.h and I've made some changes to hybrid_font.h as requested.  I'm still thinking about the swing threshold stuff, more to come today.

sound/hybrid_font.h
SA-22C-smoothswing: I've added a check in the SB_Motion function that only sets swinging to false after speed falls below 80% of the threshold.  Still have to test and see if that value works well.  Should that threshold also be a config variable?

SA-22C-smoothswing:
Added back in a ducking parameter for mixhum.  Some users are encountering volume clipping during accent swings, testing a ducking parameter to see if that fixes it.  The parameter is combined with swing_strength to reduce mixhum by a percentage between 0 and 100.

sound/hybrid_font.h
```diff
@@ -120,7 +132,49 @@ class HybridFont : public SaberBase {
...
   }
+  
+  void StartSwing() override {
+    if (!guess_monophonic_) {
+      if (swing_player_) {
+        // avoid overlapping swings, based on value set in ProffieOSSwingOverlap.  Value is
+        // between 0 (no overlap) and 1.0 (full overlap)
+        if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
+          RefPtr<BufferedWavPlayer> overlap_swing = swing_player_;
+          swing_player_ = PlayPolyphonic(&swng);
```
profezzorn: If you move this down a few lines (after the Free()) then there is no need for the overlap_swing variable.

sound/hybrid_font.h
```diff
@@ -120,7 +132,49 @@ class HybridFont : public SaberBase {
...
+          overlap_swing->set_fade_time(overlap_swing->length() - overlap_swing->pos());
```
profezzorn: Is this really needed? I mean, the swing sound has it's own natural fade-out, doesn't it?

sound/hybrid_font.h
```diff
@@ -120,7 +132,49 @@ class HybridFont : public SaberBase {
...
+          overlap_swing->FadeAndStop();
+          overlap_swing.Free();
+        }
+      }
+      else if (!swing_player_) {
+        swing_player_ = PlayPolyphonic(&swng);
+        swing_player_->set_volume_now(0);
```
profezzorn: This is weird, we should not need to do this.
If this is what makes the crackle go away, then I have a theory:
When a new swing starts, it starts at default volume, which may be significantly higher than what we will set it to in SetSwingVolume(), and maybe the rapid changing of the volume is what causes the crackling.

We should probably be doing two things:
1) set the volume to the correct value, as calculated by SetSwingVolume() *before* we start playing the swing.
2) I'm not sure what the default fade time is, but we probably want to set it to something like 20ms or something as that avoids introducing a lot of noise from setting the volume too rapidly.

sound/hybrid_font.h
```diff
@@ -120,7 +132,49 @@ class HybridFont : public SaberBase {
...
+      PlayMonophonic(&swing, &hum);
+    }
+  }
 
+  float SetSwingVolume(float swing_strength, float mixhum) override {
+    if(swing_player_) {
+      if (swing_player_->isPlaying()) {
+        float accent_volume = powf(swing_strength, config_.ProffieOSSwingVolumeSharpness) * config_.ProffieOSMaxSwingVolume;
+        swing_player_->set_volume(accent_volume);
+        mixhum = mixhum - mixhum * (config_.ProffieOSSmoothSwingDucking * swing_strength);
```
profezzorn: seems like we should use accent_volume, not swing strength here
I'm still not convinced we need this.

sound/smooth_swing_v2.h
```diff
@@ -16,6 +16,18 @@ class SmoothSwingV2 : public SaberBasePassThrough {
       STDOUT.println("Warning, swingl and swingh should have the same number of files.");
     }
     swings_ = std::min<size_t>(swingl.files_found(), swingh.files_found());
+    //check for swngxx files to use as accent swings
```
profezzorn: add space after //

sound/smooth_swing_v2.h
```diff
@@ -144,11 +162,14 @@ class SmoothSwingV2 : public SaberBasePassThrough {
...
             STDOUT.println(hum_volume);
           }
-	  if (on_) {
-	    // We need to stop setting the volume when off, or playback may never stop.
-	    A.set_volume(mixhum * mixab);
-	    B.set_volume(mixhum * (1.0 - mixab));
-	  }
+          if (on_) {
+            // We need to stop setting the volume when off, or playback may never stop.
+            if (accent_swings_present) {
```
profezzorn: we don't need to check if accent swings are present here.
Calling SetSwingVolume should be safe even if accent swings are not present.

sound/hybrid_font.h
SA-22C: Ok I’ll test that out.

sound/hybrid_font.h
SA-22C: I’ll test without and see.

sound/hybrid_font.h
SA-22C: Won’t that introduce a delay in switching swing sounds?

profezzorn: No, and I'm not sure why you would think it would.

sound/hybrid_font.h
SA-22C-smoothswing: I'll test with accent_volume and see how it works.  I think we want this parameter for a couple reasons:

Testers have reported audio issues in a couple of circumstances:
1) the volume crackling that appears to be fixed or at least reduced frequency by setting player volume to 0
2) Additional volume crackling issues on loud or bass heavy accent swings that are 'fixed' by reducing accent swing volume.  This parameter lets users and font makers choose which trade-off to make: they can reduce accent swing volume using the max parameter or they can reduce swing pair volume during swings using the ducking parameter.

sound/smooth_swing_v2.h
SA-22C-smoothswing: It's safe provided that the defaults are set so that swing pairs are not ducked.  I'll make the required changes.

sound/hybrid_font.h
SA-22C-smoothswing: seems to function the same with the overlay_swing variable removed and no fade.

sound/hybrid_font.h
SA-22C-smoothswing: removed fade time

sound/hybrid_font.h
SA-22C-smoothswing: added new config variable ProffieOSSwingLowerThreshold, set to .8 by default.  This works great on monophonic fonts to prevent double swings and still be responsive.

sound/hybrid_font.h
SA-22C-smoothswing: It's required, odd-sounding cutoffs occur without it.

sound/hybrid_font.h
SA-22C-smoothswing: All I had to do was set fade_time to 2 just before setting volume in SetSwingVolume and all odd pops and crackles went away.

sound/hybrid_font.h
SA-22C-smoothswing: using accent_Volume works well, and via testing the ducking parameter is required to reduce peaks below 32k

sound/hybrid_font.h
SA-22C-smoothswing: Put fade time back in.  During testing, odd cutoffs occurred without it.

sound/hybrid_font.h
SA-22C-smoothswing: after deeper testing across multiple fonts, not having the combination of overlay_player_ and fade out causes odd cut off for some swings.  Having this separate player and the fade and stop function makes all overlays sound as natural as possible.

sound/hybrid_font.h
```diff
@@ -120,7 +134,49 @@ class HybridFont : public SaberBase {
...
+    } else {
+      PlayMonophonic(&swing, &hum);
+    }
+  }
 
+  float SetSwingVolume(float swing_strength, float mixhum) override {
+    if(swing_player_) {
+      if (swing_player_->isPlaying()) {
+        float accent_volume = powf(swing_strength, config_.ProffieOSSwingVolumeSharpness) * config_.ProffieOSMaxSwingVolume;
+        swing_player_->set_fade_time(2);
```
profezzorn: This sets the fade time to 2 seconds, which is really slow.
Basically it means that the volume doesn't change.

sound/hybrid_font.h
```diff
@@ -120,7 +134,49 @@ class HybridFont : public SaberBase {
...
+        swing_player_->set_volume(accent_volume);
+        mixhum = mixhum - mixhum * (config_.ProffieOSSmoothSwingDucking * accent_volume);
+      }
```
profezzorn: } else {

sound/hybrid_font.h
profezzorn: we can still have fade and stop without having a separate variable.

sound/hybrid_font.h
```diff
@@ -282,13 +338,16 @@ class HybridFont : public SaberBase {
...
 	  !(lockup.files_found() && SaberBase::Lockup())) {
         swinging_ = true;
-        Play(&swing, &swng);
+        StartSwing();
       }
-    } else {
+      float swing_strength = std::min<float>(1.0, speed / config_.ProffieOSSwingSpeedThreshold);
+      SetSwingVolume(swing_strength, 1.0);
+    } else if (swinging_ && speed <= config_.ProffieOSSwingSpeedThreshold * 
+               ProffieOSSwingLowerThreshold, config_.ProffieOSSwingLowerThreshold) {
```
profezzorn: the comma here probably doesn't do what you think.
Besides, it's probably better to just do:

speed <= config_.ProffieOSSwingLowerThreshold and default it to ~200

sound/hybrid_font.h
SA-22C: Hmm.  Well I misunderstood the units on that variable. ;)

sound/hybrid_font.h
SA-22C-smoothswing: that is simpler, yes.  fixed.

sound/hybrid_font.h
SA-22C-smoothswing: set to 0.020 (20ms)

sound/hybrid_font.h
SA-22C-smoothswing: I'll test again with the proper fade time set in SetSwingVolume but for whatever reason if I don't use a separate variable in my testing, I get an odd 'hiccup' on swing starts for some fonts.

profezzorn: If the code as written above gives you a hiccup, but the other code doesn't, then there is a bug somewhere, and we should find and fix it...

SA-22C-smoothswing: I think the issue arises because we're setting volume on the swing_player so if we don't hand off the overlain swing to a different variable, we get odd audio effects because both swings are getting volume set by SetSwingVolume.

SA-22C-smoothswing: hiccup issue resolved with correct fade-time variable set in SetSwingVolume.  a value of 0.02 fixes hum hiccups and seems to eliminate crackles.  Resolved.

SA-22C-smoothswing: fixed, hiccup and crackles seem resolved with setting a short fade_time in SetSwingVolume (20ms)

PR #17, opened 2019-06-10
hybrid_font.h changes and additions

SA-22C-smoothswing:
Support added for NEC slsh and Plecter stab.

I need some assistance on the stab effect.  I've got it triggering reliably but it often gets stepped on by a swing sound.  I'm not sure how to prevent that from happening.

Once we get that sorted, I'll move on to ignition hum based on orientation and properly detecting/playing Plecter spin sounds.


profezzorn:
This change contain multiple changes, which is not really a good idea.
It also contains to comments about why any of those changes are for or how they work.
Finally it contains lots of whitespace-only changes, which it shouldn't.
My inclination is to close this pull request and break it up into multiple smaller requests, each containing exactly one feature or bugfix.

SA-22C-smoothswing:
Ok, lets close it.  I'll break my changes up into smaller chunks.

PR #18, opened 2019-06-14
Slash support - NEC and smooth swing

SA-22C:
Added support for NEC slash sounds both for plain NEC fonts and for smooth swing.


sound/effect.h
```diff
@@ -362,6 +362,7 @@ EFFECT(pwroff);
 EFFECT(clash);
 EFFECT(force);  // also polyphonic
 EFFECT(stab);   // also polyphonic
+EFFECT(spin);
```
profezzorn: Why add spin to implement slash?

sound/hybrid_font.h
```diff
@@ -145,14 +147,32 @@ class HybridFont : public SaberBase {
...
+        PlayMonophonic(&swing, &hum);
+      }
+    }
+  }
+  
+  void StartSlash() override {
+    if (swing_player_) {
+      // avoid overlapping swings, based on value set in ProffieOSSwingOverlap.  Value is
+      // between 0 (full overlap) and 1.0 (no overlap)
+      if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
```
profezzorn: So how is this going to work?
By the time we get here, we will have almost always started a swing very recently.
Because of inertia, we'll start at a speed near zero, and once we reach 200 we'll start a swing, and then we'll get here soon after, and then we won't play a slash because we just started a swing.

I have two possible ideas for how we deal with this:
1) If a swing is playing, we transition to slash regardless of overlap. (However, if a *slash* is playing, we take the overlap into account.)
2) Instead of looking at the swing speed to decide if it's a slash or a swing, we could look at the derivative.   When we reach the swing speed, if the speed is accelerating rapidly, we could just start a slash right away instead of playing a swing first.

sound/hybrid_font.h
```diff
@@ -363,19 +383,29 @@ class HybridFont : public SaberBase {
...
   
   bool swinging_ = false;
+  bool slashing_ = false;
   void SB_Motion(const Vec3& gyro, bool clear) override {
     float speed = sqrtf(gyro.z * gyro.z + gyro.y * gyro.y);
-    if (speed > config_.ProffieOSSwingSpeedThreshold) {
-      if (!swinging_ && state_ != STATE_OFF &&
-	  !(lockup.files_found() && SaberBase::Lockup())) {
-        swinging_ = true;
+    if (state_ != STATE_OFF && !(lockup.files_found()
```
profezzorn: I wonder if we can break out the code that calls startswing/startslash into a a single function which is called from both hybrid and smoothswing fonts?

sound/effect.h
SA-22C: Was experimenting with using X acceleration to trigger stabs but pulled it out for this merge.  Missed the spin declaration in effects.h.  Will remove.

sound/hybrid_font.h
SA-22C: I agree that it seems counter intuitive based on the code but in testing this works decently.  I think your idea of transitioning to slash regardless of overlap will sound better.  I'll give that a good.

sound/hybrid_font.h
SA-22C: I think we can.  We'll just need to pass in a variable to tell the new generic swing function which swing sound it will be playing.  Do you recommend a enumerated state list, so if STATE_SLASHING is set in domotion, then we can have the case statement in generic swing.

sound/hybrid_font.h
profezzorn: I actually think using the acceleration would sound the best, I just don't know if it would *work* the best.

sound/hybrid_font.h
profezzorn: I was thinking more along the line of just passing in the speed or the gyro vector and letting the function decide what to do.

sound/effect.h
SA-22C: removed.

sound/hybrid_font.h
SA-22C: Using an acceleration change derived from speed, sampled every 10 ms

sound/hybrid_font.h
SA-22C: I'm just passing the acceleration variable into the start swing function.

SA-22C:
Ok, the derived acceleration value is in, used to trigger slashes.  However, it appears somewhat inconsistent, so that every other swing is detected as a slash.  Something is causing the acceleration/speed values to be under-represented during a subsequent swing/slash.

SA-22C:
I'm seeing odd audio lags on your current code base, with and without my changes.  blaster block files will show as playing on the serial monitor but no sounds are played.  When swinging the saber in a smooth swing font with accents, the accent swing will be distorted and elongated, as if it's 'stuck'.

profezzorn:
The audio lag seems unrelated to this change and should probably be discussed elsewhere.
Could you perhaps report an "issue" for that instead?

sound/hybrid_font.h
```diff
@@ -367,11 +372,16 @@ class HybridFont : public SaberBase {
   bool swinging_ = false;
   void SB_Motion(const Vec3& gyro, bool clear) override {
     float speed = sqrtf(gyro.z * gyro.z + gyro.y * gyro.y);
+    if (millis() - last_swing_millis_ >= 10 ) {
```
profezzorn: This is going to be a bit uneven.
I would think that the acceleration is changing relatively sharply around the time that the swing triggers.
So it will then depend on where in the 10-millisecond cycle the swing occurs.

Maybe something like this would work better:
   uint32_t now = micros();
   uint32_t delta_micros = now - last_swing_micros_;
   last_swing_micros_ = now;
   float delta = delta_micros * 0.000001;
   float current_acceleration  = (speed - last_speed_) / delta;
   float filter_factor = powf(0.2, delta);
   swing_acceleration_ = swing_acceleration_ * filter_factor + current_acceleration_ * (1 - filter_factor);

Ultimately, it might be that acceleration isn't the best way to do this though.

common/saber_base.h
```diff
@@ -74,7 +74,7 @@ class SaberBase {
   // 1.0 = kDefaultVolume
   // This is really just for sound fonts.
   virtual void SetHumVolume(float volume) {}
-  virtual void StartSwing() {}
+  virtual void StartSwing(float swing_acceleration_) {}
```
profezzorn: underscore at end is only for variables in objects

sound/hybrid_font.h
```diff
@@ -147,11 +149,14 @@ class HybridFont : public SaberBase {
           swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
           swing_player_->FadeAndStop();
           swing_player_.Free();
-          swing_player_ = PlayPolyphonic(&swng);
         }
       }
-      else if (!swing_player_) {
-        swing_player_ = PlayPolyphonic(&swng);
+      if (!swing_player_) {
+        if (fabs(swing_acceleration_) > config_.ProffieOSSlashAccelerationThreshold && slsh.files_found()) {
```
profezzorn: why fabs() ?
It doesn't make any sense, because if swing_accelration is negative, it means that the we reached the swing threshold by slowing down?

sound/hybrid_font.h
```diff
@@ -147,11 +149,14 @@ class HybridFont : public SaberBase {
...
+          swing_player_ = PlayPolyphonic(&slsh);
+        } else if (!swinging_){
```
profezzorn: adding the !swinging_ check is unrelated to the slash implementation, and I'm not sure it's correct either.

sound/hybrid_font.h
```diff
@@ -162,7 +167,7 @@ class HybridFont : public SaberBase {
     if(swing_player_) {
       if (swing_player_->isPlaying()) {
         float accent_volume = powf(swing_strength, config_.ProffieOSSwingVolumeSharpness) * config_.ProffieOSMaxSwingVolume;
-        swing_player_->set_fade_time(0.04);
+        swing_player_->set_fade_time(0.1);
```
profezzorn: this is unrelated to the slash implementation

sound/hybrid_font.h
```diff
@@ -369,16 +374,23 @@ class HybridFont : public SaberBase {
...
 	  !(lockup.files_found() && SaberBase::Lockup())) {
-        swinging_ = true;
-        StartSwing();
-      }
+      StartSwing(swing_acceleration_);
+      swinging_ = true;
       float swing_strength = std::min<float>(1.0, speed / config_.ProffieOSSwingSpeedThreshold);
-      SetSwingVolume(swing_strength, 1.0);
-    } else if (swinging_ && speed <= config_.ProffieOSSwingSpeedThreshold) {
+      //SetSwingVolume(swing_strength, 1.0);
```
profezzorn: Why is this commented out now?

sound/smooth_swing_v2.h
```diff
@@ -95,6 +104,11 @@ class SmoothSwingV2 : public SaberBasePassThrough {
     // degrees per second
     // May not need to smooth gyro since volume is smoothed.
     float speed = sqrtf(gyro.z * gyro.z + gyro.y * gyro.y);
+    if (millis() - last_swing_millis_ >= 10 ) {
+      swing_acceleration_ = (speed - last_speed_) / (millis() - last_swing_millis_);
```
profezzorn: We shouldn't be calculating the acceleration in two places.
Also, this calculation is significantly different from the one in the other code, which is not going to be good.

sound/smooth_swing_v2.h
```diff
@@ -95,6 +104,11 @@ class SmoothSwingV2 : public SaberBasePassThrough {
     // degrees per second
     // May not need to smooth gyro since volume is smoothed.
     float speed = sqrtf(gyro.z * gyro.z + gyro.y * gyro.y);
+    if (millis() - last_swing_millis_ >= 10 ) {
```
profezzorn: doing this every 10 ms is pretty coarse, I like the other acceleration calculation better.

common/saber_base.h
SA-22C: > In general, how is the acceleration thing working for you?
> Does it detect the difference between a slash and a swing?

It is working pretty well, yes.  I need to test with a different font but with the font I've been using so far, it's been consistent.

sound/hybrid_font.h
SA-22C: In theory yes.  In practise, it seems like the acceleration is almost always negative.

sound/hybrid_font.h
SA-22C: Yes, I found a bug in the swing code while I was testing.  This check is to ensure that swing files don't play more than once during a long sweeping swing.

sound/hybrid_font.h
SA-22C: yes, I'll revert.

sound/hybrid_font.h
SA-22C: I was getting terrible crackles and pops on slashes that overlap heavily, just commented out during testing until I could get back to that issue and see what the issue is.

sound/smooth_swing_v2.h
SA-22C: Yes, I haven't updated the smooth_swing_v2.h file yet to match the work in hybrid_font.

sound/smooth_swing_v2.h
SA-22C: Agreed, this is old code and needs to be replaced.

sound/hybrid_font.h
profezzorn: That a problem that we should figure out, not work around.
Maybe the filter factor is too big. Maybe try 0.01 instead of 0.2 or something?

sound/hybrid_font.h
profezzorn: That's cool, but bugfixes should really go in separate pull requests.

sound/smooth_swing_v2.h
profezzorn: That seems like a reasonable way to do it.
Another way would be to move more of the smoothswing stuff into hybrid font and just send thresholds to it, not sure if that's any better though.

sound/hybrid_font.h
SA-22C: I'll give that a go

sound/hybrid_font.h
SA-22C: Provided I reference this value as an absolute, it works great.  If I don't, the values are mostly negative.  Changing the filter factor doesn't make any difference.

sound/hybrid_font.h
SA-22C: as mentioned, changing the filter_factor doesn't alter the values of swing_acceleration_

I'm not sure why we are seeing negative values but after extensive testing, using the absolute value to test for slashes works perfectly.  If I had to guess, there seems to be an oddity that swing speed slows down to near zero when a swing sound plays, perhaps something is blocking the gyro from reading properly?  In any event, using absolute sidesteps this issue.

sound/hybrid_font.h
SA-22C: This structure has to change as part of the migration of the speed related code into the startswing function.

sound/hybrid_font.h
SA-22C: reverted

sound/smooth_swing_v2.h
SA-22C: all speed related code is migrated to hybrid_font.h StartSwing except for the speed stuff required for smooth swing v2.

sound/smooth_swing_v2.h
SA-22C: removed, replaced with code in hybrid_font.h

common/saber_base.h
SA-22C: further testing shows it working very consistently.

sound/hybrid_font.h
profezzorn: Unfortunately, it's not enough for the code to work.
It must also make sense, or nobody (including me) will ever be able to change it, understand it or fix bugs in it.

Using fabs() doesn't make sense.
Let's say we plot swing speed over time. Then the swing threshold is a line parallel to the X axis. Once the swing speed goes over that line, we trigger a swing or slash.  At the point in time when we go over the line, the slope of the curve MUST BE POSITIVE, because otherwise the swing speed would be going down, not up.

Sidestepping the issue just isn't ok.

SA-22C: Ok, I've reworked the start swing function and I've got slashes triggering predictably now, but there were a bunch of changes to make it work.

sound/hybrid_font.h
```diff
@@ -138,24 +140,57 @@ class HybridFont : public SaberBase {
...
           swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
           swing_player_->FadeAndStop();
           swing_player_.Free();
-          swing_player_ = PlayPolyphonic(&swng);
         }
       }
-      else if (!swing_player_) {
-        swing_player_ = PlayPolyphonic(&swng);
+      if (!swing_player_) {
+        if (doSlash_) {
```
profezzorn: This doesn't seem to work the way I imagined it would.
The whole point of using the acceleration was to have one start point (based on speed) and at that point we would decide if it's a slash or a swing based on the acceleration.

This code seems to play a slash as soon as the acceleration is high enough, which could happen before or after the swing gets kicked off. It seems like the swing and slash would fight each other quite a bit this way....

SA-22C: You're right of course.  I've corrected this error and the slashes are very predictable now.

common/saber_base.h
```diff
@@ -82,7 +82,7 @@ class SaberBase {
   // 1.0 = kDefaultVolume
   // This is really just for sound fonts.
   virtual void SetHumVolume(float volume) {}
-  virtual void StartSwing() {}
+  virtual void StartSwing(const Vec3& gyro, float swingThreshold, float slashThreshold) {}
```
profezzorn: Why are swingThreshold and slashThreshold arguments?

sound/hybrid_font.h
```diff
@@ -138,25 +137,43 @@ class HybridFont : public SaberBase {
...
+          // avoid overlapping swings, based on value set in ProffieOSSwingOverlap.  Value is
+          // between 0 (full overlap) and 1.0 (no overlap)
+          if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
+            swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
+            swing_player_->FadeAndStop();
+            swing_player_.Free();
+          }
         }
+        if (!swing_player_){
+          if (accel_.len() > slashThreshold_ && slsh.files_found() && !swinging_) {
```
profezzorn: This doesn't make any sense to me.
accel_.len() is the length of the acceleration vector, however, without subtracting out the "down" vector, the length isn't very meaningful. Also, the way this is written, I would imagine that up and down swings would get very different results for slashes.

If we subtract out the "down" vector, accel.len() will give us how fast the hilt is accelerating in some direction.
However, hilt movement and blade movement aren't really the same thing. For some styles of swinging they might be the same, but not for all.

common/saber_base.h
SA-22C-smoothswing: Because they are sourced from smoothsw.ini for smooth swing and config.ini for legacy fonts.

sound/hybrid_font.h
SA-22C-smoothswing: In testing this value seemed consistent for triggering on a slashing motion (a long fast swing) and the values were reliably larger on the faster motions.   i can experiment with the down vector and see how things change.

profezzorn: Like I pointed out before: It's not enough that it works, it must also make sense.
Code that just works, but doesn't make sense is impossible to maintain, update or rewrite.

SA-22C: Ok, so tested with subtracting fusor.down from accel.  Slsh seems to be easier to trigger now (which makes sense, because the value is lower due to the above) and it seems more consistent.  I did some testing with the modified accel.len printing during swings and it seems very consistently tied to swing speed no matter how the hilt is moving.  So this change appears to make triggering a slash a little more repeatable.

profezzorn: It's still weird to me to use acceleration for this though, as hilt acceleration comes from moving the hilt. But I think of a "slash" as moving the *blade* quickly, not the hilt.

Also, accel_ - fusor.down() is the same as mss, so we should probably just have a fusor.mss() function IF that is what we should use for slash detection.

SA-22C: Ok, I can change to mss easily enough.  However, what do you think would be a better metric to detect blade vs hilt acceleration?

profezzorn: I think you mean "what would be a better metric to detect slash vs. swing" ?

I still think the rate-of-change for the gyro might be a good way to do this. The fusor actually calculates this already, but doesn't expose it. I can try to fix that....

profezzorn: I don't know if you already saw this, but try using this for the threshold:

// fusor.gyro_slope() is in degrees per second per microsecond
Vec3 gyro_slope = fusor.gyro_slope() * 1000000; 
float dss = sqrtf(gyro_slope.z * gyro_slope.z + gyro_slope.y * gyro_slope.y);

(Then use "dss" instead of accel_.len())

SA-22C-smoothswing: I was playing with slope but didn't think to add the multiplier.  I'll do that and see what I come up with.

profezzorn: Might be better to put that multiplier in the fusor, as degrees per second per second is a much more reasonable unit than degrees per second per microsecond.

SA-22C-smoothswing: hmm, the values seem very high.  I'm seeing deg/s/s values of 20k +

profezzorn: That does seem high.
Unfortunately I haven't used this code enough to develop an intuition for whether that is unreasonable or not. I also haven't tested the code very much, so there could be bugs...

I remember we tried calculating delta-gyro before, but I don't remember what range of values you ended up with then.

SA-22C-smoothswing: The same issue occurred, where the values were crazy high in addition to being inconsistent.  It seems pretty unlikely that I could accelerate a hilt at 20 Km/s. ;)  The good news is that the values this time seem to be more consistent than before so even though the high numbers seem off, they're consistently off.

profezzorn: The unit is not m/s though. It's degrees per second per second, which is very different.
If we convert everything to radians, it might be more like meters per second though.
20k degrees per second = 348 radians per second per second. If the blade is one meter, the tip would experience an acceleration of 348 m/s/s, which is about 35G, which seems high, but not inconceivable. If we had an accelerometer in the tip of the blade we could test it. :)

SA-22C-smoothswing: Ok, did some more testing with the dss value.  It seems consistent and tunable, so success.  I've decided to have the value in radians/s/s so we don't have to have values in the config file in the 5 digit range.

sound/hybrid_font.h
```diff
@@ -138,25 +138,46 @@ class HybridFont : public SaberBase {
...
-        // between 0 (full overlap) and 1.0 (no overlap)
-        if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
-          swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
-          swing_player_->FadeAndStop();
-          swing_player_.Free();
-          swing_player_ = PlayPolyphonic(&swng);
+
+  void StartSwing(const Vec3& gyro, float swingThreshold_, float slashThreshold_) override {
+    Vec3 gyro_slope = fusor.gyro_slope();
+    float rss = sqrtf(gyro_slope.z * gyro_slope.z + gyro_slope.y * gyro_slope.y)/57.3;
```
profezzorn: add a comment like // radians per second per second
Also, instead of /57.3, do: * (M_PI / 180.0);
It's more clear, and potentially faster (multiplication is a little simpler than division).

sound/hybrid_font.h
```diff
@@ -138,25 +138,46 @@ class HybridFont : public SaberBase {
...
+    float speed = sqrtf(gyro.z * gyro.z + gyro.y * gyro.y);
```
profezzorn: rename "speed" to "swing_speed".
Use fusor.swing_speed() to calculate it.

sound/hybrid_font.h
```diff
@@ -138,25 +138,46 @@ class HybridFont : public SaberBase {
...
+        if (swing_player_) {
+          // avoid overlapping swings, based on value set in ProffieOSSwingOverlap.  Value is
+          // between 0 (full overlap) and 1.0 (no overlap)
+          if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
+            swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
+            swing_player_->FadeAndStop();
+            swing_player_.Free();
+          }
         }
+        if (!swing_player_){
```
profezzorn: missing space before {

sound/hybrid_font.h
```diff
@@ -138,25 +138,46 @@ class HybridFont : public SaberBase {
...
+          STDOUT.println(rss);
```
profezzorn: remove

sound/hybrid_font.h
```diff
@@ -138,25 +138,46 @@ class HybridFont : public SaberBase {
...
-      if (swing_player_) {
-        // avoid overlapping swings, based on value set in ProffieOSSwingOverlap.  Value is
-        // between 0 (full overlap) and 1.0 (no overlap)
-        if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
-          swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
-          swing_player_->FadeAndStop();
-          swing_player_.Free();
-          swing_player_ = PlayPolyphonic(&swng);
+
+  void StartSwing(const Vec3& gyro, float swingThreshold_, float slashThreshold_) override {
```
profezzorn: underscores at the end of variables should only be used for object level variables.
Local variables and arguments should not have them.

sound/hybrid_font.h
```diff
@@ -138,25 +138,46 @@ class HybridFont : public SaberBase {
...
+            swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
+            swing_player_->FadeAndStop();
+            swing_player_.Free();
+          }
         }
+        if (!swing_player_){
+          STDOUT.println(rss);
+          if (rss > slashThreshold_ && slsh.files_found() && !swinging_) {
+            swing_player_ = PlayPolyphonic(&slsh);
+          } else if (!swinging_){
```
profezzorn: missing space before {

sound/hybrid_font.h
```diff
@@ -138,25 +138,46 @@ class HybridFont : public SaberBase {
...
+          // between 0 (full overlap) and 1.0 (no overlap)
+          if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
+            swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
+            swing_player_->FadeAndStop();
+            swing_player_.Free();
+          }
         }
+        if (!swing_player_){
+          STDOUT.println(rss);
+          if (rss > slashThreshold_ && slsh.files_found() && !swinging_) {
```
profezzorn: !swinging_ needs to be in an if statement surrounding this if statement:

if (!swinging) {
   if (rss ......) {
   } else {
   }
}

sound/smooth_swing_v2.h
```diff
@@ -26,7 +26,16 @@ class SmoothSwingV2 : public SaberBasePassThrough {
       accent_swings_present = true;
     } else {
       accent_swings_present = false;
-      STDOUT.print("Accent Swings NOT Detected: ");
+      STDOUT.println("Accent Swings NOT Detected: ");
+    }
+    if (slsh.files_found() > 0 && smooth_swing_config.AccentSlashAccelerationThreshold > 0.0) {
```
profezzorn: I think we should actually default slashes to be on if present AND accent swings are enabled.
I think that might mean that we don't need to have separate variables for smooth swing and non-smoothswing fonts.

sound/hybrid_font.h
SA-22C-smoothswing: ok cool, will get that changed.

sound/hybrid_font.h
SA-22C-smoothswing: ok, will do.

sound/smooth_swing_v2.h
SA-22C-smoothswing: I think we need to have separate settings for the same reason we have separate settings for swing and accent swings: hybrid fonts tend to have different threshold requirements than non-hybrid

sound/hybrid_font.h
SA-22C-smoothswing: resolved

sound/hybrid_font.h
SA-22C-smoothswing: resolved

sound/hybrid_font.h
SA-22C-smoothswing: resolved

sound/hybrid_font.h
SA-22C-smoothswing: resolved

sound/hybrid_font.h
SA-22C-smoothswing: resolved

sound/hybrid_font.h
SA-22C-smoothswing: resolved

sound/hybrid_font.h
SA-22C-smoothswing: Resolved.

sound/smooth_swing_v2.h
profezzorn: I don't know if it is strictly needed, but it's probably going to confuse users more if we don't separate them, so let's go with that, at least for now.

profezzorn: Accent slashes don't make any sense without accent swings, right?
If so, I think this entire if statement should be moved up to inside the "Accent swings detected" code block.

sound/hybrid_font.h
```diff
@@ -138,25 +138,48 @@ class HybridFont : public SaberBase {
...
-        if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
-          swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
-          swing_player_->FadeAndStop();
-          swing_player_.Free();
-          swing_player_ = PlayPolyphonic(&swng);
+
+  void StartSwing(const Vec3& gyro, float swingThreshold, float slashThreshold) override {
+    Vec3 gyro_slope = fusor.gyro_slope();
+    //Radians per second per second
+    float rss = sqrtf(gyro_slope.z * gyro_slope.z + gyro_slope.y * gyro_slope.y) * (M_PI/180);
```
profezzorn: space before and after /

sound/hybrid_font.h
```diff
@@ -138,25 +138,48 @@ class HybridFont : public SaberBase {
...
-        // between 0 (full overlap) and 1.0 (no overlap)
-        if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
-          swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
-          swing_player_->FadeAndStop();
-          swing_player_.Free();
-          swing_player_ = PlayPolyphonic(&swng);
+
+  void StartSwing(const Vec3& gyro, float swingThreshold, float slashThreshold) override {
+    Vec3 gyro_slope = fusor.gyro_slope();
+    //Radians per second per second
```
profezzorn: space after //

sound/hybrid_font.h
```diff
@@ -138,25 +138,48 @@ class HybridFont : public SaberBase {
...
+          if (swing_player_->pos() / swing_player_->length() >= config_.ProffieOSSwingOverlap) {
+            swing_player_->set_fade_time(swing_player_->length() - swing_player_->pos());
+            swing_player_->FadeAndStop();
+            swing_player_.Free();
+          }
         }
+        if (!swing_player_) {
+          if (!swinging_) {
+            if (rss > slashThreshold && slsh.files_found()) {
+              swing_player_ = PlayPolyphonic(&slsh);
```
profezzorn: before we check this in; there is an alternate way that may potentially be interesting....
Basically, we could pick which swing sound to play based on the slope.
So instead of having separate swings and slashes, we would need a progression of swing sounds, and we would pick a sound based on how "aggressive" the swings are.
WDYT?

sound/hybrid_font.h
```diff
@@ -400,26 +423,13 @@ class HybridFont : public SaberBase {
...
-      }
-      float swing_strength = std::min<float>(1.0, speed / config_.ProffieOSSwingSpeedThreshold);
-      SetSwingVolume(swing_strength, 1.0);
-    } else if (swinging_ && speed <= config_.ProffieOSSwingSpeedThreshold) {
-      swinging_ = false;
-    }
-    float vol = 1.0f;
-    if (!swinging_) {
-      vol = vol * (0.99 + clamp(speed/200.0, 0.0, 2.3));
+    if (state_ != STATE_OFF &&
```
profezzorn: This and the next line can fit on one line.

sound/smooth_swing_config.h
```diff
@@ -13,6 +13,7 @@ class SmoothSwingConfigFile : public ConfigFile {
     CONFIG_VARIABLE(Transition2Degrees, 160.0f);
     CONFIG_VARIABLE(MaxSwingVolume, 3.0f);
     CONFIG_VARIABLE(AccentSwingSpeedThreshold, 0.0f);
+    CONFIG_VARIABLE(AccentSlashAccelerationThreshold, 0.0f);
```
profezzorn: Should we default this to on?

sound/smooth_swing_v2.h
```diff
@@ -113,13 +122,13 @@ class SmoothSwingV2 : public SaberBasePassThrough {
           break;
         }
         state_ = SwingState::ON;
-        
+
       case SwingState::ON:
         // trigger accent swing
-        if (speed >=smooth_swing_config.AccentSwingSpeedThreshold &&
-            accent_swings_present &&
+        if (accent_swings_present &&
```
profezzorn: this and next line can be one line

sound/hybrid_font.h
SA-22C: That is an interesting notion.  It brings to mind another idea that I've had in my head for a while but never really tried to implement: a third looped hum.  Rather than having a pair that transitions from A - B, this third hum will increase in volume (and pitch) as the saber swings faster to provide an overlay that, if it works like I think it might, could potentially do away with static sounds altogether.

In any event, there are plenty of fonts that exist right now that can take advantage of slsh so I'd like to get this checked in as is and, after I do a couple other things, come back to thinking about how we either: a) implement a progressive set of static sounds or b) look at another looped sound that reacts differently enough to the movement of the saber that it can function as an accent that changes enough that we don't need static sounds.

sound/smooth_swing_config.h
SA-22C: If we do, we should default accents to on too, no?

sound/hybrid_font.h
```diff
@@ -8,11 +9,12 @@ class IgniterConfigFile : public ConfigFile {
     CONFIG_VARIABLE(volHum, 15);
     CONFIG_VARIABLE(volEff, 16);
     CONFIG_VARIABLE(ProffieOSSwingSpeedThreshold, 250.0f);
-    CONFIG_VARIABLE(ProffieOSSwingVolumeSharpness, 1.0f);
-    CONFIG_VARIABLE(ProffieOSMaxSwingVolume, 3.0f);
+    CONFIG_VARIABLE(ProffieOSSwingVolumeSharpness, 0.5f);
```
SA-22C: Updated defaults for accent swings to tested values that produce a much better sound for fonts that don't have values set in config.ini

profezzorn: This is something that we should try to avoid doing as much as possible.
Changing the defaults means that we may cause problems for some people.
Are you SURE these new values are better?

sound/hybrid_font.h
profezzorn: sounds good to me.

sound/smooth_swing_config.h
profezzorn: no
I mean that the can default to on IF accent swings are on.

sound/hybrid_font.h
SA-22C: yes, did quite a bit of testing with Dmitry and Jesse Kirkbride.

sound/smooth_swing_config.h
SA-22C: ok, set default value to 260, so if accent swing threshold is set in smoothsw.ini, slashes will default to on as well, provided there are slsh files present.

PR #23, opened 2019-10-16
Save current preset

SA-22C:
Config file flag SAVED_PRESET and functions in current_preset.h and prop_base.h to save the current preset (and volume level) to savedpreset.ini on the SD card and then reference that file when booting so that the saber boots to the last used preset.


common/current_preset.h
```diff
@@ -271,10 +271,30 @@ class CurrentPreset {
...
     Clear();
+    #ifdef SAVED_PRESET
+    SavePreset(preset);
+    #endif
     LOCK_SD(true);
     if (!Load(preset)) Set(preset);
     LOCK_SD(false);
   }
+
+  bool SavePreset(int preset) {
```
profezzorn: This doesn't actually save the preset, but it saves what preset we are currently on.
Also, it saves the volume.
I don't think we should take a number as argument here, as we can get the preset number from the current preset.

props/prop_base.h
```diff
@@ -293,6 +297,36 @@ class PropBase : CommandParser, Looper, protected SaberBase {
 #endif    
   }
 
+  void ResumePreset() {
+    FileReader f;
+    int newPreset;
+    int newVolume;
+    if (f.Open("savedpreset.ini") && f.FileSize() > 0) {
```
profezzorn: The code that reads the preset file and the code that writes the preset file should be in the same file.

common/current_preset.h
profezzorn: Maybe it would be better to add this to presets.ini?

props/prop_base.h
```diff
@@ -293,6 +297,36 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+        if (!strcmp(variable, "preset")) {
+          newPreset = f.readIntValue();
+        continue;
+        }
+        if (!strcmp(variable, "volume")) {
+          newVolume = f.readIntValue();
+        }
+      }
+      f.Close();
+      SetPreset(newPreset, false);
```
profezzorn: So does this mean that we will save the preset right after reading it from the file?
If so, that means a lot of extra writing, which can wear out the SD card.

common/current_preset.h
```diff
@@ -271,10 +271,30 @@ class CurrentPreset {
...
+    if (preset >= 0 && preset <= current_config->num_presets) {
+      STDOUT.println("Saving Current Preset");
+      LOCK_SD(true);
+      FileReader c;
+      LSFS::Remove("savedpreset.ini");
+      c.Create("savedpreset.ini");
+      char value[30];
+      itoa(preset, value, 10);
+      c.write_key_value("preset", value);
```
profezzorn: In the TD code, this is actually done with "rotate_presets", which simply re-orders the presets in presets.ini and saves it.
Not sure which way is better, but I wanted to point out the option at least.

common/current_preset.h
```diff
@@ -271,10 +271,30 @@ class CurrentPreset {
 
   void SetPreset(int preset) {
     Clear();
+    #ifdef SAVED_PRESET
```
profezzorn: I don't like using #define SAVED_PRESET for this, it's not not descriptive enough.
Maybe something like:

#define SAVE_STATE
or
#define SAVE_CURRENT_STATE
or
#define SAVE_PRESET_AND_VOLUME

or, we could make it two defines:
#define SAVE_ACTIVE_PRESET
#define SAVE_VOLUME

common/current_preset.h
```diff
@@ -271,10 +271,30 @@ class CurrentPreset {
...
+      STDOUT.println("Saving Current Preset");
+      LOCK_SD(true);
+      FileReader c;
+      LSFS::Remove("savedpreset.ini");
+      c.Create("savedpreset.ini");
+      char value[30];
+      itoa(preset, value, 10);
+      c.write_key_value("preset", value);
+      itoa(dynamic_mixer.get_volume(), value, 10);
+      c.write_key_value("volume", value);
```
profezzorn: Not sure if it makes sense to save the volume if there is no way to change it...

common/current_preset.h
```diff
@@ -271,10 +271,30 @@ class CurrentPreset {
...
     LOCK_SD(true);
     if (!Load(preset)) Set(preset);
     LOCK_SD(false);
   }
+
+  bool SavePreset(int preset) {
+    if (preset >= 0 && preset <= current_config->num_presets) {
+      STDOUT.println("Saving Current Preset");
+      LOCK_SD(true);
+      FileReader c;
```
profezzorn: what does "c" stand for here?

common/current_preset.h
```diff
@@ -271,10 +271,30 @@ class CurrentPreset {
...
+    #ifdef SAVED_PRESET
+    SavePreset(preset);
+    #endif
     LOCK_SD(true);
     if (!Load(preset)) Set(preset);
     LOCK_SD(false);
   }
+
+  bool SavePreset(int preset) {
+    if (preset >= 0 && preset <= current_config->num_presets) {
```
profezzorn: When is this if statement helpful?

props/prop_base.h
profezzorn: Is the SD locked here?

common/current_preset.h
SA-22C: Yes, but once this hopefully gets merged, was going to propose a volume menu.

profezzorn: I'm not a big fan of menues, but as long as it can be enabled/disabled with defines that might be ok.

common/current_preset.h
```diff
@@ -271,10 +271,30 @@ class CurrentPreset {
...
+      STDOUT.println("Saving Current Preset");
+      LOCK_SD(true);
+      FileReader c;
+      LSFS::Remove("savedpreset.ini");
```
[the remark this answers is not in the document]
SA-22C: Ok, but for presets.ini and presets.tmp, we're updating one section of a larger file.  For this function we're writing two key values, replacing the file structure each time.  In this circumstance, why does writing a tmp file and then confirming the tmp file contents need to occur?

props/prop_base.h
```diff
@@ -293,6 +297,36 @@ class PropBase : CommandParser, Looper, protected SaberBase {
 #endif    
   }
 
+  void ResumePreset() {
```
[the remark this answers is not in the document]
SA-22C: Not sure what you mean here.  I'm using the same key-value pairs that are seen elsewhere in the file-writing code.

common/current_preset.h
SA-22C: ok, moving to SAVE_STATE

common/current_preset.h
SA-22C: renamed to f

common/current_preset.h
SA-22C: Removed, remnant from old way of calling that function

props/prop_base.h
SA-22C: fixed

SA-22C: moved savestate function to prop_base.h

common/current_preset.h
SA-22C: I think it's better to keep it separate because it's doing a different job.  Save State is a bookmark to where the saber was when it was powered off last.

common/current_preset.h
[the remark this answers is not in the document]
SA-22C: renamed to curstate.ini

SA-22C:
moved the code to prop_base.h in its entirety, actually makes more logical sense since the save state is a prop function, not a preset function.

props/prop_base.h
SA-22C: we need to save the state each time the font is changed.  I've moved the savestate (renamed savepreset) function to the next_ and prev_preset functions in prop_base.h, so I don't think i'm doing more than the required read/writes to make this work.

props/prop_base.h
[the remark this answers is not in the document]
SA-22C: resolved, using config.h.

common/current_preset.h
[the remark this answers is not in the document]
SA-22C: resolved

props/prop_base.h
```diff
@@ -300,15 +316,103 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+    if (!isSpace(f->Read())) return false;
+    if (toLower(f->Read()) != 'e') return false;
+    if (toLower(f->Read()) != 'n') return false;
+    if (toLower(f->Read()) != 'd') return false;
+    f->Seek(0);
+
+    return true;
+  }
+
+  bool SaveState(int preset) {
```
profezzorn: I think too much stuff got copied from current_preset.h
For this file, I think all we really need is:

   save to curstate.tmp
   save to curstate.ini

If reading the ini file fails (or if the "end" is missing), we need to read the tmp file.
Sorry for making this so confusing.

Also, if there is code that we can re-use generalize and use in both places, we should do that as that uses less memory than simply copying and modifying.

props/prop_base.h
```diff
@@ -300,15 +316,103 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+	}
+      }
+      f2.Close();
+      f.Close();
+      return true;
+    }
+    return false;
+  }
+
+    bool CreateINI() {
```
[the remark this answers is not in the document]
profezzorn: Also, I don't think we need it.

props/prop_base.h
```diff
@@ -1,6 +1,16 @@
 #ifndef PROPS_PROP_BASE_H
 #define PROPS_PROP_BASE_H
 
+class SaveStateFile : public ConfigFile {
+public:
+  void SetVariable(const char* variable, float v) override {
+    CONFIG_VARIABLE(preset, 0);
+    CONFIG_VARIABLE(volume, VOLUME);
```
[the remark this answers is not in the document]
SA-22C: I set it to 0 so I can use the same set preset call regardless of the result of reading the savestate file.

props/prop_base.h
```diff
@@ -1159,12 +1264,13 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
   }
-  
+
   virtual bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) = 0;
-  
+
 protected:
   CurrentPreset current_preset_;
   LoopCounter accel_loop_counter_;
+  SaveStateFile savestate_;
```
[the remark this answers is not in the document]
SA-22C: resolved

props/prop_base.h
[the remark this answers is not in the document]
SA-22C: removed

props/prop_base.h
```diff
@@ -300,15 +316,103 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
     STDOUT.println("BAD BLADE");
 #ifdef ENABLE_AUDIO
     talkie.Say(talkie_error_in_15, 15);
     talkie.Say(talkie_blade_array_15, 15);
-#endif    
+#endif
+  }
+
+  void ResumePreset() {
+    savestate_.ReadInCurrentDir("curstate.ini");
```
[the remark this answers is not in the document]
SA-22C: resolved

props/prop_base.h
```diff
@@ -222,19 +232,25 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
-    
+#endif
+
     if (on) On();
     if (announce) SaberBase::DoNewFont();
   }
 
   // Go to the next Preset.
   virtual void next_preset() {
+    #ifdef SAVE_STATE
```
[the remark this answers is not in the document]
SA-22C: fixed

props/prop_base.h
SA-22C: ok, pulled out the createini stuff and put a couple checks on the save state to try and open the ini, then the tmp.  Not sure about any code generalization right now but I can see a much larger effort to combine all the current preset and config file code into one base that reads and writes to generic ini files with key-pair variables.

props/prop_base.h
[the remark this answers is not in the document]
SA-22C: I set it to 0 so I can use the same setpreset call regardless of the outcome of the read-in of curstate.ini / tmp.

SA-22C:
I know it's not related but I just uploaded my changes to support font_config.txt in CFX fonts

sound/smooth_swing_cfx_config.h
```diff
@@ -0,0 +1,31 @@
+#ifndef SOUND_SMOOTH_SWING_CFX_CONFIG_H
```
profezzorn: Please make a separate pull request for this.

SA-22C: pull request #25 opened.

props/prop_base.h
```diff
@@ -296,15 +312,94 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+    if (!OpenState(&f, "curstate.ini")) {
+      if(!OpenState(&f, "curstate.tmp")) {
+      } else {
+        savestate_.Read("curstate.tmp");
+      }
+    } else { 
+      savestate_.Read("curstate.ini");
+    }
+    SetPreset(savestate_.preset, false);
+    if (savestate_.volume <= VOLUME) {
```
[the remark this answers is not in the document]
SA-22C: I've tested it with volumes less than max volume and it works for me.  'default' is the max volume specified by the VOLUME define in the config file and the preset volume can be any value less than (or equal to max volume)

props/prop_base.h
SA-22C: removed update ini and readstate, changed config_file.h to return a bool if it can't open a file.

common/config_file.h
```diff
@@ -6,9 +6,9 @@
 // Reads an config file, looking for variable assignments.
 // TODO(hubbe): Read config files from serialflash.
 struct ConfigFile {
-  void Read(FileReader* f) {
+  bool Read(FileReader* f) {
     SetVariable("=", 0.0);  // This resets all variables.
-    if (!f || !f->IsOpen()) return;
+    if (!f || !f->IsOpen()) return false;
```
profezzorn: Good start, but not enough.
We need to know if the file is truncated or not, so I think we need to also check that there is an "end" at the end of the file.

Might be a good idea to return an enum with three values, something like:

CONFIG_FILE_READ_FAIL
CONFIG_FILE_READ_OK
CONFIG_FILE_READ_OK_WITH_END

Since hand-written config files aren't expected to have "end" at the end, only automatically generated files will have that.

props/prop_base.h
```diff
@@ -296,15 +312,51 @@ class PropBase : CommandParser, Looper, protected SaberBase {
   } while(0);
 
     ONCEPERBLADE(ACTIVATE);
+    #ifdef SAVE_STATE
```
profezzorn: no indentation for preprocessor statements

props/prop_base.h
```diff
@@ -296,15 +312,51 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+  }
+
+  void ResumePreset() {
+    FileReader f;
+    SaveStateFile savestate;
+    if (!savestate.Read("curstate.ini")) {
+      savestate.Read("curstate.tmp");
+    }
+    SetPreset(savestate.preset, false);
+    if (savestate.volume <= VOLUME) {
```
profezzorn: This still doesn't work if:

curstate.ini has the volume set to VOLUME
the dynamic mixer volume is less than VOLUME

The code will assume that the savefile doesn't have a volume in it and not set the volume, so the volume will remain low.

common/config_file.h
profezzorn: It occured to me at lunch that there is a much easier way to do this.
We could make "end=1" be the end marker, then add an "end" variable to SaveStateFile, and default it to zero.
Then savefile.end will tell us if the file was read successfully or not.

PR #24, opened 2019-10-30
Initial Blaster Support

adays1234:
Creating a Pull request for initial blaster changes just to get dialogue going.

I know blaster.h needs some work.


props/blaster.h
```diff
@@ -0,0 +1,192 @@
...
+#include "prop_base.h"
+
+#define PROP_TYPE Blaster
+
+class Blaster : public PropBase {
+public:
+  Blaster() : PropBase() {}
+  const char* name() override { return "Blaster"; }
+
+#ifdef DELAYED_OFF
```
profezzorn: What is DELAYED_OFF for?

props/blaster.h
```diff
@@ -0,0 +1,192 @@
...
+      case EVENTID(BUTTON_POWER, EVENT_DOUBLE_CLICK, MODE_ON):
+        StartOrStopTrack();
+        return true;
+
+      case EVENTID(BUTTON_AUX, EVENT_PRESSED, MODE_ON):
+        Fire();
+        return true;
+
+      case EVENTID(BUTTON_AUX, EVENT_RELEASED, MODE_ON):
+        if (blaster_mode == MODE_AUTO) SaberBase::DoEndLockup();
```
profezzorn: What happens if the mode changes between pressing and releasing AUX?

props/blaster.h
```diff
@@ -0,0 +1,192 @@
...
+  // Mode states to handle kill vs stun effects
+  enum BlasterMode {
+    MODE_STUN,
+    MODE_KILL,
+    MODE_AUTO
+  };
+
+  BlasterMode blaster_mode = MODE_STUN;
+
+  void SetBlasterMode(BlasterMode to_mode) {
```
profezzorn: make it virtual to make it easy for people to inherit and extend, same for nextblastermode and fire

props/blaster.h
```diff
@@ -0,0 +1,192 @@
...
+      case EVENTID(BUTTON_AUX, EVENT_PRESSED, MODE_ON):
+        Fire();
+        return true;
+
+      case EVENTID(BUTTON_AUX, EVENT_RELEASED, MODE_ON):
+        if (blaster_mode == MODE_AUTO) SaberBase::DoEndLockup();
+        return true;
+
+      case EVENTID(BUTTON_AUX2, EVENT_PRESSED, MODE_ON):
+        beginArm();
```
profezzorn: detonators and phases behave differently though.
On a detonator, there is a button, and as long as you're holding it, everything is safe.
Once you release a button, a 4-second countdown starts and then BOOM.

For phasers, you can set it for overload, and it starts making a noise, which becomes louder and louder until it explodes, which can take anything from 20 seconds to a few minutes depending on phaser model I think. However, at any time, it is possible to undo the overload and make it not explode.

On the STNG, overload is apparently done by pressing a few buttons on the back of the phaser. But I like the method they use in the ST pilot better, where they twist the barrel to select different modes, and and if you twist it enough (or maybe leave it between modes) the overload begins.

sound/effect.h
```diff
@@ -465,6 +465,26 @@ EFFECT(ccbegin);
 EFFECT(ccend);
 EFFECT(ccchange);
 
+// Blaster effects
```
profezzorn: A lot less overlap than I thought.
We really do need to figure out a way to only enable the effects that we need, but we don't need to do that in this CL.

props/blaster.h
adays1234: Delayed off was carried over from TD since I wanted to preserve self destruct

config/blaster_v3_config.h
```diff
@@ -0,0 +1,38 @@
...
+#include "v3_config.h"
+#define NUM_BLADES 2
+#define NUM_BUTTONS 2
+#define VOLUME 2000
+const unsigned int maxLedsPerStrip = 144;
+#define CLASH_THRESHOLD_G 1.0
+#define ENABLE_AUDIO
+#define ENABLE_WS2811
+#define ENABLE_SD
+#define ENABLE_POWER_FOR_ID PowerPINS<bladePowerPin1, bladePowerPin2, bladePowerPin3>
```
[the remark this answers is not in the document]
adays1234: Hah. Ok. That was my bad. I began with a clone of my saber config. Will fix

props/blaster.h
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+
+  void NextBlasterMode() {
+    switch(blaster_mode) {
+      case MODE_STUN:
+        SetBlasterMode(MODE_KILL);
+        return;
+      case MODE_KILL:
+#ifdef ENABLE_AUTO
```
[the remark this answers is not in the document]
adays1234: Makes sense. Will change

config/blaster_v3_config.h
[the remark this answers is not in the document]
adays1234: Though I could make the argument for detachable barrels. 😝

props/blaster.h
adays1234: Should the self destruct sequence be a mode like kill, stun, auto then maybe?

profezzorn: I think that makes sense.

adays1234: Ok. I can do that. 👍

profezzorn:
Do you think we should try to put this in ProffieOS 3.x?
If so, I would want to merge this within a few days or so.

adays1234:
I would love to. Most of your suggestions are super easy.

Couple of them I need to think through. Will try to get them all done for an update in the next day or so.

config/blaster_v3_config.h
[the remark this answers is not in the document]
adays1234: Fixed. removed ID pins, and added ENABLE_BLASTER_AUTO define.

props/blaster.h
adays1234: didn't change this until I revisit the self-destruct.

props/blaster.h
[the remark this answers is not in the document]
adays1234: Fixed.

props/blaster.h
adays1234: Fixed. Would appreciat thoughts.. what I did was add a firing_ bool and don't allow mode changes during the firing sequence.

props/blaster.h
adays1234: done.

props/blaster.h
adays1234: So this is the one that is going to require some thought and work.. do we want to merge as-is, or postpone until I rework the entire self-destruct sequence?

adays1234:
Awww bugger. How the heck did my personal Mk1 config get in here? Ugh. Sometimes I fail to understand why merges do that... I need to figure out how to roll that back.

props/blaster.h
profezzorn: In that case, maybe this if statement should check if (firing_) ?

props/blaster.h
profezzorn: Either way is fine with me I think.
Probably the best thing to do would be to merge this soon, and put it in 3.x, and if you have the time, or the 3.x release drags out for some reason, you can submit updates. Otherwise it 3.x would have the code the way it is, which I think is ok.

config/mk1.h
```diff
@@ -0,0 +1,75 @@
+#ifdef CONFIG_TOP
```
profezzorn: Let me know when you get rid of this file, then maybe we can do the merge?

adays1234: Totally fine. So my merge was in the wrong direction between my branches resulting in this. I need to figure out how to revert a multi parent merge. 😳

Worst case I will close this Pr, cherry pick commits for a new branch and reopen the PR under a new branch.

profezzorn: You can also just commit a change on top of the merge that resolves the problem.

profezzorn: I will flatten the change anyway when I merge it, so it doesn't matter if the history is clean or not. :)

adays1234: Ok well in that case I will clean up my own mess later and for now just deleted the file from the branch.

props/blaster.h
adays1234: Well, we need to check the mode being auto. The firing_ set to true happens in the Fire() function and set to false on release.

What I wound up doing is NOT allowing a mode change if firing_

adays1234:
I'll also take another full blown pass at this this coming weekend.

I plan to make the firing_ lock / mode sequence more elegant.
Also will make self destruct a lockup and work out a better destruct sequence.
Also plan to start the phaser prop with strength levels like we discussed

profezzorn:
> I'll also take another full blown pass at this this coming weekend.
> 
> I plan to make the firing_ lock / mode sequence more elegant.
> Also will make self destruct a lockup and work out a better destruct sequence.
> Also plan to start the phaser prop with strength levels like we discussed

So do you think I should merge this in the meantime?
I don't think it will make a big difference either way if I do.

adays1234:
I would if you feel comfortable. People can play with the basics and start building props (such as megtooth)... it just won't be quite as fancy and clean yet. Can always organize that later.

I have tested this code and it works for basic blaster setups... so no harm in doing so.

It allows you to get the support into 3.x and we can iterate from there (which I will be doing).

I mostly wanted to give you a sense that I'm not done fully with tinkering with this yet and also give you a sense of what is coming next.

PR #25, opened 2019-10-30
Cfx font values

SA-22C:
add support to read CFX values from font_config.txt in the font folder.


profezzorn:
Looks good to me. :)

PR #29, opened 2019-12-05
fix issue with multiple blade detection on ignition when blade is in

ishtob:


profezzorn:
Seems ok, but I'm not sure why this fixes anything?
Can you explain?

ishtob:
unsure either, but now not getting blade change trigger when the blade is plugged in while the blade is on, though serial is putting out Select_ON / Select_OFF (btw, think this naming should be changed for the debug readout)

ishtob:
so reverted the change, started playing with the time floating, and setting it to 100 seems to fix tne problem without loosing the detection triggering while blade is on

ishtob:
vote for floating latch implemented, making this PR outdated. closing

PR #30, opened 2019-12-08
Immediate on when kill key / switch is enabled to on. Finite blaster shots implemented. Corrected hum start. New button names, reload and clipin clipout implemented 

adays1234:


common/events.h
```diff
@@ -14,9 +14,10 @@ enum BUTTON : uint32_t {
   // D-pad and FIRE/RELOAD/CLIP_DETECT have the same numbers,
```
profezzorn: Update comment too

props/blaster.h
```diff
@@ -61,9 +51,25 @@ class Blaster : public PropBase {
   }
 
   bool auto_firing_ = false;
+  int shots_fired_ = 0;
+
+#ifdef BLASTER_SHOTS_UNTIL_EMPTY
+  const int max_shots_ = BLASTER_SHOTS_UNTIL_EMPTY;
+#else
+  const int max_shots_ = -1;
```
profezzorn: Defaulting to a billion might make things simpler...

props/blaster.h
```diff
@@ -77,10 +83,53 @@ class Blaster : public PropBase {
...
 #endif
     }
   }
 
+  virtual void Reload() {
+    shots_fired_ = 0;
+
+#ifdef ENABLE_AUDIO
+    if (SFX_reload.files_found()) {
+      hybrid_font.PlayCommon(&SFX_reload);
```
profezzorn: Calling reload here is known as a "layer violation".
Basically it means that you're doing something in the wrong place.
It's supposed to be up to the hybrid font how to handle a reload event.
I'm ok with checking this in for now, because it's better than not handling reload at all, but the problem with this is that no events is sent to the blades, so it's not possible to add any effects to a reload events.

props/blaster.h
```diff
@@ -77,10 +83,53 @@ class Blaster : public PropBase {
...
+  // Pull in parent's SetPreset, but turn the blaster on.
+  virtual void SetPreset(int preset_num, bool announce) override {
+    PropBase::SetPreset(preset_num, announce);
+
+    if (!SaberBase::IsOn()) {
+      On();
+    }
+
+#ifdef ENABLE_AUDIO
+    hybrid_font.SB_On();
```
profezzorn: Already handled by On().

props/blaster.h
```diff
@@ -77,10 +83,53 @@ class Blaster : public PropBase {
...
+      hybrid_font.PlayCommon(&SFX_clipin);
+    }
+#endif
+  }
+
+  // Pull in parent's SetPreset, but turn the blaster on.
+  virtual void SetPreset(int preset_num, bool announce) override {
+    PropBase::SetPreset(preset_num, announce);
+
+    if (!SaberBase::IsOn()) {
```
profezzorn: I think we should only do this if there is no "poweron" sound in the font, what do you think?

props/blaster.h
profezzorn: Still think it's called by On():

https://github.com/profezzorn/ProffieOS/blob/4c2f05f90ddd057ee38d7a500c741b4ebd666a8a/common/saber_base.h#L49
https://github.com/profezzorn/ProffieOS/blob/4c2f05f90ddd057ee38d7a500c741b4ebd666a8a/props/prop_base.h#L88

Also SB_On() should only be called by SaberBase::DoOn();

props/blaster.h
profezzorn: It's easy to add an SFX_poweron check here though.

adays1234: Ah so you mean wrap everything after the parent's SetPreset inside an if (!SFX_poweron) Check. Got it.

adays1234:
PR updated with latest revisions from fork. Tested.

blades/abstract_blade.h
```diff
@@ -109,6 +111,47 @@ class AbstractBlade : public BladeBase, public SaberBase {
     return GetPrimaryBlade() == this;
   }
 
+  // Blaster Effects
+  void SB_Stun() override {
+    addEffect(EFFECT_STUN, (200 + random(700)) / 1000.0f);
```
profezzorn: It doesn't make sense for the "location" of the stun effect to be random, so just make it zero I think. (Same for all of the effects below.)

blades/abstract_blade.h
```diff
@@ -109,6 +111,47 @@ class AbstractBlade : public BladeBase, public SaberBase {
...
+  void SB_Mode() override {
+    addEffect(EFFECT_MODE, (200 + random(700)) / 1000.0f);
+  }
+  void SB_Range() override {
+    addEffect(EFFECT_RANGE, (200 + random(700)) / 1000.0f);
+  }
+  void SB_Empty() override {
+    addEffect(EFFECT_EMPTY, (200 + random(700)) / 1000.0f);
+  }
+  void SB_Full() override {
```
profezzorn: I get SB_Empty() (click when you try to fire with an empty clip) But what is this for?

blades/abstract_blade.h
```diff
@@ -109,6 +111,47 @@ class AbstractBlade : public BladeBase, public SaberBase {
     return GetPrimaryBlade() == this;
   }
 
+  // Blaster Effects
```
profezzorn: This long list of identical functions makes me think that there has to be a better way to do this.
Probably outside the scope of this change though...

props/blaster.h
```diff
@@ -61,23 +45,74 @@ class Blaster : public PropBase {
...
+        pairSelection = rand() % autoCount;
+        SFX_auto.Select(pairSelection);
+        SFX_blast.Select(pairSelection);
+    }
+  }
+
+  virtual void Reload() {
+    shots_fired_ = 0;
+    SaberBase::DoReload();
+    SaberBase::DoFull();
```
profezzorn: This seems redundant.
Why not just have DoReload() play both sounds?

sound/hybrid_font.h
```diff
@@ -119,8 +119,12 @@ class HybridFont : public SaberBase {
     }
     hum_player_ = next_hum_player_;
     next_hum_player_.Free();
-    hum_player_->PlayOnce(f);
-    current_effect_length_ = hum_player_->length();
+
+    if (f) {
```
profezzorn: I'm confused, what is this supposed to do?

sound/hybrid_font.h
```diff
@@ -300,6 +304,22 @@ class HybridFont : public SaberBase {
   void SB_Force() override { PlayCommon(&SFX_force); }
   void SB_Blast() override { Play(&SFX_blaster, &SFX_blst); }
   void SB_Boot() override { PlayPolyphonic(&SFX_boot); }
+
+  // Blaster effects, auto fire is handled by begin/end lockup
+  void SB_Stun() override { PlayPolyphonic(&SFX_stun); }
```
profezzorn: Shouldn't this (and all the ones below) be PlayCommon instead of PlayPolyphonic?

blades/blade_base.h
```diff
@@ -16,6 +16,21 @@ enum BladeEffectType {
   EFFECT_IGNITION = 1 << 9,
   EFFECT_RETRACTION = 1 << 10,
   EFFECT_CHANGE = 1 << 11, // used for click to change
+
+  // Blaster Effect Types
+  EFFECT_STUN = 1 << 12, // used to pass STUN to allow alternate colors for stun type blasts
```
profezzorn: Lots of bits, but I suppose we could potentially share them between props if needed, kind of like we share bits between D-pads and blaster buttons.

sound/hybrid_font.h
```diff
@@ -300,6 +304,22 @@ class HybridFont : public SaberBase {
...
+  void SB_Fire() override { PlayPolyphonic(&SFX_blast); }
+  void SB_ClipIn() override { PlayPolyphonic(&SFX_clipin); }
+  void SB_ClipOut() override { PlayPolyphonic(&SFX_clipout); }
+  void SB_Reload() override { PlayPolyphonic(&SFX_reload); }
+  void SB_Mode() override { PlayPolyphonic(&SFX_mode); }
```
profezzorn: I suggest making some sort of fallback here in case "mode.wav" is not available.
Maybe use the beeper? (Or maybe talkie?)

props/blaster.h
```diff
@@ -61,23 +45,74 @@ class Blaster : public PropBase {
...
   virtual void Fire() {
-    if (!SaberBase::IsOn()) return; // Don't allow firing when off.
+    if (max_shots_ != -1) {
+      if (shots_fired_ >= max_shots_) {
+        SaberBase::DoEmpty();
+        return;
+      }
+    }
 
     if (blaster_mode == MODE_AUTO) {
```
profezzorn: Random thought: we could specify the chance of jamming in the config file.
Then we could use clash or shake gestures to trigger un-jamming. :)

blades/abstract_blade.h
adays1234: I'm not sure... it is in the blaster core font. I think maybe some sort of power gauge going up? I wanted a hook to it since it is in the font but it is not currently implemented in the prop itself.

blades/abstract_blade.h
adays1234: Yeah I figured based on the growing list I saw this going the way of SABERFUN but was not about to make the assumption of doing it.

props/blaster.h
adays1234: I'm not entirely sure where this is supposed to happen but I wanted different effects for the different font files so they could be styled independently and still use the base approach of the SaberBase created functions.

sound/hybrid_font.h
adays1234: So THIS is what fixed the hum. Without a power sound it was trying to play a nullptr and throwing weird stuff in serial monitor. This just ensured it was not trying to play a non existent file and allowed hum to start properly. It is a file existence safety check to make sure a poweron or out exists before trying to play it because in blasters they don't

sound/hybrid_font.h
adays1234: Ah yep. My bad. Will fix

blades/blade_base.h
adays1234: Do you have a preference as to which I overlap with?

blades/abstract_blade.h
adays1234: Will fix. I copied blast for these. Good catch.

sound/hybrid_font.h
adays1234: Yeah I can do that. Honestly I had not figured out all the hex stuff of talkie yet but want to since I would actually like it to "say" what mode it is in along with a mode sound anyway. But I am a little lost on how to use Talkie

props/blaster.h
adays1234: Already thought of that and was gonna add it but wanted to get this in first. But that is exactly how I wanted Jam/Unjam to work. Random chance to create a jam and shake to unjam.

adays1234:
Changes made. Basic jam / unjam on clash. Shake is not working correctly but probably because I am missing some of the swing stuff so it's probably not picking up the strokes?

blades/blade_base.h
profezzorn: No need to do any overlap yet.
I think we'll cross that bridge when we come to it.

props/blaster.h
profezzorn: If we don't know what it's supposed to do, it's probably better to not trigger that effect at all for now.
Some fonts will use it, and if it's triggered at the wrong time, it will sound weird.

props/blaster.h
profezzorn: yes, please do.

sound/hybrid_font.h
profezzorn: Hmm, so when was it trying to play a null? During lockup handling? Or somewhere else?
I'm trying to make sure that this is the right fix and not just hiding a problem.

props/blaster.h
adays1234: Does the SFX_poweron Check still even need to happen? With the fix inside PlayMonophonic I am just calling On() now.

props/blaster.h
adays1234: Yeah ok. I'll pull the call to it but leave the capability for it there? So once we figure out how it is actually supposed to be used I can add the correct behavior in the blaster prop.

sound/hybrid_font.h
adays1234: SB_On was calling PlayMonophonic(NULL, &SFX_hum).

Because PlayMonophonic takes two effect arguments but only the second one had a check to see if it actually was there. So it assumed the first argument was always a properly formed Effect pointer and in the case of blaster it is not.

blades/blade_base.h
adays1234: Ok

props/blaster.h
profezzorn: It would make it possible to provide an poweron.wav / poweroff.wav and wire up a power button, which seems like it might be nice?

sound/hybrid_font.h
profezzorn: SB_On() does not call PlayMonophonic(NULL, &SFX_hum);
It calls PlayMonophonic(&SFX_poweron, &SFX_hum);
Now, SFX_poweron might not have any files in it, but that's still different from a null pointer.
Importantly, the check you have here tests for null pointers, not pointers to effects without files, so I don't think this if statement does anything in SB_On(). It might do something in other places where PlayMonophonic is might actually be called with a null pointer, but that seems like it is not what you intended...

props/blaster.h
profezzorn: Btw, shake is a new gesture, so it's possible it just doesn't work very well.
You can use "monitor strokes" to see if it is working or not.

sound/hybrid_font.h
adays1234: Yes I know the call was not explicitly NULL. With this if check, hum starts correctly. Without it, serial monitor shows it trying to play a garbage file (looks like a memory address to something not correct) and hum doesn't start until a different effect is fired. So SOMETHING is not "just working" in an immediate call to On() with no poweron file.

props/blaster.h
adays1234: Ah. Yeah that makes sense.

I should add the explicit BUTTON_POWER to trigger On() too then.

props/blaster.h
```diff
@@ -152,51 +204,42 @@ class Blaster : public PropBase {
     PollNextAction();
   }
 
-  // Make clash do nothing
-  void Clash(bool stab) override {}
+  // Make clash do nothing except unjam if jammed.
+  void Clash(bool stab) override {
+    if( is_jammed_ ) {
```
profezzorn: if (is_jammied_) {

sound/hybrid_font.h
profezzorn: It's not enough that it works, we must also understand why.

The only way this actually does anything is if f is NULL. If f is not null, the code will behave the same as before the change. As far as I can tell, SB_On() does not call PlayMonophonic with f=NULL. The only places that may call PlayMonophonic() with f=NULL is SB_BeginLockup() and SB_EndLockup(), so if this code fixes the hum sound, then it must have something to do with the lockup functions.

props/blaster.h
profezzorn: I think so.

props/blaster.h
adays1234: Ah thanks. Yeah my IDE code autostyler sometimes tries to be too helpful.

sound/hybrid_font.h
adays1234: Ok. Let me back up and explain what I am seeing in the code.

hybrid_font. In Activate(). monophonic_hum_ is set by OR'ing a bunch of potential monophonic sounds: SFX_poweon, SFX_poweoff, etc. you added SFX_blast to this with the earlier font detection fix you made, which was good. However. In a blaster it is the only thing that returns true. So, monophonic_hum_ is now NOT null. 

Then we jump to SB_On() which checks monophonic_hum_ (it will pass this check and get inside the block). Inside that block it calls PlayMonophonic(&SFX_poweron, &SFX_hum). In blasters, the first argument is NOT assigned a proper Effect pointer.

Now. In PlayMonophonic, it just blindly calls Play() on the underlying player without any verification of the first argument. 

The underlying player gets Play() called on it without a properly formed Effect pointer. In fact, it could be any memory segment that has been cast to an Effect pointer because in this case nothing has actually assigned it or initialized it because poweron does not exist.

The bug here is that if a font is determined to be monophonic and monophonic_hum_ is assigned, then there is an unsafe and implicit assumption SFX_poweron is a properly formed Effect pointer.

The if (f) is trying to ensure that it is not a null. If it should not be an if (f) and should do something other than that, then I guess I fail to understand why things like if (SFX_<anything>) type of checks in the code elsewhere work. We could make it an if (f.files_found()) but i think the problem is not that the font is empty. It is the pointer had not been properly initialized because poweron is assumed to exist in this code path and it doesn't.

It has nothing to do with lockup.

sound/hybrid_font.h
adays1234: Added beep. I'll explore talkie when I get a chance

props/blaster.h
adays1234: I have triggered it with Clash and will circle back and try to add Shake but I think I have to as you say enable some of the gestures that may be turned off in the blaster due to the empty overrides (I had to add code back into the empty Clash override to make the clash work)

props/blaster.h
adays1234: Fixed.

props/blaster.h
adays1234: Done.

sound/hybrid_font.h
profezzorn: You're mixing up objects and pointers to objects, which is why there is some confusion here.
SFX_poweron is a global instance of the Effect class. It is a memory region, and it's existence or non-existence is not predicated by if any wav files are found or not. It always exists.

When we use SFX_poweron in an if statement, what's actually happening is that the function Effect::operator bool() gets called. It checks if any files were found and returns true or false based on that. The compiler will do this whenever we use any of the SFX_* objects as a boolean.

Because I don't want to make copies of Effect objects, all the methods that operate on Effects take a pointer to an Effect object as argument. (Like "Effect *effect")  Ampersand is used to get the pointer to the object, and that pointer will point to the object, regardless of whether it any wav files have been found or not.  When a pointer is used as a boolean, Effect::operator bool() is not called. Only null pointers are treated as false.

adays1234: Part of this is I am miscommunicating. Yes I know how pointers work and are dereferenced (it just has been almost 15 years since I wrote any significant amount of C/C++ code).

I also am aware that calling if on an object will use the bool() method on the class. And if on a pointer to the object will check if the pointer itself is null (null pointer).

So... then we need to understand why in the case of blaster is a pointer to SFX_poweron a null pointer when the poweron does not exist in the font.

How do I get to the Arduino equivalent of gdb? I want to start stepping through the code.

profezzorn: I only know how to run gdb on proffieboards on linux.  (Not sure how to do it on teensys)
You also need an st-link v2 hooked up to the debug pins on the board, then you start up then you follow this post:

http://therebelarmory.com/post/113056/thread

Although, nowadays it seems better to use "gdb-multiarch" from the linux system instead of the gdb that comes with the arduino tools.

props/blaster.h
profezzorn: sounds good to me

sound/hybrid_font.h
adays1234: Ugh. Ok so after hours of scouring the PJRC forums... looks like Teensy does not have a good way to do this in the 3.2 hardware. 3.5+ it added debug pins.

So... how do you suggest digging into this? Printf statements everywhere and serial monitor I guess?

profezzorn: printf is probably the only thing you can do right away.
It's a reasonably good way to debug anyways, debuggers are really much better for inspecting the current state than for figuring out the flow of a program.

profezzorn: So I tried this with and without the if statement and the hum worked equally well either way as far as I can tell.

adays1234: Alright. Well for now I removed this. Maybe the other changes have fixed this. I will retest everything and see if the hum still starts correctly but for the sake of finishing up this PR I removed the nullptr check since it was the last blocker.

PR #32, opened 2019-12-24
Updated saber_shtok_buttons.h prop file

ShtokyD:
Updated saber_shtok_buttons.h prop file with Color Change mode support. Added new modified saber_shtok_buttonsG.h prop file with stab gesture blade ignition and twist gesture retraction for both 1- and 2-button options.
[ProffieOS-master.zip](https://github.com/profezzorn/ProffieOS/files/3998238/ProffieOS-master.zip)


profezzorn:
This pull request is targeting the 3.x branch, which is not right. It should be targeting the master branch.

PR #33, opened 2019-12-26
Update saber_shtok_buttons.h file

ShtokyD:
Updated with Color Changing support for both button options, added Preset select backwards for 1-button option.

### DIFF
--- a/props/saber_shtok_buttons.h
+++ b/props/saber_shtok_buttons.h
@@ -1,3 +1,31 @@
+// 1 Button:
+// Activate Muted - fast double click while OFF
+// Activate blade - short click while OFF
+// Play/Stop Music - hold 1 second and release while ON
+// Turn the blade off - hold and wait till blade is off while ON (like in Plecter boards)
+// Next Preset - hold 1 second and release while OFF
+// Previous Preset - hold and wait while OFF
+// Lockup - hold + hit clash while ON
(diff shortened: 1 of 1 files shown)


props/saber_shtok_buttons.h
```diff
@@ -61,25 +96,34 @@ class SaberShtokButtons : public PropBase {
...
+#endif
       Off();
     }
         return true;
 
-// Force
+// Force and Color Change mode
       case EVENTID(BUTTON_NONE, EVENT_TWIST, MODE_ON | BUTTON_POWER):
+#ifndef DISABLE_COLOR_CHANGE
+        if (accel_.x < -0.15) {
```
profezzorn: It's better to use fusor.down().x < -0.15
The down() vector handles noise and rotation better.

props/saber_shtok_buttons.h
```diff
@@ -1,3 +1,31 @@
+// 1 Button:
```
profezzorn: These comments are a great addition :)

props/saber_shtok_buttons.h
ShtokyD: I just used these lines from the original code, I don't know coding :)

profezzorn: Guess I'll update it later then.

PR #35, opened 2020-01-25
Change monophonic hum check for blasters to use stun check. Helps resolve hybrid blaster/saber 

adays1234:
Minimal innocuous chance for blaster support to use the stun file instead of blast. Does not fix font detection issues overall but stop-gap for hybrid sabers/blasters.


sound/hybrid_font.h
```diff
@@ -72,7 +72,7 @@ class HybridFont : public SaberBase {
     font_config.ReadInCurrentDir("config.ini");
     STDOUT.print("Activating ");
     // TODO: Find more reliable way to figure out if it's a monophonic or polyphonic font!!!!
-    monophonic_hum_ = SFX_poweron || SFX_poweroff || SFX_pwroff || SFX_blast;
+    monophonic_hum_ = SFX_poweron || SFX_poweroff || SFX_pwroff || SFX_stun;
```
profezzorn: how about something like this:

 monophonic_hum_ = (SFX_in || SFX_out) ? false : (SFX_poweron || SFX_poweroff || SFX_pwroff || SFX_blast);

PR #40, opened 2020-02-08
Play sound when battery reaches min voltage

bellini13:
I've expanded your work on the low battery detection and warning, so that besides shutting the saber off when it reaches min voltage...  I also play a wav file named lowbatt.wav.  the wav file is played every 10-15 seconds.   It looks first at the font/preset location, and if it does not find it, it then looks at the root of the sdcard.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -741,21 +741,22 @@ public:
 
   virtual void CheckLowBattery() {
     if (battery_monitor.low()) {
-      // TODO: FIXME
       if (current_style() && !current_style()->Charging()) {
-	LowBatteryOff();
-	
-	if (millis() - last_beep_ > 5000) {
+	if (SaberBase::IsOn()) {
+		LowBatteryOff();
+	} else if (millis() - last_beep_ > 15000) {  // (was 5000)
+		STDOUT.print(battery_monitor.battery());
+		STDOUT.println("v");
+		STDOUT.println(" ");
 #ifdef ENABLE_AUDIO
-	  // TODO: allow this to be replaced with WAV file
-	  talkie.Say(talkie_low_battery_15, 15);
+		SaberBase::DoLowBatt();
 #endif
-	  STDOUT << "Battery low :" << battery_monitor.battery() << "\n";
-	  last_beep_ = millis();
+		last_beep_ = millis();
 	}
       }
     }
   }
+
   
   uint32_t last_beep_;
   float current_tick_angle_ = 0.0;
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -745,9 +745,7 @@ public:
 	if (SaberBase::IsOn()) {
 	  LowBatteryOff();
 	} else if (millis() - last_beep_ > 15000) {  // (was 5000)
-	  STDOUT.print(battery_monitor.battery());
-	  STDOUT.println("v");
-	  STDOUT.println(" ");
+	  STDOUT << "Low battery: " << battery_monitor_.battery() << " volts\n";
 	  SaberBase::DoLowBatt();
 	  last_beep_ = millis();
 	}
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -748,9 +748,7 @@ public:
 	  STDOUT.print(battery_monitor.battery());
 	  STDOUT.println("v");
 	  STDOUT.println(" ");
-#ifdef ENABLE_AUDIO
 	  SaberBase::DoLowBatt();
-#endif
 	  last_beep_ = millis();
 	}
       }
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -742,17 +742,16 @@ public:
   virtual void CheckLowBattery() {
     if (battery_monitor.low()) {
       if (current_style() && !current_style()->Charging()) {
-	if (SaberBase::IsOn()) {
-	  LowBatteryOff();
-	} else if (millis() - last_beep_ > 15000) {  // (was 5000)
-	  STDOUT << "Low battery: " << battery_monitor_.battery() << " volts\n";
-	  SaberBase::DoLowBatt();
-	  last_beep_ = millis();
+        LowBatteryOff();
+        if (millis() - last_beep_ > 15000) {  // (was 5000)
+          STDOUT << "Low battery: " << battery_monitor.battery() << " volts\n";
+          SaberBase::DoLowBatt();
+          last_beep_ = millis();
 	}
       }
     }
   }
-  
+
   uint32_t last_beep_;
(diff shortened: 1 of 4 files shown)


props/prop_base.h
```diff
@@ -741,21 +741,22 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
-	LowBatteryOff();
-	
-	if (millis() - last_beep_ > 5000) {
+	if (SaberBase::IsOn()) {
+		LowBatteryOff();
+	} else if (millis() - last_beep_ > 15000) {  // (was 5000)
+		STDOUT.print(battery_monitor.battery());
+		STDOUT.println("v");
+		STDOUT.println(" ");
 #ifdef ENABLE_AUDIO
```
profezzorn: We can remove the #ifdef since this can be done even if sound is not supported.

props/prop_base.h
```diff
@@ -741,21 +741,22 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
     if (battery_monitor.low()) {
-      // TODO: FIXME
       if (current_style() && !current_style()->Charging()) {
-	LowBatteryOff();
-	
-	if (millis() - last_beep_ > 5000) {
+	if (SaberBase::IsOn()) {
+		LowBatteryOff();
+	} else if (millis() - last_beep_ > 15000) {  // (was 5000)
+		STDOUT.print(battery_monitor.battery());
```
profezzorn: Maybe change this (and the following two lines) to

STDOUT << "Low battery: " << battery_monitor_.battery() << " volts\n";

props/prop_base.h
```diff
@@ -741,21 +741,22 @@ class PropBase : CommandParser, Looper, protected SaberBase {
 
   virtual void CheckLowBattery() {
     if (battery_monitor.low()) {
-      // TODO: FIXME
       if (current_style() && !current_style()->Charging()) {
-	LowBatteryOff();
-	
-	if (millis() - last_beep_ > 5000) {
+	if (SaberBase::IsOn()) {
+		LowBatteryOff();
```
profezzorn: All indentation needs to be two spaces.

sound/hybrid_font.h
```diff
@@ -548,6 +548,28 @@ class HybridFont : public SaberBase {
     return current_effect_length_;
   }
 
+	void SB_LowBatt() override
+	{
+		STDOUT.println(" ");
```
profezzorn: Too much debug print outs.
I don't think we need any of these checked in.

sound/hybrid_font.h
```diff
@@ -548,6 +548,28 @@ class HybridFont : public SaberBase {
...
+		STDOUT.println("|| SB_LowBatt function start");
+		STDOUT.println(" ");
+
+		//play the fonts low battery sound if it exists.  Otherwise try to play a wave file on the root of the SD Card
+		if (SFX_lowbatt.files_found()) {
```
profezzorn: No need to call files_found, just "if (SFX_lowbatt) {" will do.

sound/hybrid_font.h
```diff
@@ -548,6 +548,28 @@ class HybridFont : public SaberBase {
...
+			PlayCommon(&SFX_lowbatt);
+		} else {
+			MountSDCard();
```
profezzorn: Let's remove all of this.
If we want to support a global location for files, we should just add it to the effect class so that it can be used for any and all files.

props/prop_base.h
```diff
@@ -741,16 +741,12 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
       if (current_style() && !current_style()->Charging()) {
-	LowBatteryOff();
-	
-	if (millis() - last_beep_ > 5000) {
-#ifdef ENABLE_AUDIO
-	  // TODO: allow this to be replaced with WAV file
-	  talkie.Say(talkie_low_battery_15, 15);
-#endif
-	  STDOUT << "Battery low :" << battery_monitor.battery() << "\n";
+	if (SaberBase::IsOn()) {
```
profezzorn: LowBatteryOff() already checks SaberBase::IsOn(), why do it again?

sound/hybrid_font.h
```diff
@@ -548,7 +548,8 @@ class HybridFont : public SaberBase {
     return current_effect_length_;
   }
 
-
+  void SB_LowBatt() override { PlayCommon(&SFX_lowbatt); }
```
profezzorn: We still need the talkie fallback if SFX_lowbatt is not available, guess I should have been more clear.

props/prop_base.h
bellini13: Because we do something different if the saber is off.  So i still needed an on/off check.  If it is on, then I allow your function to run, which deals with turning the saber off.   If the saber is off, then I execute my logic to play the sound every 15 seconds

sound/hybrid_font.h
bellini13: ok i have added that

sound/hybrid_font.h
```diff
@@ -548,7 +548,20 @@ class HybridFont : public SaberBase {
...
 
-
+  void SB_LowBatt() override
+  {
+    //play the fonts low battery sound if it exists
+    //TODO: if lowbatt does not exist in the font, then play a global sound (instead of a beep or a talkie)
+    if (SFX_lowbatt) {
+      PlayCommon(&SFX_lowbatt);
+    } else {
+      STDOUT << "Battery low :" << battery_monitor.battery() << "\n";
```
profezzorn: This printout is not needed since already have something very similar in CheckLowBattery()

sound/hybrid_font.h
```diff
@@ -548,7 +548,19 @@ class HybridFont : public SaberBase {
     return current_effect_length_;
   }
 
-
+  void SB_LowBatt() override
+  {
+    //play the fonts low battery sound if it exists
+    //TODO: if lowbatt does not exist in the font, then play a global sound (instead of a beep or a talkie)
```
profezzorn: This todo can be removed, "global" sounds are handled by the font search path now.

sound/hybrid_font.h
```diff
@@ -548,7 +548,19 @@ class HybridFont : public SaberBase {
     return current_effect_length_;
   }
 
-
+  void SB_LowBatt() override
+  {
```
profezzorn: This { should be at the end of the previous line.

sound/hybrid_font.h
```diff
@@ -548,7 +548,19 @@ class HybridFont : public SaberBase {
...
+    //play the fonts low battery sound if it exists
```
profezzorn: add a space after //

PR #42, opened 2020-03-12
Responsive Styles

fdarosa2663:
added responsive_styles.h and made change to LB button to make control less cumbersome for users

### DIFF
--- a/ProffieOS.ino
+++ b/ProffieOS.ino
@@ -400,7 +400,6 @@ struct is_same_type<T, T> { static const bool value = true; };
 #include "styles/transition_effect.h"
 #include "styles/transition_loop.h"
 
-
 // functions
 #include "functions/ifon.h"
 #include "functions/change_slowly.h"
--- a/ProffieOS.ino
+++ b/ProffieOS.ino
@@ -400,6 +400,7 @@ struct is_same_type<T, T> { static const bool value = true; };
 #include "styles/transition_effect.h"
 #include "styles/transition_loop.h"
 
+
 // functions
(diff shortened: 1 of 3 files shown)


ProffieOS.ino
```diff
@@ -1,1666 +1,1670 @@
...
-
-#include "common/clock_control.h"
-
-void loop() {
-#ifdef MTP_RX_ENDPOINT
-  mtpd.loop();
-#endif
-  Looper::DoLoop();
-}
+/*
```
profezzorn: This diff shows every line in ProffieOS.ino has been changed somehow.
I don't know why that is, but I suspect that maybe line endings have been changed from NL to CRLF-NL
Whatever the reason, it makes it very hard to see what the actual changes are.

fdarosa2663:
The only change to the .ino was to include the responsive_styles.h below transitions. I made the changes in Arduino 1.8.9 would that cause?  

----added----
// responsive styles
#include "styles/responsive_styles.h"
------------------------

should I replace with original .ino and make same change again?

ProffieOS.ino
fdarosa2663: The only change to the .ino was to include the responsive_styles.h below transitions. I made the changes in Arduino 1.8.9 would that cause?

----added----
// responsive styles
#include "styles/responsive_styles.h"
should I replace with original .ino and make same change again?

profezzorn: You need to fix the difff though, or I can't merge the change.

fdarosa2663:
will redo to fix differences

PR #51, opened 2020-04-03
update sa22c props file to include lb and melt

SA-22C:


props/saber_sa22c_buttons.h
```diff
@@ -27,6 +29,8 @@
 // Volume UP - short click while OFF and in VOLUME MENU
 // Prev Preset - hold and wait while OFF
 // Color Change mode - hold + toggle AUX while ON
+// Lightning Block - hold + CLASH when ON
```
profezzorn, line 32: Seems kind of weird to use clash for lightning block, but I don't really mind if you don't... :)

SA-22C: It is a bit weird, yeah. Not sure how else to do it unless I figure out the suggested double click and hold event suggested on Facebook

PR #53, opened 2020-04-06
EVENT_DOUBLE_CLICK_HELD

SA-22C:
My minor updates to enable EVENT_DOUBLE_CLICK_HELD


buttons/button_base.h
```diff
@@ -24,20 +24,29 @@ class ButtonBase : public Looper,
       while (!DebouncedRead()) YIELD();
       prop.Event(button_, EVENT_PRESSED);
       if (millis() - push_millis_ < 500) {
-        prop.Event(button_, EVENT_DOUBLE_CLICK);
+        doubleclicked_ = true;
```
SA-22C, line 27: Instead of triggering EVENT_DOUBLE_CLICK now, we flag that a second button press has occurred so that when the button is released, we can evaluate if it is a double-click or a double-click-and-held event.

buttons/button_base.h
```diff
@@ -24,20 +24,29 @@ class ButtonBase : public Looper,
...
       }
       while (DebouncedRead()) {
         if (millis() - push_millis_ > 300) {
-          prop.Event(button_, EVENT_HELD);
+          if (!doubleclicked_) {
+            prop.Event(button_, EVENT_HELD);
+          } else {
+            prop.Event(button_, EVENT_DOUBLE_CLICK_HELD);
+            doubleclickedandheld_ = true;
+          }
```
SA-22C, line 39: Here we handle the three HELD events to ensure that if a double-click occurs, we don't triger the single-click HELD events.  The second bool doubleclickedandheld_ is used to check if, when the button is released, if the DOUBLE_CLICK event should fire.

buttons/button_base.h
```diff
@@ -50,6 +59,13 @@ class ButtonBase : public Looper,
       }
       while (DebouncedRead()) YIELD();
       prop.Event(button_, EVENT_RELEASED);
+      if (doubleclicked_) {
+        if (!doubleclickedandheld_) {
+          prop.Event(button_, EVENT_DOUBLE_CLICK);
+        }
+        doubleclicked_ = false;
+      }
+      doubleclickedandheld_ = false;
```
SA-22C, line 68: If the button is released and a HELD event wasn't triggered, the DOUBLE_CLICK event fires.  The two bools used to control the doubleclick behavior are reset.

buttons/button_base.h
```diff
@@ -81,6 +97,8 @@ class ButtonBase : public Looper,
     STDOUT.println(" button");
   }
 
+  bool doubleclicked_ = false;
+  bool doubleclickedandheld_ = false;
   const char* name_;
```
SA-22C, line 102: doubleclicked_ is used to flag that a double click has started
doubleclickedandheld_ is used to flag that the second click was held down long enough to be a HELD event, to ensure that DOUBLE_CLICK doesn't fire after a DOUBLE_CLICK_AND_HELD

buttons/button_base.h
profezzorn: This changes how existing props work.
I think we still need to emit some event here. It can either be EVENT_DOUBLE_CLICK, or it could be something else.
If we use something other than EVENT_DOUBLE_CLICK, we need to do a search-and-replace in existing props to use the new event.

buttons/button_base.h
```diff
@@ -24,20 +24,29 @@ class ButtonBase : public Looper,
...
         push_millis_ = millis();
         current_modifiers |= button_;
       }
       while (DebouncedRead()) {
         if (millis() - push_millis_ > 300) {
-          prop.Event(button_, EVENT_HELD);
+          if (!doubleclicked_) {
+            prop.Event(button_, EVENT_HELD);
+          } else {
+            prop.Event(button_, EVENT_DOUBLE_CLICK_HELD);
```
profezzorn, line 37: if we're going to have EVENT_DOUBLE_CLICK_HELD, why not add DOUBLE_CLICK_EVENT_HELD_MEDIUM, DOUBLE_CLICK_EVENT_HELD_LONG while we are at it?

buttons/button_base.h
SA-22C: It does change how they work in theory, but in practise it seems pretty unnoticeable that double-click events happens after the second button release rather than the second button press.  If we don't supress the double-click event here, the double-click-and-hold doesn't work, as a spurious double-click will always precede it.

profezzorn: We can move EVENT_DOUBLE_CLICK to later, but we should still emit *some* event at this point.
Existing props that don't care about double-click-and-hold events can use the event that we emit here instead of the more complicated ones that we emit later.  Maybe the event we emit here should be called EVENT_DOUBLE_PRESSED?

SA-22C: done

buttons/button_base.h
SA-22C: done

buttons/button_base.h
profezzorn: Please also update the existing props to use EVENT_DOUBLE_PRESSED instead of EVENT_DOUBLE_CLICKED.

SA-22C: done

profezzorn:
I just wanted to explain the delay here....
I've been thinking about this quite a bit in the past couple of days, trying to figure out how I want this to work. I think it's important that props have the ability to react to the buttons in any way they choose. That doesn't automatically mean that having all of these events in the buttons themselves is a bad idea though....  Anyways, I just wanted to drop a note to say that I haven't forgotten about this.

SA-22C:
No worries, thanks for the update. 

-Matthew McGeary

profezzorn:
I've made some updates to the button code, hopefully it will cover the cases you need, but it probably means that this PR should be closed. Let me know if it works for you or not.

SA-22C:
Closing, the new button changes will make this pull obsolete.  Thanks.

profezzorn:
So have you tried using the new button events?
I was hoping to get an updated prop file before 4.x goes live. :)

SA-22C:
I haven't yet, doing some online courses that are taking up all my spare
time lately.

PR #58, opened 2020-05-06
Fix "blast" effect on ColorChange button press

fdarosa2663:
I found a missing "return true;" in the prop file that was triggering a blast effect on ToggleColorChangeMode()


profezzorn:
nice catch

fdarosa2663:
thanks, yeah at first I kept thinking I was double clicking or something and then decided to look at the code because it kept occurring

PR #60, opened 2020-05-09
updated props file

SA-22C:
Updated my props file to use the new button events and enable all new effects for 1,2 and 3 button sabers.


props/saber_sa22c_buttons.h
```diff
@@ -103,236 +165,250 @@ class SaberSA22CButtons : public PropBase {
...
+    return true;
+  // 2 and 3 button color change
 #ifndef DISABLE_COLOR_CHANGE
-          if (SaberBase::GetColorChangeMode() != SaberBase::COLOR_CHANGE_MODE_NONE) {
-            // Just exit color change mode.
-            // Don't turn saber off.
-            ToggleColorChangeMode();
-            return true;
-          }
+  case EVENTID(BUTTON_AUX, EVENT_FIRST_CLICK_SHORT, MODE_ON | BUTTON_POWER):
```
SA-22C, line 267: When performing hold power and toggle aux for color change mode, I'm getting a blast effect when the aux button is released.  Any suggestions on how I might stop that from happening, without compromising my ability to have a double-click happen WITHOUT the first click also producing a blast?

profezzorn: I updated button_base to not emit saved events for events that have already triggered an action.
I think that should fix the problem, please let me know if it works or not.

SA-22C: That did it, thanks!

PR #70, opened 2020-08-13
Pullup and pulldonw option on buttons

artandtechvof:
For some of my builds I find it more convenient to have pulldown resistors on my buttons vs the default pullup resistors of the ProffieOS.

Reading up on capabilities of the Teensy and Proffieboard STM, this is perfectly do-able and would only require an alteration to the button.h file. Adding an extra argument with default declaration would make it backward compatible with all current code.

To use it the button definition in the config file would look something like this;

#ifdef CONFIG_BUTTONS
Button PowerButton(BUTTON_POWER, powerButtonPin, "pow", false);  // this has pulldown resistor
Button AuxButton(BUTTON_AUX, auxPin, "aux", true); // this has pull up resistor
Button Aux2Button(BUTTON_AUX2, aux2Pin, "aux2"); // this has pull up resistor by default when the argument is not given

#endif


profezzorn:
While this is a good idea, and a reasonable implementation. I think I would prefer if this was done as a separate class.
I don't trust that the compiler will optimize out the code if not used otherwise. It only uses a handful of bytes, but since there is an easy way to avoid that, I say why not?

artandtechvof:
Juts tested what the amount off additional program space is for single button and 4 button setup;

1 button setup;
- Proffieboard without Pulldown option:  Sketch uses 201136 bytes (81%) of program storage space. Maximum is 245760 bytes.
- Proffieboard with Pulldown option set false:  Sketch uses 201144 bytes (81%) of program storage space. Maximum is 245760 bytes.  -> 8 bytes extra, with option pull down...

4-button setup
- Proffieboard without Pulldown option:  Sketch uses 201328 bytes (81%) of program storage space. Maximum is 245760 bytes.
- Proffieboard with Pulldown mod:        Sketch uses 201384 bytes (81%) of program storage space. Maximum is 245760 bytes.  -> 56 additional bytes when option are left default
- Proffieboard with Pulldown mod:        Sketch uses 201392 bytes (81%) of program storage space. Maximum is 245760 bytes.  -> 64 additional bytes when option are set true/false/true/false

(for Teensy compilation this is exaclty the same)

In my opinion this is reasonable, but might only be useful for a very few builders, In stead of having a separate class for this pull-down option, could an #ifdef also be used? So, if you put in your config file #define ENABLE_PULLDOWN, and then in the button.h use the #ifdef ENABLE_PULLDOWN to make the pulldown option available. Doing so would keep the code exactly the same when the pulldown option is not needed??

An additional class is still an option if you prefer to enable it that way.

profezzorn:
Using a define would be a last resort in my opinion as they make for ugly code and bad type-safety.
(If you spell the define wrong, nothing happens!)
A separate class is still the right way to do this IMHO.

PR #84, opened 2020-09-01
Update ProffieOS.ino

bellini13:
Add support for a configurable startup delay

### DIFF
--- a/ProffieOS.ino
+++ b/ProffieOS.ino
@@ -1696,11 +1696,15 @@ void setup() {
   // Wait for all voltages to settle.
   // Accumulate some entrypy while we wait.
   uint32_t now = millis();
+
+int PROFFIEOS_STARTUP_DELAY = 1000;
 #ifdef DOSFS_CONFIG_STARTUP_DELAY
-#define PROFFIEOS_STARTUP_DELAY DOSFS_CONFIG_STARTUP_DELAY
-#else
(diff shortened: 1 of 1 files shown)


ProffieOS.ino
```diff
@@ -1696,11 +1696,15 @@ void setup() {
   // Wait for all voltages to settle.
   // Accumulate some entrypy while we wait.
   uint32_t now = millis();
+
+int PROFFIEOS_STARTUP_DELAY = 1000;
```
profezzorn, line 1700: I don't trust the compiler to optimize this, it would be better if we just do it with #if and #ifdefs I think.

bellini13: As far as I can tell, there is no else statements with #ifdef, which makes having to deal with this a bit sloppy looking.  

I had also asked this question on TRA forum, but did not receive feedback.   Could you address this here?

I'm not understanding why DOSFS_CONFIG_STARTUP_DELAY is needed?  Is this for backwards compatibility?  Can we ignore that constant and just assume the default startup value of 100?  Or do older OS versions or Teensy need 1000 ?

bellini13: also, the code I submitted is using #ifdef, so i'm a bit confused as to what the compiler concern is

profezzorn:
I have added CONFIG_STARTUP_DELAY in a different checkin.

PR #87, opened 2020-09-05
Synced in out audio, functions WavLen<>, WavPos<>, WavRem<>, Pct<>, Mult<>

artandtechvof:
This is the second attempt of merging the new functions that allow the wav-lentgh to be used a a paramater for bladestyle effects as discussed and demonstrated on the rebelarmory.

There is a warning that these will not merge automatically, so I want to start checking what is needed to get this into the main fork (if at all?)
Also start sparring with you to see if you want to alter the way this feature is implemented.

Currently there is no coupling between sound effect and blade effect directly. This is all distributed by saber_base.h, When an effect is activated, balde and- sound classes do there own thing. The blade style does not know directly which soundfile is selected and which buffered_wavplayer is playing this sound.
My solution was to introduce an extra item to the wav_player 'effectname' and '2nd effectname'. (because pstoff is added as second effect to the wavplayer, and therefore is not standalone like preon)
The WavLen functions have a look-up table to cross-correlate EFFECT_NAME and SFX# effect names. Using the soundeffect name and blade effect name, we can now cycle trough the wavplayers (there are only a couple of them) to find the one that is playing the effect we are after, and thus being able to track its length, current position and remaining time.

I have been looking into using the OneshotEffectDetector to figure out which effect is active, but was unable to retrieve the wav_player that was used for the effect.

The helper functions Pct<> and Mult<> were separated from the WavLen functions as you suggested, and can be re-used for any purpose within the OS architecture.

Generic wav timer alterations to the code;
playwav -> debug and effectname()
bufferedwav -> effectname()
effect.h name_ -> public
blade_base.h -> added EFFECT_PSTOFF
Optional ifon -> debug info
added;
functions/wav_time
functions/wav_effect
functions/mult
functions/pct
ProffieOS.ino -> include wav_time.time, pct.h, mult.h


functions/mult.h
```diff
@@ -0,0 +1,28 @@
...
+
+template<class F, class V>
+class Mult {
+public:
+  void run(BladeBase* blade) {
+    f_.run(blade);
+    v_.run(blade);
+  }
+  int getInteger(int led) {
+    return clampi32((f_.getInteger(led) * v_.getInteger(led)) >> 15, 0, 32768);
```
profezzorn, line 21: no clamp

functions/mult.h
```diff
@@ -0,0 +1,28 @@
+#ifndef FUNCTIONS_MULT_H
+#define FUNCTIONS_MULT_H
+
+// Usage: Pct<F, V>
```
profezzorn, line 4: Documentation is for the Pct<> function.
Abbreviations should be avoided unless they are completely obvious, so I think Percentage should be the right thing here.

functions/pct.h
```diff
@@ -0,0 +1,28 @@
...
+// Gets Percentage V of value F, 
+// Percentages over 100% are allowed and will effectively be a multiplier. 
+// Can be used to multiply or devide any number  
+// F, V: INTEGER
+// return value: INTEGER
+
+class BladeBase;
+
+template<class F, class V>
+class Pct {
```
profezzorn, line 14: Pct isn't really needed, however, we could easily make an alias for it that just uses Mult<>

sound/effect.h
```diff
@@ -297,7 +297,10 @@ class Effect {
     }
 
     default_output->print("Playing ");
-    default_output->println(filename);
+    default_output->print(filename);
```
profezzorn, line 300: why are we printing the same filename twice?

functions/wav_effect.h
```diff
@@ -0,0 +1,173 @@
...
+       case EFFECT_NONE:         return "none";
+      } 
+     return "none"; 
+  }
+};
+
+const char* pstoff_ = "pstoff";
+
+#endif
+/*  blade_base.h
```
profezzorn, line 50: I don't understand what this comment is doing here?

functions/wav_effect.h
```diff
@@ -0,0 +1,173 @@
+#ifndef FUNCTIONS_WAV_EFFECT_H
+#define FUNCTIONS_WAV_EFFECT_H
+
+#include "../blades/blade_base.h"
+
+
+class WavEffect{
+public:	
+
+  const char* GetEffectName(BladeEffectType type) {
```
profezzorn, line 10: Why do we need to know the names of effects enums?

functions/wav_time.h
```diff
@@ -0,0 +1,143 @@
...
+        //value_ = 32767;
+        shown_ = false;
+    }
+  }
+  int getInteger(int led) { return value_; }
+  int value_ = 32767;
+private:
+  int idx;
+  const char* eff_name_;
+  bool found_;
```
profezzorn, line 45: found_ should not be an object variable, it should be a local variable

functions/wav_time.h
```diff
@@ -0,0 +1,143 @@
...
+// EFFECT_NAMENUM : any off the available effect form the style edittor
+// Returns length of wav file returned as const int in [ms]
+// returned value: INTEGER
+//
+template<BladeEffectType EFFECT>
+class WavLen {
+ public:
+  void run(BladeBase* blade) {
+    found_ = false;
+    eff_name_ = wav_effect.GetEffectName(EFFECT);//find correspondig SoundEffectName for BladeEffectType
```
profezzorn, line 18: why are we doing this by name? Why not by enum?

functions/wav_time.h
```diff
@@ -0,0 +1,143 @@
...
+    for (size_t i = 0; i < NELEM(wav_players); i++) {
```
profezzorn, line 19: I'm not a big fan of iterating over the wav players to find stuff.
What if there are two playing the same effect?

functions/wav_time.h
```diff
@@ -0,0 +1,143 @@
...
+        if(strstr(eff_name_, wav_players[i].effectname()) && wav_players[i].isPlaying()) {found_ = true; idx=(int)i;}
```
profezzorn, line 20: Interacting with the wav players can be dangerous, because they are updated by interrupts and so these functions can change their values in the middle of executing this code, which makes it really hard to know if this will work as intended or not. Especially if the compiler decides to reorder things.

functions/wav_time.h
```diff
@@ -0,0 +1,143 @@
...
+};
+
+
+// Usage: WavPos<EFFECT_NAME>  
+// EFFECT_NAME : any off the available effect form the style edittor
+// Returns current position [0..32768] of wav file returned as const int in [ms]
+// returned value: INTEGER
+//
+template<BladeEffectType EFFECT>
+class WavPos {
```
profezzorn, line 58: What is this actually for?

functions/wav_time.h
```diff
@@ -0,0 +1,143 @@
...
+template<BladeEffectType EFFECT>
+class WavLen {
+ public:
+  void run(BladeBase* blade) {
+    found_ = false;
+    eff_name_ = wav_effect.GetEffectName(EFFECT);//find correspondig SoundEffectName for BladeEffectType
+    for (size_t i = 0; i < NELEM(wav_players); i++) {
+        if(strstr(eff_name_, wav_players[i].effectname()) && wav_players[i].isPlaying()) {found_ = true; idx=(int)i;}
+        if(strcmp(eff_name_,"pstoff")==0) {if(strstr(eff_name_, wav_players[i].effect2ndname()) && wav_players[i].isPlaying()) {found_ = true; idx=(int)i;}}
+            if (found_) { value_ = clampi32(wav_players[idx].length()*1000, 0, 32768);}
```
profezzorn, line 22: no clamp

functions/wav_time.h
```diff
@@ -0,0 +1,143 @@
...
+            if (found_ && !shown_) {
```
profezzorn, line 23: Do we really need to print things out other than for debugging?
It uses up FLASH memory for something that most users never see.
Also, if we're going to print something outside of commands, we should use default_output instead of STDOUT.

functions/wav_time.h
```diff
@@ -0,0 +1,143 @@
...
+};
+
+
+// Usage: WavRem<EFFECT_NAME>  
+// EFFECT_NAME : any off the available effect form the style edittor
+// Returns remaining time [32768..0] of wav file returned as const int in [ms]
+// returned value: INTEGER
+//
+template<BladeEffectType EFFECT>
+class WavRem {
```
profezzorn, line 104: Why?

sound/effect.h
```diff
@@ -394,6 +397,11 @@ class Effect {
   }
 
   Effect* next_;
+  Effect* current_;
+  
+  // All files must start with this prefix. Public so other functions can use it
+  const char* name_ = "";
```
profezzorn, line 403: We don't need this name.
If we have a pointer to the Effect, we can just use that to identify what is being played.

sound/playwav.h
```diff
@@ -291,7 +300,11 @@ class PlayWav : StateMachine, public AudioStream {
           len_ -= bytes_to_skip;
           start_ = 0.0;
         }
-
+        
+        default_output->print(" length ");
+        default_output->print(length());
+        default_output->println(" [s]");
+        duration_ = length();
```
profezzorn, line 307: why?
We already have a length() function.

sound/playwav.h
```diff
@@ -371,12 +389,22 @@ class PlayWav : StateMachine, public AudioStream {
...
+     return effect2ndname_;
+  }  
+  
 private:
   volatile bool run_ = false;
   Effect* volatile effect_ = nullptr;
   Effect::FileID new_file_id_;
   Effect::FileID old_file_id_;
   char filename_[128];
+  const char* effectname_;
```
profezzorn, line 406: isn't this the same as current_->name_ ?

sound/playwav.h
```diff
@@ -371,12 +389,22 @@ class PlayWav : StateMachine, public AudioStream {
...
+  const char* effect2ndname_;
```
profezzorn, line 407: isn't this the same as next_->name_ ?

sound/playwav.h
```diff
@@ -392,6 +420,7 @@ class PlayWav : StateMachine, public AudioStream {
 
   size_t len_ = 0;
   volatile size_t sample_bytes_ = 0;
+  float duration_ = 0.0;
```
profezzorn, line 423: isn't this the same as length()

functions/wav_time.h
```diff
@@ -0,0 +1,143 @@
...
+            }
+    }
+    if (!found_){
+        if(shown_) {STDOUT << "WavLen stopped for effect '" << eff_name_ <<"', value retained is: " << value_ << "\n";}
+        //should we retain value after the effect has gone? Only reset it when new effect is selected?
+        //value_ = 32767;
+        shown_ = false;
+    }
+  }
+  int getInteger(int led) { return value_; }
```
profezzorn, line 40: In some cases (transitions for example) ProffieOS tends to snapshot the length of the transition when the transitions start. I don't see how this code can guarantee that the length of the associated wav file is actually known at that time.

functions/mult.h
artandtechvof: clamp removed

functions/mult.h
artandtechvof: resolved

functions/pct.h
artandtechvof: Indeed, I have made the necessary adjustments

sound/effect.h
artandtechvof: First print is the wav file including path, second is the effect name, as a debug output to check that the correct effect was selected

functions/wav_effect.h
artandtechvof: These comments were a reminder for me to make sure the cross correlation of these names was correct.

functions/wav_effect.h
artandtechvof: We need to know the name to be able to look for the name in the attributies of the wav_players
Monophonic/polyphonic can have multiple names EFFECT_IGNITION -> poweron,out,poweronf

Str comparison does not feel right, but it was the most efficient way to solve the problem of not knowing which soundfile was chosen for a certain effect, and thus being able to retrieve the length and position of the sound playing.

functions/wav_time.h
artandtechvof: resolved

sound/playwav.h
artandtechvof: Was a remainder of some earlier experiments, removed now. I did not actually use it for my functions.

functions/wav_time.h
artandtechvof: We need to have the str comparison of the wav_player attribute

functions/wav_time.h
artandtechvof: Is that even possible (clash maybe??). What would the current effect be. Surely one of those is an older effect still playing?

functions/wav_time.h
artandtechvof: Every loop, the wavplayers are checked to find were the current sound is playing. even if the order or wavplayer number would change while playing the sound, we still would find the correct one right?

functions/wav_time.h
artandtechvof: If you look at my demo video, you see I can also control the extension and retraction for the blade directly using the WavPos<> function. It becomes interesting when you use a Mult<WavPos, WavPos>. Then you get non linear extension and retraction, very cool effect.

functions/wav_time.h
artandtechvof: Agreed for dveveloping and debugging this is very usefull, but does not need to be active once the final version is ready (or maybe via #define ENABLE_DEVELOPER_COMMANDS?)

STDOUT is used a lot throughout ProffieOS, so I figured it wouldn't hurt to use is. But if preference is to use default_output instead, that is an easy fix.

functions/wav_time.h
artandtechvof: Here's an example of non lienar control of extension and retraction of the blade, controlled by the position of the wav-file;

    StylePtr<
      Layers<
            DeepPink,
            BlastL<White>,
            LockupL<AudioFlicker<Blue,White>>,
            TransitionEffectL<TrConcat<TrInstant,AlphaL<White,Bump<Int<16384>>>,TrFadeX<WavLen<EFFECT_CLASH>>>,EFFECT_CLASH>,
            TransitionEffectL<TrConcat<TrInstant,AlphaL<White,Bump<Int<30768>>>,TrFadeX<WavLen<EFFECT_CLASH>>>,EFFECT_STAB>,
          //Make sure Transition Effect has enough time to do its thing, allocate time for extension effect, and immediate go into retraction effect
            InOutTrL<TrDelayX<WavLen<EFFECT_IGNITION>>,TrInstant>, 
          //position controller extension
            TransitionEffectL<TrConcat<TrInstant,AlphaL<DeepPink,SmoothStep< Scale<Mult<Mult<WavRem<EFFECT_IGNITION>,WavRem<EFFECT_IGNITION>>,WavRem<EFFECT_IGNITION>>,Int<32768>,Int<0> >,Int<-1>> >,TrDelayX<WavLen<EFFECT_IGNITION>>>,EFFECT_IGNITION>, 
          //position controller retraction
            TransitionEffectL<TrConcat<TrInstant,AlphaL<DeepPink,SmoothStep< Scale<Mult<Mult<WavPos<EFFECT_RETRACTION>,WavPos<EFFECT_RETRACTION>>,WavPos<EFFECT_RETRACTION>>,Int<32768>,Int<0> >,Int<-1>> >,TrDelayX<WavLen<EFFECT_RETRACTION>>>,EFFECT_RETRACTION> 
            >
    >()

functions/wav_time.h
artandtechvof: removed for WavLen,

WavPos and WavRem need them to prevent glitches at end of file, when buffer is not yet empty, but file is read.

functions/wav_effect.h
artandtechvof: Removed comments

functions/wav_time.h
artandtechvof: Seems to be working fine, even for styles that have multiple transitions in them.

sound/playwav.h
artandtechvof: removed now

sound/playwav.h
artandtechvof: Yes, that should be the case

sound/playwav.h
artandtechvof: Yes that should be the case

sound/effect.h
artandtechvof: This is were I struggle I have to confess. I can see the Effect* for current_ and next_ int the effecht.h. 
Currently I 'nick' the effectname and effect2ndname from the PlayOnce() and initialize them there.
If you could give me a hint how to get an Effect* directly to the currently playing wav for the effect, that would help a lot.

profezzorn:
I don't think this is the right way to do this.
I like some of the ideas here, but I think this CL just has too many problems.

#1: using string comparisons is terrible
#2: Once the wav player moves on and plays a different effect, the value is lost, which is probably not what you want
#3: There is no guarantee that the value is actually known in time

Probably the solution to #1 is to do something I've been thinking about for a while, which is to promote the EFFECT_* enum into SaberBase and use it to actually trigger events. So instead of one function per effect, there would be one function which takes an EFFECT_????? enum value as an argument instead. That one function could then save the duration into a simple array of int32_t keyed on the EFFECT_* value, then the blade style could then go look it up there afterwards.

That only leaves the problem #3, which may in fact not be a problem at all, but we need to be sure.
If we really need the current progress, I would just add another entry to the array which saves the start time, but I'm still not entirely convinced that that is a good idea.

artandtechvof:
I agree its not the most elegant solution to this problem. This is more prototyping for proof of concept, and it seems to work, with the desired effect. From this exercise I also learned that having the Percentage<> and Mult<> functions are helper functions that should exist when using the WavLen function to get the maximum effect from these functions.
WavPos and WavRem will give you some additional info, but for now its only playing around with the non linear extension and retraction. I feel that could also be done with a diffrent type of InOutHelper or InOutFunc.

#1 Agreed that str comparsion is not want you want
#2 The WavLen will retain the length of the last played effect it was used for, so it will keep track of length, even after the effect has finished.
#3 had me worried too, somehow it seems to work, and to be sure, I first started of by giving the WavLen an initial time of 32676ms to make sure we were not setting thing to 0ms before they would even happen. It also seems that the sound is started in time for the blade effect to get the correct value, even if I set initial time to 0.

Having more coupling between bladestyle/effect and sound effect via SaberBase was what I was hoping to find when I started this little project. It will make these kind of function much easier to implement, and it somehow feels more logic that sound and style are aware of each other.

For now this solution will remain on our fork, for anybody to play have a look at.

How could we proceed in embedding this more into the core of ProffieOS? Could I start with experimenting with expanding the SaberBase to keep track of style ánd sound? I feel this is a mayor architectural change for ProffieOS, that you want to be in the lead of. Wherever I can assist or develop for you I would love to do so.

profezzorn:
One potential problem with the functions-to-enum conversion is that it makes it impossible for the compiler to optimize out things that aren't used. Now, because of the way it's written, it can't actually do that anyways, but one can dream right? Either way, regardless of who actually does it, I will be reviewing the changes and make sure they are working the way I want to, so it is possible for someone else to do it.

When it comes to incorporate things into ProffieOS, I do have fairly high standards, which can make it difficult sometimes. I sometimes just end up re-implementing things myself because it's easier. There are a few rules of thumb that I try to go by though...

A) small changes are much easier to get merged (For instance, you could start with Mult<> in it's own pull request, which would be super-easy to get in.)
B) while prototyping things on your own is fine, if your goal is to get it merged, having a conversation with me about the best way to implement something might be a requirement
C) optional features should ideally take zero memory and cpu when not used 

However, at the end of the day, there are always going to be some features that I feel doesn't fit in ProffieOS, that other people really want, and if those features cant be done that fit rule (C) above, then having them in a fork of ProffieOS is probably the best thing.

In my opinion, WavLen<> would be a great addition to ProffieOS, but I'm not sure about WavPos<> and WavRem<>. What I am sure about though is that the implementation will need work before it can be merged...

profezzorn:
Are you still working on this PR?
I'd love to have WavLen<> at some point, but it might be better to create a new PR for it?

artandtechvof:
> Are you still working on this PR?
> I'd love to have WavLen<> at some point, but it might be better to create a new PR for it?

Yes, that might be better. We have established, that it would require a fundamental change to SaberBase as mentioned in [https://github.com/profezzorn/ProffieOS/pull/87#issuecomment-687884751](url)

So maybe for now, close this pull request, and wait until we figure out how to make the changes to SaberBase?

artandtechvof:
Hi Fredrik,

Hope you are doing well,

I would like to pick this subject up again. Having done a lot of other stuff lately I would have to get familiar with the latest changes you have made to ProffieOS. Would you be willing to give me a head start please, and suggest what changes to the OS you have made, would be suitable to get the WavLen<> functions fully ProffieOS compatible?

( I noticed that some people are integrating the my current version of it manually into the latest 5.7, and I also received some question about it via DM on TRA )

Hoping hearing from you
Regards,
Rob.

PR #88, opened 2020-09-06
DimBlade change in real-time

artandtechvof:
modified:   blades/dim_blade.h
	modified:   common/saber_base.h
	modified:   config/default_proffieboard_config.h
	modified:   props/prop_base.h

Permanently embedded DimBlade option in ProffieOS. It will allow for real-time brightness alteration
using a Menu (comparable to colorchange) or serial commands (dim ##, get_dim)

Maybe it woul even be nice to have this with a #define DIMBLADE 100, that would automaticaly wraps the DimBlade(#,) around the BladeConfig? (This would then affect all blades defined, including accent leds and crystal chambers)?

### DIFF
--- a/blades/dim_blade.h
+++ b/blades/dim_blade.h
@@ -1,13 +1,24 @@
 #ifndef BLADES_DIM_BLADE_WRAPPER_H
 #define BLADES_DIM_BLADE_WRAPPER_H
 
-class DimBladeWrapper : public BladeWrapper, BladeStyle {
+class DimBladeWrapper : public BladeWrapper, BladeStyle 
+#ifdef ENABLE_MENU_DIMBLADE
+      ,protected SaberBase
+#endif
+      {
 public:
   DimBladeWrapper(BladeBase* blade, int fraction) {
     blade_ = blade;
+#ifdef ENABLE_MENU_DIMBLADE
+    SaberBase::SetBrightness((float)fraction/16384.0*100.0);
+#else
     fraction_ = fraction;
-  }
+#endif
+    }
   void set(int led, Color16 c) override {
+#ifdef ENABLE_MENU_DIMBLADE
+    int fraction_ = SaberBase::GetCurrentBrightness();
+#endif
     Color16 ret;
     ret.r = clampi32((c.r * fraction_) >> 14, 0, 65535);
     ret.g = clampi32((c.g * fraction_) >> 14, 0, 65535);
@@ -62,7 +73,9 @@ public:
   
 private:
   BladeStyle *current_style_ = nullptr;
+#ifndef ENABLE_MENU_DIMBLADE
   int fraction_;
+#endif
 };
 
 // Reduces the brightness of the blade.
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -81,6 +81,10 @@ public:
     ENTER_COLOR_CHANGE,
     EXIT_COLOR_CHANGE,
     CHANGE_COLOR,
+#ifdef ENABLE_MENU_DIMBLADE
+    ENTER_DIM_CHANGE,
+    EXIT_DIM_CHANGE,
+#endif
   };
 
   // 1.0 = kDefaultVolume
@@ -196,12 +200,42 @@ public:                                                         \
     }
   }
 
+#ifdef ENABLE_MENU_DIMBLADE
+  static uint32_t GetCurrentBrightness() {
+    return current_brightness_;//value from 0~16384
+  }
+  // For smooth updates or restore.
+  static void SetBrightness(float v) {
+    current_brightness_ = (int)(v * 16384 / 100.0);  
+    current_brightness_ = clampi32(current_brightness_, 0, 16384) ;
+  }
+
+  enum DimChangeMode {
+    DIM_CHANGE_MODE_NONE,
+    DIM_CHANGE_MODE_SMOOTH
+  };
+
+  static DimChangeMode GetDimChangeMode() { return dim_change_mode_; }
+  static void SetDimChangeMode(DimChangeMode  mode) {
+    dim_change_mode_ = mode;
+    if (mode == DIM_CHANGE_MODE_NONE) {
+      DoChange(EXIT_DIM_CHANGE);
+    } else {
+      DoChange(ENTER_DIM_CHANGE);
+    }
+  }
+#endif
+
 private:
   static bool on_;
   static LockupType lockup_;
   static uint32_t last_motion_request_;
   static uint32_t current_variation_;
   static ColorChangeMode color_change_mode_;
+#ifdef ENABLE_MENU_DIMBLADE
+  static uint32_t current_brightness_;
+  static DimChangeMode dim_change_mode_;
+#endif
   SaberBase* next_saber_;
 };
 
--- a/config/default_proffieboard_config.h
+++ b/config/default_proffieboard_config.h
@@ -12,7 +12,7 @@
 #include "proffieboard_config.h"
 #define NUM_BLADES 2
 #define NUM_BUTTONS 2
-#define VOLUME 1000
+#define VOLUME 100
 const unsigned int maxLedsPerStrip = 144;
 #define CLASH_THRESHOLD_G 1.0
 #define ENABLE_AUDIO
@@ -20,11 +20,12 @@ const unsigned int maxLedsPerStrip = 144;
 #define ENABLE_WS2811
 #define ENABLE_SD
 #define ENABLE_SERIAL
+#define ENABLE_MENU_DIMBLADE
 #endif
 
 #ifdef CONFIG_PRESETS
 Preset presets[] = {
-  { "TeensySF", "tracks/venus.wav",
+  { "common;TeensySF", "tracks/venus.wav", //menu sounds are in common folder
     StyleNormalPtr<CYAN, WHITE, 300, 800>(),
     StyleNormalPtr<CYAN, WHITE, 300, 800>(), "cyan"},
   { "SmthJedi", "tracks/mars.wav",
@@ -68,7 +69,9 @@ Preset presets[] = {
     StyleNormalPtr<BLACK, BLACK, 300, 800>(), "Battery\nLevel"}
 };
 BladeConfig blades[] = {
- { 0, WS2811BladePtr<97, WS2811_800kHz>(),
+ { 0, DimBlade(56.7,WS2811BladePtr<97, WS2811_800kHz>()), // sets initial blade brightness to 100.0% 
+                                                           // can be altered via menu or serial commands 
+                                                           // when ENABLE_MENU_DIMBLADE is defined
      SimpleBladePtr<CreeXPE2WhiteTemplate<550>,
                     CreeXPE2BlueTemplate<240>,
                     CreeXPE2BlueTemplate<240>, NoLED, bladePowerPin4, bladePowerPin5, bladePowerPin6, -1>(),
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -839,6 +839,18 @@ public:
 
 #endif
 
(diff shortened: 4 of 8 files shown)


profezzorn:
Am I right that this doesn't do anything unless you have a DimBlade() in your blade definition?

profezzorn:
I think we either need to build this so that it works even if you don't have DimBlade() in your config file, or make it so that it takes zero memory if you don't have DimBlade() in your config file. Having it always take memory, but only work if you use DimBlade() is not a great option.

artandtechvof:
Embedding it to work without DimBlade() aorund you blade config, would imply that it always works for all blades.
I feel that you should be able to indiacte which blades van be dimmed or not. You might want to have your accent leds and crystal chamber remain on full brightness when dimming your main/crossguard/double blades.

I quikcly looked into the possibility of having the DimBlade function set a #define to indicate it was actually used during compile time. But I think there might be better solutions. I also agree it should not polute the code when no DimBlade is used for anyu of the configured blade.
I do not think that a #define DIMBLADE in the config top ánd still having to set DimBlade() in your blade config would be the way to go.
 
I will have a look at this, and see if I can come up with a fitting solution.

profezzorn:
An alternative would be to have DimBlade inherit from CommandParser and add the commands there. The result could be kept in a static variable inside DimBlade.  This would make the commands only be available if DimBlade is used.

Making menues based on whether dimblade is being used is trickier, I think we would need a config define for that.

artandtechvof:
How about this idea;
All functions that can use the DimBlade functions directly or via menus are to be enabled by #define ENABLE_MENU_DIMBLADE
If it is not defined, the DimBlade will remain as it is now with a static brightness set at compile time and will not have any commands to alter brightness.
When #define ENABLE_MENU_DIMBLADE is used, serial commands and menu option become available to alter brightness of the blade.
No extra code when de define is not that, and all the functionality with just 1 define.

profezzorn:
I think that's reasonable.

artandtechvof:
> I think that's reasonable.

It is in the latest commit, would you consider merging this version?

common/saber_base.h
```diff
@@ -196,12 +200,42 @@ public:                                                         \
...
+  static uint32_t GetCurrentBrightness() {
+    return current_brightness_;//value from 0~16384
+  }
+  // For smooth updates or restore.
+  static void SetBrightness(float v) {
+    current_brightness_ = (int)(v * 16384 / 100.0);  
+    current_brightness_ = clampi32(current_brightness_, 0, 16384) ;
+  }
+
+  enum DimChangeMode {
```
profezzorn, line 213: There is no need for an enum that only has two values, a "bool" will work just as well.

props/prop_base.h
```diff
@@ -839,6 +839,18 @@ class PropBase : CommandParser, Looper, protected SaberBase {
 
 #endif
 
+#ifdef ENABLE_MENU_DIMBLADE
+   if (change_brightness_){
```
profezzorn, line 843: space between ) and {

props/prop_base.h
```diff
@@ -839,6 +839,18 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+        float delta = fmodf(fusor.angle2() - current_tick_angle_, M_PI * 2);//+ or - 2*pi from start point 0
```
profezzorn, line 844: I think it would make more sense to use the up/down angle to control brightness since it doesn't loop around.
Also, maybe we should just have the brightness and color change be controlled at the same time?

props/prop_base.h
```diff
@@ -866,6 +878,23 @@ class PropBase : CommandParser, Looper, protected SaberBase {
   uint32_t last_on_time_;
 #endif
 
+#ifdef ENABLE_MENU_DIMBLADE
+  bool change_brightness_ = false;
+  bool BrightnessChangeMode(){return change_brightness_;}
```
profezzorn, line 883: lots of spaces missing:
bool BrightnessChangeMode() { return change_brightness_; }

props/prop_base.h
```diff
@@ -866,6 +878,23 @@ class PropBase : CommandParser, Looper, protected SaberBase {
   uint32_t last_on_time_;
 #endif
 
+#ifdef ENABLE_MENU_DIMBLADE
+  bool change_brightness_ = false;
```
profezzorn, line 882: Why do we need to have this both here and in SaberBase?

blades/dim_blade.h
```diff
@@ -1,13 +1,24 @@
 #ifndef BLADES_DIM_BLADE_WRAPPER_H
 #define BLADES_DIM_BLADE_WRAPPER_H
 
-class DimBladeWrapper : public BladeWrapper, BladeStyle {
+class DimBladeWrapper : public BladeWrapper, BladeStyle 
+#ifdef ENABLE_MENU_DIMBLADE
+      ,protected SaberBase
```
profezzorn, line 6: I don't see why you inherit saberbase here...

blades/dim_blade.h
```diff
@@ -1,13 +1,24 @@
...
+#endif
+      {
 public:
   DimBladeWrapper(BladeBase* blade, int fraction) {
     blade_ = blade;
+#ifdef ENABLE_MENU_DIMBLADE
+    SaberBase::SetBrightness((float)fraction/16384.0*100.0);
```
profezzorn, line 13: Actually, I think this works really weird.
Basically, it seems like the original setting of DimBlade is completely ignored.
It would make more sense to me if the brightness setting from SaberBase and the argument were multiplied.

config/default_proffieboard_config.h
```diff
@@ -12,14 +12,15 @@
 #include "proffieboard_config.h"
 #define NUM_BLADES 2
 #define NUM_BUTTONS 2
-#define VOLUME 1000
+#define VOLUME 100
```
profezzorn, line 15: why is this change here?

config/default_proffieboard_config.h
```diff
@@ -68,7 +69,9 @@ Preset presets[] = {
     StyleNormalPtr<BLACK, BLACK, 300, 800>(), "Battery\nLevel"}
 };
 BladeConfig blades[] = {
- { 0, WS2811BladePtr<97, WS2811_800kHz>(),
+ { 0, DimBlade(100.0,WS2811BladePtr<97, WS2811_800kHz>()), // sets initial blade brightness to 100.0%
```
profezzorn, line 72: why is this change here?

props/prop_base.h
```diff
@@ -866,6 +878,23 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+  bool BrightnessChangeMode(){return change_brightness_;}
+  void ToggleBrightnessChangeMode() {
```
profezzorn, line 884: This change doesn't seem to add a way to actually access this menu, except from the serial monitor.

common/saber_base.h
artandtechvof: My idea was for future implementation, were you don't want a smooth volume adjustment, but stepwise, in 10 levels for instance. Comparable to ColoChange<>.
Current implementation only allows for smooth color change. By adding DIM_CHANGE_MODE_STEPPED you could make an volume adjustment that steps trough a number of values.

props/prop_base.h
artandtechvof: From a user perspective this would yield some very akward angles to get the color ánd brightness exactly as you want, when twist is color, and up down is brightness. You might have your color correct, but then your brightness is not what you want. When adjusting your brightness color is most likely affected to.
In my opinion these adjustments should be independent.

What I am aiming for, (and already have working on my sabers) is to have a scroll-able menu, where selection of a menu and adjustment is the same for all menu's you want to add to your saber. This makes selecting and adjusting much more intuitive for an user.

In my saber I have a menu order like this :  Color menu, Volume menu, Brightness menu

What I have (for a 1 button saber);
- longheld when off: menu mode becomes active: ignites saber and selects Color Change menu, adjustment by turning hilt.
- shortclick when in menu mode: next menu: Volume menu, adjustment by twisting hilt only +/- 90 degree twist is needed to go from level 0 to MAX_VOLUME (#define VOLUME controls initial volume after startup/pragramming). Adjustment saturates on max and minlevel, no roll-over.
- shortclick when in menu mode: next menu: Brightness menu, adjusment by twisting hilt, only +/- 90 degree twist is needed to go from 0 to max brightness (saturates on max and minlevel, no roll-over)  
- shortclick when last menu item is active ; exit menu mode. saber remains on.
- longheld while in any menu: exit menu item and menu mode immediately. saber remains on.

Doing it this way, you can add as many custom menu's as you want, it will not require additional buttons for each individual menu, and it is easier to understand how to go trough menu items and adjust them.

For multi-button setups, you can still have certain menu items directly accesable, by some button combinations (like key-shortcuts). It is also up to the installer that wants to make their own propfile, how they would like to acces the menu items.
An example prop file how to use these items should be included, like the saber_fett263_buttons and saber_sa22c_buttons, or this should be added to the default saber.h prop file. 

I am preparing the pull request for the volume menu (with a #define ENABLE_MENU_VOLUME ) as a stand alone addition to the ProffieOS.

Ultimately I would like to have #define ENABLE_MENUS. This will enable ENABLE_MENU_VOLUME and ENABLE_MENU_DIMBLADE (and optinal any other menu that you build yoursels), and at the same time will supress 
 DoChange(EXIT_COLOR_CHANGE);, DoChange(EXIT_DIM_CHANGE); and DoChange(EXIT_VOL_CHANGE);
Otherwise scrolling trough the menus would trigger the exit sound of the previous and the entry sound of the current menu at the same time.
Exiting the last menu item would trigger the  DoChange(EXIT_MENU); and a generic menu exit sound or quote can be used to exit menu mode. This will allow to exit menus with the same common sound (in my case I use a spoken quote 'exiting config mode'.

To clarify this, I will make a youtube video explaining my menu mode, and post it on TRA.

props/prop_base.h
artandtechvof: You are correct. Made the adjsutment to use SaberBase::GetDimChangeMode()

blades/dim_blade.h
artandtechvof: If i don't, I can not acces SaberBase::SetBrightness((float)fraction/16384.0*100.0) ??

blades/dim_blade.h
artandtechvof: No, The setting value in your config.h file DimBlade(##.#, ...  will actually be your initial brightness for your blade. (with or without the ENABLE_MENU_DIMBLADE
It also mean that you don't have to initialize it in ProffieOS.ino, 

SaberBase::DimChangeMode SaberBase::dim_change_mode_ =
  SaberBase::DIM_CHANGE_MODE_NONE;
uint32_t SaberBase::current_brightness_; // does not need to be initialized, takes setting from config.h

config/default_proffieboard_config.h
artandtechvof: For debugging late at night  :-)
(should be reset to correct version once you are satisfied with all the adjustments for dim_blade pull request)

config/default_proffieboard_config.h
artandtechvof: For testing, should have made used own config for this? Will reset for the final version..

props/prop_base.h
artandtechvof: Please look at my earlier comment own how I would like to proceed with implementing multiple menus, and the way they are accessed.

https://github.com/profezzorn/ProffieOS/pull/88#discussion_r484763570

artandtechvof:
I n this latest commit there is an example propfile, that uses colorchange and dimblade. The volume-menu and cfg-mode are commented out, so that you can see what my idea is.

profezzorn:
This is now implemented, but differently.  See defines: DYNAMIC_BLADE_DIMMING and SAVE_BLADE_DIMMING

PR #90, opened 2020-09-08
   Addition of Mult<> and Percentage<> functions

artandtechvof:
new file:   functions/mult.h
   modified:   ProffieOS.ino

 These functions were developed as helper functions for WavLen<>.
 Before WavLen<> is finalized, these functions can be added as stand-alone functions already.

 Mult<> does an 16.15 fixedpoint multiplication of 2 values, returns Int<>. It aslo has an alias Percentage<> that returns a percentage of the input as Int<>

### DIFF
--- a/functions/mult.h
+++ b/functions/mult.h
@@ -30,13 +30,12 @@ private:
 // Usage: Percentage<F, V>
 // Gets Percentage V of value F, 
 // Percentages over 100% are allowed and will effectively be a multiplier. 
-// Can be used to multiply or devide any number  
 // F, V: INTEGER
 // example Percentage<WavLen<EFFECT_IGNITION>,25>
-//   this will give you 25% of the length of the wav file
+// this will give you 25% of the length of the wav file
 // return value: INTEGER
 
 template<class F, int V> //floats do no work in class templates...
-  using Percentage = Mult< F, Int<(int)(V*32768/100.0f)>>;
+  using Percentage = Mult<F, Int<(int)(V * 32768 / 100.0f)>>;
 
 #endif
--- a/functions/mult.h
+++ b/functions/mult.h
@@ -1,0 +1,42 @@
+#ifndef FUNCTIONS_MULT_H
+#define FUNCTIONS_MULT_H
+
+// Usage: Mult<F, V>
+// Fixed point multiplication of values F * V, 
+// fixed point 16.15 arithmetic (32768 = 1.0)
+// (2*2 would not result in 4), 
+// (16384 * 16384 = 8192, representation of 0.5*0.5=0.25) 
+// all blade functions use this method of fixed point calculations
+// F, V: INTEGER, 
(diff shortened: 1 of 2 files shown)


functions/mult.h
```diff
@@ -0,0 +1,41 @@
...
+// Usage: Percentage<F, V>
+// Gets Percentage V of value F, 
+// Percentages over 100% are allowed and will effectively be a multiplier. 
+// F, V: INTEGER
+// example Percentage<WavLen<EFFECT_IGNITION>,25>
+// this will give you 25% of the length of the wav file
+// return value: INTEGER
+
+template<class F, int V> //floats do no work in class templates...
+  using Percentage = Mult<F, Int<(int)(V * 32768 / 100.0f)>>;
```
profezzorn, line 39: Does this actually work? I'm not sure float arithmetics are allowed in template resolution.
Also, the float arithmetics makes basically no difference here, so I think we can just make it V * 32768 / 100

functions/mult.h
```diff
@@ -0,0 +1,41 @@
+#ifndef FUNCTIONS_MULT_H
+#define FUNCTIONS_MULT_H
+
+// Usage: Mult<F, V>
+// Fixed point multiplication of values F * V, 
+// fixed point 16.15 arithmetic (32768 = 1.0)
+// (2*2 would not result in 4), 
+// (16384 * 16384 = 8192, representation of 0.5*0.5=0.25) 
+// all blade functions use this method of fixed point calculations
```
profezzorn, line 9: all -> most

Some functions returns milliseconds.

functions/mult.h
```diff
@@ -0,0 +1,41 @@
...
+private:
+  F f_;
+  V v_;
+};
+
+// Usage: Percentage<F, V>
+// Gets Percentage V of value F, 
+// Percentages over 100% are allowed and will effectively be a multiplier. 
+// F, V: INTEGER
+// example Percentage<WavLen<EFFECT_IGNITION>,25>
```
profezzorn, line 34: Let's not use an example that isn't checked in yet...

functions/mult.h
artandtechvof: Yes it does, I was wondering that myself, and tried it with an without explicit declaring Int<(int)((float)V * 32768 / 100.0f)
Worked in both cases. Might depend on compiler optimization? To be sure we could make it an (float)V anyway.
(or just leave out the float altogether, like you suggest)

profezzorn: please just get rid of the float math. :)

functions/mult.h
artandtechvof: (just lobbying...)

artandtechvof:
Made some adjustments to the code, hope it is acceptabel this way.

profezzorn:
Needs an #include line in ProffieOS.ino, then I can merge it.

PR #91, opened 2020-09-09
Added Location to ResponsiveDrag, Melt and Stab

fdarosa2663:
To give users more control over location of SmoothStep<> for the Responsive "tip" effects, LOCATION defaults to original settings so it's backwards compatible.


profezzorn:
I'm confused, how can the location be anything other than the tip for these actions?

fdarosa2663:
Sorry, If users want to move further from the tip or closer to the tip (or to use a Function) this allows it.  The SIZE1 and SIZE2 values only affect the 2nd value in SmoothStep which increases the "step" area (which has a larger or smaller gradient as a result, but some users want to move the first value in the SmoothStep to keep the "step" from having too much gradient but still increase or decrease the distance from the tip, so the LOCATION allows the first value in SmoothStep to be controlled by user in addition to the 2nd value.

styles/responsive_styles.h
```diff
@@ -44,13 +46,15 @@ using ResponsiveDragL =
...
 // TRANSITION1 & TRANSITION2 = transition Begin & End
 // SIZE1 & SIZE2 control MELT area limits for TwistAngle
+// LOCATION control SmoothStep location
 template<class COLOR = Mix<TwistAngle<>,OrangeRed,Red>,
          class TR1 = TrWipeIn<600>,
          class TR2 = TrWipe<600>,
          class SIZE1 = Int<4000>,
-         class SIZE2 = Int<10000>>
+         class SIZE2 = Int<10000>,
+         class LOCATION - Int<30000>>
```
profezzorn, line 55: Did you test this?
Because it says - instead of = here and I don't think that will compile.

fdarosa2663: My mistake, typo. I fixed in my branch does it show up or do I edit it here too?

PR #92, opened 2020-09-10
Adding Gesture Prop File with "Battle Mode" and "Smart Lockup/Drag/Melt"

fdarosa2663:
New prop file to enable the new gesture features, working on write up, demo video and changes to library to support testing.


props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+
+#ifndef MOTION_TIMEOUT
+#define MOTION_TIMEOUT 60 * 15 * 1000
+#endif
+
+#ifndef LOCKUP_THRESHOLD_G
+#define LOCKUP_THRESHOLD_G 2.0
+#endif 
+
+#ifndef PROPS_SABER_FETT263_G_BUTTONS_H
```
profezzorn, line 57: This should be first. (Either at the top of the file, or right after the comments.)

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+
+#include "prop_base.h"
+#include "../sound/hybrid_font.h"
+
+#undef PROP_TYPE
+#define PROP_TYPE SaberFett263Buttons
+
+// The Saber class implements the basic states and actions
+// for the saber.
+class SaberFett263Buttons : public PropBase {
```
profezzorn, line 68: SaberFett263Buttons -> SaberFett263GButtons

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+public:
+SaberFett263Buttons() : PropBase() {}
+  const char* name() override { return "SaberFett263Buttons"; }
+
+
```
profezzorn, line 73: extra empty line

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+  bool swinging_ = false;
+  void Loop() override {
+    PropBase::Loop();
+    if(!SaberBase::IsOn() ) {
+      if (millis() - saber_off_time_ < MOTION_TIMEOUT) {
+        SaberBase::RequestMotion();
+        if (!swinging_ && fusor.swing_speed() > 250) {
+            swinging_ = true;
+            Event(BUTTON_NONE, EVENT_SWING);
+            }
```
profezzorn, line 84: indented wrong

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+            if (swinging_ && fusor.swing_speed() < 100) {
+            swinging_ = false;
+            }
+      }
+    }
+    if (!swinging_ && fusor.swing_speed() > 250) {
```
profezzorn, line 90: Why do this code twice?

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+      swinging_ = true;
+      Event(BUTTON_NONE, EVENT_SWING);
+    }
+    if (!swinging_ && fusor.swing_speed() > 120 && fusor.swing_speed() < 250 && millis() > clash_swing_ + (100 * LOCKUP_THRESHOLD_G)) {
```
profezzorn, line 94: Please add comments explaining this if statement

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+    
+    if(auto_lockup_on_) {
```
profezzorn, line 96: space after if

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+          if (SaberBase::Lockup()) {
+          SaberBase::DoEndLockup();
```
profezzorn, line 98: indented wrong

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+  // SA22C Volume Menu
+  void VolumeUp() {
+    STDOUT.println("Volume up");
+    if (dynamic_mixer.get_volume() < VOLUME) {
+      dynamic_mixer.set_volume(std::min<int>(VOLUME + VOLUME * 0.1,
+					     dynamic_mixer.get_volume() + VOLUME * 0.10));
+      beeper.Beep(0.5, 2000);
+      STDOUT.print("Current Volume: ");
+      STDOUT.println(dynamic_mixer.get_volume());
+    }
```
profezzorn, line 132: } else {

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+  void VolumeDown() {
+    STDOUT.println("Volume Down");
+    if (dynamic_mixer.get_volume() > (0.10 * VOLUME)) {
+      dynamic_mixer.set_volume(std::max<int>(VOLUME * 0.1,
+					     dynamic_mixer.get_volume() - VOLUME * 0.10));
+      beeper.Beep(0.5, 2000);
+      STDOUT.print("Current Volume: ");
+      STDOUT.println(dynamic_mixer.get_volume());
+    }
+    else{
```
profezzorn, line 146: } else {

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+          // Don't turn saber off.
+          ToggleColorChangeMode();
+          return true;
+        }
+#endif
+        if (!swinging_ && !battle_mode_) {
+          swing_blast_ = false;
+          Off();
+          saber_off_time_ = millis();
+  }
```
profezzorn, line 234: indented wrong

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+      case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON):
+	      if (swing_blast_) {
+	      SaberBase::DoBlast();
+	     }
+           return true;
+      
+#ifndef DISABLE_COLOR_CHANGE
+      case EVENTID(BUTTON_POWER, EVENT_CLICK_SHORT, MODE_ON | BUTTON_AUX):
+        ToggleColorChangeMode();
+	return true;
```
profezzorn, line 263: indented wrong

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+        SaberBase::DoBlast();
+        return true;
+	
+      case EVENTID(BUTTON_AUX, EVENT_CLICK_LONG, MODE_ON):
+        swing_blast_ = true;
+        hybrid_font.SB_Blast();
+        return true;
+
+      case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON):
+	      if (swing_blast_) {
```
profezzorn, line 255: indented wrong

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+        return true;
+	
+      case EVENTID(BUTTON_POWER, EVENT_CLICK_SHORT, MODE_OFF | BUTTON_AUX):
+        previous_preset();
+        return true;
+
+// Gesture Controls
+
+// Battle Mode
+case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON | BUTTON_AUX):
```
profezzorn, line 331: indented wrong

props/saber_fett263_G_buttons.h
profezzorn, line 94: This seems really weird:  millis() > clash_swing_ + (100 * LOCKUP_THRESHOLD_G)
What it seems to do is to put a maximum length on the lockup?
But why use a define called LOCKUP_THRESHOLD_G for that? The _G indicates that the unit is earth gravities, not deciseconds.  (And why are we using deciseconds?)

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+    return false;
+  }
+private:
+  bool pointing_down_ = false;
+  bool swing_blast_ = false;
+  bool mode_volume_ = false;
+  bool auto_lockup_on_ = false;
+  bool auto_melt_on_ = false;
+  bool battle_mode_ = false;
+  uint32_t clash_swing_ = millis();
```
profezzorn, line 439: clash swing?
What does that mean?
Can we come up with a better name?

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+          SaberBase::DoEndLockup();
+          SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+          auto_melt_on_ = false;        
+          } 
+        }
+    }
+
+ }
+  
+
```
profezzorn, line 122: extra empty line

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
...
+    if(auto_melt_on_) {
+       if (!swinging_ && fusor.swing_speed() > 60 && fusor.swing_speed() < 120 && millis() > clash_swing_ + (100 * LOCKUP_THRESHOLD_G)) {  
+          if (SaberBase::Lockup()) {
+          SaberBase::DoEndLockup();
+          SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+          auto_melt_on_ = false;        
+          } 
+        }
+    }
+
```
profezzorn, line 119: extra empty line

props/saber_fett263_G_buttons.h
```diff
@@ -0,0 +1,445 @@
+// Fett263 Buttons with Gesture Controls and Battle Mode with Smart Lockup
```
profezzorn, line 1: I wonder if it would have been better to just add all of this to the existing fett263 prop file and use defines to turn features on and off?

props/saber_fett263_G_buttons.h
fdarosa2663: One is for EVENT_SWING while on, the second is for the SWING ON ignition, some users want to increase the speed needed to ignite the blade so having the version for !SaberBase::InOn() allows without changing the EVENT_SWING  for other uses while ON, like the Multi-Blast.

props/saber_fett263_G_buttons.h
fdarosa2663: This allows "fine tuning" for the physical bounce clashing two sabers creates, I misunderstood the "G" from the CLASH_THRESHOLD_G, I have changed to LOCKUP_THRESHOLD_DELAY.  This delay gives the user time to steady their saber to keep lockup active before the end lockup begins.  After the actual clash this delay "ignores" the pull away that gets created if the sabers bounce off object.  Depending on the hilt and user they will need to be able to adjust this delay.

props/saber_fett263_G_buttons.h
fdarosa2663: I renamed to clash_impact_, it's used to give a delay before checking next gesture

props/saber_fett263_G_buttons.h
fdarosa2663: For backwards compatible I was worried with all the new gestures it may confuse users, there are a lot of gesture additions to the G file that are related and thought too many defines would be cumbersome so having the "G" was to signify the addition of the gestures so users just pick prop and use with fewer defines.  If you prefer to reduce the number of prop files I can modify and we can just make a single with multiple defines, I am ok with either approach.

props/saber_fett263_G_buttons.h
fdarosa2663: fixed

props/saber_fett263_G_buttons.h
fdarosa2663: fixed

props/saber_fett263_G_buttons.h
profezzorn: then there needs to be comments in the code that clarifies that, otherwise it just looks redundant.

props/saber_fett263_G_buttons.h
profezzorn: If all the new defines default to "off", then backwards compatibility is not a problem.
From what I can tell from this file, it looks like you have most or all of those defines anyways.
Since this seems to be a natural extension of the other prop file, I think it would be better to just work on that file.
If this was a completely new, approach, with very little shared functionality, then a new file would be best.

props/saber_fett263_G_buttons.h
profezzorn: clash impact isn't a whole lot better, because it doesn't give any clue as to what the unit of the variable is.
Maybe last_clash_impact_  or clash_impact_millis_?

props/saber_fett263_G_buttons.h
profezzorn: LOCKUP_THRESHOLD_DELAY is better, but why should it be specifies the delay in deciseconds? It should either be in seconds or in milliseconds I think.

props/saber_fett263_G_buttons.h
fdarosa2663: OK, makes sense I'll make edits and then remove this file and update existing saber_fett263_buttons.h

props/saber_fett263_G_buttons.h
fdarosa2663: changed to millis

props/saber_fett263_buttons.h
```diff
@@ -1,10 +1,32 @@
...
-// Includes "Multi-Blast" Mode to enable Swing Blast control and SA22C volume menu
+// Fett263 Buttons with Gesture Controls and Battle Mode with Smart Lockup
+//
+// Includes Gesture Controls, "Battle Mode" with "Smart Lockup", "Multi-Blast" Mode (to enable Swing Blast control) and SA22C volume menu
+// Optional Defines
+//
+// FETT263_BATTLE_MODE
+// To enable Battle Mode, Swing On and Twist Off gestures
+//
+// LOCKUP_THRESHOLD_DELAY 200
```
profezzorn, line 9: Should start with FETT263_

props/saber_fett263_buttons.h
```diff
@@ -1,10 +1,32 @@
...
+// This is the "delay" threshold in millis to determine Clash vs Lockup
+//
+// FETT263_STAB_ON 
+// To enable Stab On
+//
+// FETT263_TWIST_ON 
+// To enable Twist On
+//
+// FETT263_MULTI_PHASE
+// This will enable "live" preset change to create a "Multi-Phase" saber effect with preset changes on the fly while blade is ignited.
```
profezzorn, line 19: Line is too long.

props/saber_fett263_buttons.h
```diff
@@ -19,40 +41,87 @@
...
   // EVENT_SWING
   bool swinging_ = false;
   void Loop() override {
     PropBase::Loop();
+    // Swing On gesture control this portion allows fine tuning of speed needed to ignite
+    if(!SaberBase::IsOn() ) {
+      if (millis() - saber_off_time_ < MOTION_TIMEOUT) {
+        SaberBase::RequestMotion();
+        if (!swinging_ && fusor.swing_speed() > 250) {
+	// Edit '250' value in line above to change swing on sensitivity, 250 ~ 400 work best in testing
```
profezzorn, line 85: comment is indented wrong and should go before the if statement

props/saber_fett263_buttons.h
```diff
@@ -19,40 +41,87 @@
...
+        }
+      }
+    }
+	  
     if (!swinging_ && fusor.swing_speed() > 250) {
       swinging_ = true;
       Event(BUTTON_NONE, EVENT_SWING);
     }
+	  
+    if(auto_lockup_on_) {
```
profezzorn, line 100: space after if

props/saber_fett263_buttons.h
```diff
@@ -19,40 +41,87 @@
...
+    if(!SaberBase::IsOn() ) {
+      if (millis() - saber_off_time_ < MOTION_TIMEOUT) {
+        SaberBase::RequestMotion();
+        if (!swinging_ && fusor.swing_speed() > 250) {
+	// Edit '250' value in line above to change swing on sensitivity, 250 ~ 400 work best in testing
+           swinging_ = true;
+           Event(BUTTON_NONE, EVENT_SWING);
+        }
+        if (swinging_ && fusor.swing_speed() < 100) {
+           swinging_ = false;
```
profezzorn, line 90: all indentation should be two steps, not 3

props/saber_fett263_buttons.h
```diff
@@ -19,40 +41,87 @@
...
+          if (SaberBase::Lockup()) {
+             SaberBase::DoEndLockup();
+             SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+             auto_lockup_on_ = false;        
+          } 
+        }
+    }
+	  
     if (swinging_ && fusor.swing_speed() < 100) {
+      last_swing_ = millis();
```
profezzorn, line 111: might be better to do:

if (swinging_) last_swing_ = millis();  (outside of this if statement)

props/saber_fett263_buttons.h
```diff
@@ -19,40 +41,87 @@
...
       swinging_ = false;
+    }   
+    
+    if(auto_melt_on_) {
```
profezzorn, line 115: space after if

props/saber_fett263_buttons.h
```diff
@@ -19,40 +41,87 @@
...
+       if (!swinging_ && fusor.swing_speed() > 60 && fusor.swing_speed() < 120 && millis() > clash_impact_millis_ + LOCKUP_THRESHOLD_DELAY) {  
+          if (SaberBase::Lockup()) {
+          SaberBase::DoEndLockup();
```
profezzorn, line 118: not indented

props/saber_fett263_buttons.h
```diff
@@ -100,7 +167,7 @@ SaberFett263Buttons() : PropBase() {}
         if (mode_volume_) {
           VolumeUp();
         } else {
-          On();
+        On();
```
profezzorn, line 170: not indented correctly

props/saber_fett263_buttons.h
```diff
@@ -183,13 +251,13 @@ SaberFett263Buttons() : PropBase() {}
...
         return true;
-	
+
       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON):
-	if (swing_blast_) {
-	  SaberBase::DoBlast();
-	}
-	return true;
-	
+	      if (swing_blast_) {
```
profezzorn, line 256: indented too much

props/saber_fett263_buttons.h
```diff
@@ -161,10 +228,11 @@ SaberFett263Buttons() : PropBase() {}
           return true;
         }
 #endif
-        if (!swinging_) {
+        if (!swinging_ && !battle_mode_) {
           swing_blast_ = false;
           Off();
-        }
+          saber_off_time_ = millis();
+  }
```
profezzorn, line 235: not indented correctly

props/saber_fett263_buttons.h
```diff
@@ -257,7 +325,100 @@ SaberFett263Buttons() : PropBase() {}
       case EVENTID(BUTTON_POWER, EVENT_CLICK_SHORT, MODE_OFF | BUTTON_AUX):
         previous_preset();
         return true;
+
+       // Gesture Controls
+		
+       // Battle Mode
+		    
+       #ifdef FETT263_BATTLE_MODE
```
profezzorn, line 333: All preprocessor statements (like #ifdef) should be un-indented (start at the beginning of the line)

props/saber_fett263_buttons.h
```diff
@@ -257,7 +325,100 @@ SaberFett263Buttons() : PropBase() {}
...
+       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON | BUTTON_AUX):
+           if (!battle_mode_) {
```
profezzorn, line 335: lots of things here are indented too much (should be two spaces per step)

props/saber_fett263_buttons.h
```diff
@@ -257,7 +325,100 @@ SaberFett263Buttons() : PropBase() {}
...
+              if (clash_impact_millis_ > last_swing_) {
+                 SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
+                 swing_blast_ = false;
+                 auto_lockup_on_ = true;
+                 SaberBase::DoBeginLockup();
+                 return true;
+               }
+             } else {
+               return false;
+           }
```
profezzorn, line 358: This looks wrong.
The } matches up with the else on line 356, but is indented like it's supposed to match up with the if statement on 346.
I think there is supposed to be two } here...

props/saber_fett263_buttons.h
```diff
@@ -19,40 +41,87 @@
...
+      }
+    }
+	  
     if (!swinging_ && fusor.swing_speed() > 250) {
       swinging_ = true;
       Event(BUTTON_NONE, EVENT_SWING);
     }
+	  
+    if(auto_lockup_on_) {
+       if (!swinging_ && fusor.swing_speed() > 120 && fusor.swing_speed() < 250 && millis() > clash_impact_millis_ + LOCKUP_THRESHOLD_DELAY) {
```
profezzorn, line 101: millis() > clash_impact_millis_ + LOCKUP_THRESHOLD_DELAY is going to cause problems when millis() wraps around

Instead, do: millis() - clash_impact_millis_ > LOCKUP_THRESHOLD_DELAY

props/saber_fett263_buttons.h
```diff
@@ -257,7 +325,100 @@ SaberFett263Buttons() : PropBase() {}
...
+              hybrid_font.SB_Force();
+           }
+           return true;
+
+       // Auto Lockup Mode
+       case EVENTID(BUTTON_NONE, EVENT_CLASH, MODE_ON):
+           if (battle_mode_) {
+              clash_impact_millis_ = millis();
+              swing_blast_ = false;
+              if (clash_impact_millis_ > last_swing_) {
```
profezzorn, line 349: I'm not sure I understand what this is supposed to do.
Maybe this is supposed to be if (!swinging_) { ??

props/saber_fett263_buttons.h
```diff
@@ -1,10 +1,32 @@
-// Fett263 Buttons
-// Includes "Multi-Blast" Mode to enable Swing Blast control and SA22C volume menu
+// Fett263 Buttons with Gesture Controls and Battle Mode with Smart Lockup
+//
+// Includes Gesture Controls, "Battle Mode" with "Smart Lockup", "Multi-Blast" Mode (to enable Swing Blast control) and SA22C volume menu
```
profezzorn, line 3: line is too long

props/saber_fett263_buttons.h
```diff
@@ -257,7 +325,100 @@ SaberFett263Buttons() : PropBase() {}
...
+       case EVENTID(BUTTON_NONE, EVENT_STAB, MODE_ON):
+           if (battle_mode_) {
+              clash_impact_millis_ = millis();
+              swing_blast_ = false;
+              if (clash_impact_millis_ > last_swing_) {
+                 if ((fusor.angle1() + M_PI / 2) * (32768 / M_PI) < 8000) {
+                   SaberBase::SetLockup(SaberBase::LOCKUP_DRAG);
+                 } else {
+                   SaberBase::SetLockup(SaberBase::LOCKUP_MELT);
+       	         }
```
profezzorn, line 369: bad indentation

props/saber_fett263_buttons.h
```diff
@@ -257,7 +325,100 @@ SaberFett263Buttons() : PropBase() {}
...
+              swing_blast_ = false;
```
profezzorn, line 370: bad indentation

props/saber_fett263_buttons.h
```diff
@@ -257,7 +325,100 @@ SaberFett263Buttons() : PropBase() {}
...
+             } else {
+               return false;
+           }
+
+       case EVENTID(BUTTON_NONE, EVENT_STAB, MODE_ON):
+           if (battle_mode_) {
+              clash_impact_millis_ = millis();
+              swing_blast_ = false;
+              if (clash_impact_millis_ > last_swing_) {
+                 if ((fusor.angle1() + M_PI / 2) * (32768 / M_PI) < 8000) {
```
profezzorn, line 365: I think this should be:

  if (fuser.angle1() < - M_PI / 2) {

props/saber_fett263_buttons.h
```diff
@@ -257,7 +325,100 @@ SaberFett263Buttons() : PropBase() {}
...
 	
+       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_OFF):  
+           if(millis() > 3000) { 
+              On();
+              battle_mode_ = true;
+           }
+           return true;
+
+       case EVENTID(BUTTON_NONE, EVENT_TWIST, MODE_ON):
+           if (millis() > last_twist_ + 3000) {
```
profezzorn, line 387: millis() - last_twist_ > 3000

props/saber_fett263_buttons.h
fdarosa2663: changed to FETT263_LOCKUP_DELAY

props/saber_fett263_buttons.h
fdarosa2663: It's meant to be the time at the end of the swing or at clash so when swing ends this is the time set so if there are loud swings like accents or spins it doesn't activate the lockup/clash effect, only if the blade physically stops swinging can they trigger

props/saber_fett263_buttons.h
fdarosa2663: it's if (battle_mode) do all up to return true, else return false so normal Clash occurs outside prop file

props/saber_fett263_buttons.h
fdarosa2663: this is checking the swing actual stopped (avoid clashes from sound effects like accent swing and/or spins) the blade has to impact (stop) to actually trigger, when you hit something your swing is stopped,

props/saber_fett263_buttons.h
fdarosa2663: Yeah I could use your help on the math here, it's probably "ham-fisted" but it worked.  Essentially I only want Drag if the blade is 0 to 45 degrees (approximately), to allow for slightly downward Melt actions to be used.  This equation produces the control I want but if there's a more efficient way, I believe M_PI / 2 would be 90 degrees so would M_PI/4 be 45 degrees? The angle stuff has been too long since I used in high school so it still trips me up.

props/saber_fett263_buttons.h
profezzorn: then why not just use !swinging_ ?

props/saber_fett263_buttons.h
profezzorn: M_PI /2 is 45 degrees

props/saber_fett263_buttons.h
```diff
@@ -257,7 +326,102 @@ SaberFett263Buttons() : PropBase() {}
...
+             auto_lockup_on_ = true;
+             SaberBase::DoBeginLockup();
+             return true;
+	   }
+         } else {
+           return false;
+         }
+
+       case EVENTID(BUTTON_NONE, EVENT_STAB, MODE_ON):
+         if (battle_mode_) {
```
profezzorn, line 362: for simplicity (and easier indentation) it might be better to do:

if (!battle_mode_) return false;
// rest of code here

props/saber_fett263_buttons.h
```diff
@@ -257,7 +326,102 @@ SaberFett263Buttons() : PropBase() {}
...
+           clash_impact_millis_ = millis();
+           swing_blast_ = false;
+           if (clash_impact_millis_ > last_swing_) {
```
profezzorn, line 365: This is almost always true.
Only time it is not is if last_swing_ was set less than a millisecond ago, and that only happens if a swing terminated very recently. (Unless I'm missing something)

props/saber_fett263_buttons.h
```diff
@@ -257,7 +326,102 @@ SaberFett263Buttons() : PropBase() {}
...
+
+       // Gesture Controls
+		
+       // Battle Mode
+		    
+#ifdef FETT263_BATTLE_MODE
+       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON | BUTTON_AUX):
+         if (!battle_mode_) {
+           battle_mode_ = true;
+           hybrid_font.SB_Force();
```
profezzorn, line 338: This is weird. Why play the Force sound when battle mode is enabled/disabled?

props/saber_fett263_buttons.h
```diff
@@ -45,14 +78,51 @@ SaberFett263Buttons() : PropBase() {}
...
+      }
+    }
+	  
     if (swinging_ && fusor.swing_speed() < 100) {
+      last_swing_ = millis();
       swinging_ = false;
+    }   
+    
+    if (auto_melt_on_) {
+      if (!swinging_ && fusor.swing_speed() > 60 && fusor.swing_speed() < 120 && millis() - clash_impact_millis_ > FETT263_LOCKUP_DELAY) {
```
profezzorn, line 117: this line is too long, I suggest breaking it up like:

 if (!swinging_ &&
     fusor.swing_speed() > 60 &&
     fusor.swing_speed < 120 &&
     millis() - clash_imoact_millis_ > FETT263_LOCKUP_DELAY) {
      ....
 }

also, if (A) {
      if (B) {

can be written as:

   if (A && B)

so, it should probably be:

   if (auto_melt_on_ &&
        !swinging_ &&
        ....

props/saber_fett263_buttons.h
fdarosa2663: perfect then, thank you

props/saber_fett263_buttons.h
fdarosa2663: It was to let user know it toggled but wanted it to be something that would make some "in universe" sense, like using the force to prepare for battle or recover from battle.

props/saber_fett263_buttons.h
fdarosa2663: OK, I'll test it with !swinging_ just to make sure it works as expected

props/saber_fett263_buttons.h
fdarosa2663: Not sure why but this does not work as expected, the previous equation worked perfectly with fusor.angle1() < -M_PI / 2 I cannot trigger Drag at all. reverted to original and it runs as expected with Drag only triggering if pointing straight down or up 8:00, I've changed everything else and it's all running, this change did not work so I left out.

profezzorn: My fault, you were right, it should be M_PI / 4

fdarosa2663: forgot to update, I changed to 

fusor.angle1() < - M_PI / 4 

and it's working, updated the file

props/saber_fett263_buttons.h
```diff
@@ -257,7 +331,96 @@ SaberFett263Buttons() : PropBase() {}
...
+	 }
+
+       case EVENTID(BUTTON_NONE, EVENT_STAB, MODE_ON):
+         if (!battle_mode_) return false;
+         clash_impact_millis_ = millis();
+         swing_blast_ = false;
+         if (!swinging_) {
+           if (fusor.angle1() < - M_PI / 4) {
+             SaberBase::SetLockup(SaberBase::LOCKUP_DRAG);
+             } else {
```
profezzorn, line 370: not indented right (goes for most of the lines from here to the next case)

props/saber_fett263_buttons.h
```diff
@@ -257,7 +331,96 @@ SaberFett263Buttons() : PropBase() {}
...
+               SaberBase::SetLockup(SaberBase::LOCKUP_MELT);
+       	     }
+         swing_blast_ = false;
+         auto_melt_on_ = true;
+         SaberBase::DoBeginLockup();
+         return true;
+         }
 	
+       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_OFF):  
+         if(millis() > 3000) {
```
profezzorn, line 380: Please add a comment to explain this, it looks weird, but I think maybe it's correct?
Although shouldn't swing-on be controlled by a define?

props/saber_fett263_buttons.h
```diff
@@ -257,7 +331,96 @@ SaberFett263Buttons() : PropBase() {}
...
+         }
 	
+       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_OFF):  
+         if(millis() > 3000) { 
+           On();
+           battle_mode_ = true;
+         }
+         return true;
+
+       case EVENTID(BUTTON_NONE, EVENT_TWIST, MODE_ON):
```
profezzorn, line 386: So let's see if I get this right....

if you enable FETT263_BATTLE_MODE, then you use swing to turn the saber on, and twist to turn it off?
While I think using swing to turn it on is kind of cool, it seems like a feature that is completely separate from battle mode to me.  Assuming battle mode works well, it would make the saber work a lot more like it does in the movies, but swing on/twist off is not how it works in the movies....

Anyways, this is completely up to you, but I think it would be cool if swing-on was not required for battle mode.
For my purposes, I would probably want a define that turns on battle mode all the time instead of having it be a setting.

props/saber_fett263_buttons.h
```diff
@@ -45,14 +78,56 @@ SaberFett263Buttons() : PropBase() {}
...
+	  
     if (!swinging_ && fusor.swing_speed() > 250) {
       swinging_ = true;
       Event(BUTTON_NONE, EVENT_SWING);
     }
+	  
+    if (auto_lockup_on_ && 
+    !swinging_ && 
+    fusor.swing_speed() > 120 && 
+    fusor.swing_speed() < 250 &&
```
profezzorn, line 104: So, if I understand this correctly, this is intended to create a clash-like effect by turning on lockup for 200ms?
However, if the saber is held relatively still, we keep the lockup effect going until it moves quickly, right?
But why are we checking if swing_speed is less than 250? I mean, if the saber is moving fast we should end lockup, right?

props/saber_fett263_buttons.h
fdarosa2663: Battle Mode has two activation methods, if the blade is on you can hold Aux and Swing to toggle on/off, if the blade is off and you use any gesture ignition it is also turned on (Swing On, Twist On or Stab On all activate) - the idea is if you use a gesture to ignite the controls automatically switch to the gesture "Battle Mode" if you use the power button it will be regular unless you activated battle mode previously but you can enter or exit Battle Mode anytime with Aux + Swing.
The Twist Off is a "safety" feature in Battle Mode to prevent accidental retraction, you can also just exit Battle Mode any press Power to turn off.  
I'm keeping it a "mode" at this time because it essentially removes Stab and Clash effects but if testing proves out then we can add a define to have it always on easily enough, because it's a significant change I wanted to take one step at a time.  Personally, I will keep it as a mode for my use but if users just want it default it can certainly be accomodated.  The controls can still be modified as more users test and give feedback as well, I'm certainly open to more input, these were the best controls based on what was available and what felt most natural while testing.
I'll make other fixes later today.

props/saber_fett263_buttons.h
fdarosa2663: technically, clash is made up of BEGIN_LOCKUP and END_LOCKUP, the LOCKUP_DELAY is to allow time for the user to steady the blade, especially on harder clashes as you get a "bounce" on impact the would cause END_LOCKUP before you could steady, this varies a little by user and hilt I found so I wanted to let users increase if they are hitting harder or get more bounce in their impacts so they can still get lockup time to initiate.  The BEGIN_LOCKUP effect "covers up" for this time, I will be describing it in the write up, if you increase the delay you will want to also increase the time the BEGIN_LOCKUP effect appears to conceal the lockup from showing.
The less than 250 came from a compile error during testing, because we were checking for > 250 in an earlier if statement (I believe) it wasn't compiling until I put in the < 250.  I forgot specific error but I couldn't get past it initially and this addition fixed it.  The > 120 is what really controls it, I mean you can't go from 0 to 250 without becoming greater then 120.  The 120 is the result from a lot of trial and error to find the "Goldilocks" movements that were easy to perform but also definitive enough to tell the difference between moving the saber in a lockup with opponent and "pulling away" to end it, too low and lockup became too hard to initiate too high and the movement needed to end lockup became too exaggerated and difficult to perform, the combination was the result of a lot of testing and fine-tuning to get it "just right".

profezzorn: please remove the fusor.swing_speed() < 250, it doesn't do anything and it's confusing to the reader

props/saber_fett263_buttons.h
profezzorn: but how do you know if it's turning on or off?

props/saber_fett263_buttons.h
profezzorn: This is exactly why we have prop files though. You can implement this any way you want to, and I can implement it any way I want to in my prop file.... :)

fdarosa2663: For alpha testing I propose it just as a "mode", then from there determine best way to allow users to have it be default.  I was originally trying to make it so you could have it be either a mode or always on with defines but I couldn't get it to work so I will probably need some assistance in having defines "share" functionality.  I want to get the functionality tested and proven then having it be a mode or always on once it's working optimally we can look at along with making a few other controls user defined.  I was hitting some obstacles so probably need your expertise to take it to that level.

props/saber_fett263_buttons.h
fdarosa2663: I added the define back in

props/saber_fett263_buttons.h
fdarosa2663: good point, exiting will play Exit Color Change sound instead

props/saber_fett263_buttons.h
fdarosa2663: not clear on this I have 2 spaces for each level below the case, are you meaning everything has to be indented further?

profezzorn: every { means that indentation should increase by two spaces
every } means that indentation should decrease by two spaces

an if statement with an else, should look like:

if (.....) {
  code
} else {
  code
}

The if and the two } should always be in the same column.
Having the columns line up makes it possible to see where a block of code begins and ends easily and thus makes the code much easier to read.

props/saber_fett263_buttons.h
```diff
@@ -45,14 +85,53 @@ SaberFett263Buttons() : PropBase() {}
...
-    }
-    if (swinging_ && fusor.swing_speed() < 100) {
-      swinging_ = false;
-    }
-  }
+      // Swing On gesture control this portion allows fine tuning of speed needed to ignite
+      if(!SaberBase::IsOn()) {
+        if (millis() - saber_off_time_ < MOTION_TIMEOUT) {
+          SaberBase::RequestMotion();
+        // Edit '250' value in line below to change swing on sensitivity, 250 ~ 400 work best in testing
```
profezzorn, line 92: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -45,14 +85,53 @@ SaberFett263Buttons() : PropBase() {}
...
+          !swinging_ && 
+          fusor.swing_speed() > 60 && 
+          millis() - clash_impact_millis_ > FETT263_LOCKUP_DELAY) {  
+        if (SaberBase::Lockup()) {
+          SaberBase::DoEndLockup();
+          SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+          auto_melt_on_ = false;        
+        } 
+      }
+
```
profezzorn, line 133: extra empty line

props/saber_fett263_buttons.h
```diff
@@ -257,7 +335,104 @@ SaberFett263Buttons() : PropBase() {}
...
         previous_preset();
         return true;
-	
+
+       // Gesture Controls
+		
+       // Battle Mode
+		    
+#ifdef FETT263_BATTLE_MODE
+  #define FETT263_SWING_ON
```
profezzorn, line 344: #defines should not be indented

props/saber_fett263_buttons.h
```diff
@@ -257,7 +335,104 @@ SaberFett263Buttons() : PropBase() {}
...
+  #define FETT263_TWIST_OFF
+       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON | BUTTON_AUX):
+         if (!battle_mode_) {
+           battle_mode_ = true;
+           // Force sound plays when entering Battle Mode
+           hybrid_font.SB_Force();
+         } else {
+           battle_mode_ = false;
+	   // Exit Color Change sound plays when exiting Battle Mode
```
profezzorn, line 353: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -257,7 +335,104 @@ SaberFett263Buttons() : PropBase() {}
...
+         if (!battle_mode_) return false;
+         clash_impact_millis_ = millis();
+         swing_blast_ = false;
+         if (!swinging_) {
+           SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
+           swing_blast_ = false;
+           auto_lockup_on_ = true;
+           SaberBase::DoBeginLockup();
+           return true;
+	 }
```
profezzorn, line 369: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -257,7 +335,104 @@ SaberFett263Buttons() : PropBase() {}
...
+
```
profezzorn, line 370: shouldn't there be a return here?

fdarosa2663: added

profezzorn: Now you can return the return from two lines up though, as they are the same.

props/saber_fett263_buttons.h
```diff
@@ -257,7 +330,106 @@ SaberFett263Buttons() : PropBase() {}
...
+	 }
+         return true;
+
+       case EVENTID(BUTTON_NONE, EVENT_STAB, MODE_ON):
+         if (!battle_mode_) return false;
+         clash_impact_millis_ = millis();
+         swing_blast_ = false;
+         if (!swinging_) {
+           if (fusor.angle1() < - M_PI / 4) {
+               SaberBase::SetLockup(SaberBase::LOCKUP_DRAG);
```
profezzorn, line 373: indented too much (two extra steps)

props/saber_fett263_buttons.h
```diff
@@ -257,7 +330,106 @@ SaberFett263Buttons() : PropBase() {}
...
+           } else {
+               SaberBase::SetLockup(SaberBase::LOCKUP_MELT);
```
profezzorn, line 375: indented too much (two extra steps)

props/saber_fett263_buttons.h
```diff
@@ -257,7 +330,106 @@ SaberFett263Buttons() : PropBase() {}
...
+       	   }
```
profezzorn, line 376: Indented too much (should line up with } in the else on 374)

props/saber_fett263_buttons.h
```diff
@@ -257,7 +330,106 @@ SaberFett263Buttons() : PropBase() {}
...
+       // Auto Lockup Mode
+       case EVENTID(BUTTON_NONE, EVENT_CLASH, MODE_ON):
+         if (!battle_mode_) return false;
+         clash_impact_millis_ = millis();
+         swing_blast_ = false;
+         if (!swinging_) {
+           SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
+           swing_blast_ = false;
+           auto_lockup_on_ = true;
+             SaberBase::DoBeginLockup();
```
profezzorn, line 362: indented too much (two extra steps)

props/saber_fett263_buttons.h
```diff
@@ -257,7 +330,106 @@ SaberFett263Buttons() : PropBase() {}
...
+       // Battle Mode
+		    
+#ifdef FETT263_BATTLE_MODE
+#define FETT263_SWING_ON
+#define FETT263_TWIST_OFF
+       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON | BUTTON_AUX):
+         if (!battle_mode_) {
+           battle_mode_ = true;
+           // Force sound plays when entering Battle Mode
+             hybrid_font.SB_Force();
```
profezzorn, line 345: indented too much (two extra steps)

props/saber_fett263_buttons.h
fdarosa2663: I see, done.

props/saber_fett263_buttons.h
```diff
@@ -45,14 +85,48 @@ SaberFett263Buttons() : PropBase() {}
...
     if (!swinging_ && fusor.swing_speed() > 250) {
       swinging_ = true;
       Event(BUTTON_NONE, EVENT_SWING);
-    }
+    }	  
+    if (auto_lockup_on_ && 
+        !swinging_ && 
+        fusor.swing_speed() > 120 && 
+        millis() - clash_impact_millis_ > FETT263_LOCKUP_DELAY) {
+      if (SaberBase::Lockup()) {
```
profezzorn, line 110: I think we should just merge this with the surrounding if, so like:

if (auto_lockup_on &&
     !swinging_ &&
     fusor.swing_speed() > 120 &&
     millis() - clash_impact_millis_ > FETT264_LOCKUP_DELAY &&
     SaberBase::Lockup()) {
  SaberBase::DoEndLockup();
  SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
  auto_lockup_on_ = false;
}

and the same for the auto melt below...

props/saber_fett263_buttons.h
```diff
@@ -183,13 +257,13 @@ SaberFett263Buttons() : PropBase() {}
         swing_blast_ = true;
         hybrid_font.SB_Blast();
         return true;
-	
+
       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON):
-	if (swing_blast_) {
+        if (swing_blast_) {
 	  SaberBase::DoBlast();
```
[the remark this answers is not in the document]
profezzorn: this still needs to be done

props/saber_fett263_buttons.h
```diff
@@ -257,7 +330,104 @@ SaberFett263Buttons() : PropBase() {}
...
+         if (!battle_mode_) return false;
+         clash_impact_millis_ = millis();
+         swing_blast_ = false;
+         if (!swinging_) {
+           if (fusor.angle1() < - M_PI / 4) {
+             SaberBase::SetLockup(SaberBase::LOCKUP_DRAG);
+           } else {
+             SaberBase::SetLockup(SaberBase::LOCKUP_MELT);
+       	   }
+         swing_blast_ = false;
```
profezzorn, line 376: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -257,7 +330,104 @@ SaberFett263Buttons() : PropBase() {}
...
+         auto_melt_on_ = true;
```
profezzorn, line 377: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -257,7 +330,104 @@ SaberFett263Buttons() : PropBase() {}
...
+         SaberBase::DoBeginLockup();
```
profezzorn, line 378: indent two steps

props/saber_fett263_buttons.h
profezzorn: it's the millis() > 3000 that looks kind of odd.
I assume that it's to prevent this happening on boot?  I suggest we add a comment to explain that's what it's for.
Also, you need a space after the "if".

fdarosa2663:
OK, technically this delay was already in the existing prop file in OS4.7 because when the motion ship is first started it ignites the blade if you have EVENT_SWING, this was added to delay detecting Swing Speed for ignition to prevent false ignition on boot.  I've added note.

props/saber_fett263_buttons.h
```diff
@@ -45,14 +85,46 @@ SaberFett263Buttons() : PropBase() {}
...
+        !swinging_ && 
+        fusor.swing_speed() > 60 && 
+        millis() - clash_impact_millis_ > FETT263_LOCKUP_DELAY && 
+        SaberBase::Lockup()) {
+          SaberBase::DoEndLockup();
+          SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+          auto_melt_on_ = false;        
     }
-  }
+ }
```
profezzorn, line 127: not indented properly

props/saber_fett263_buttons.h
profezzorn: (line 370 now)

props/saber_fett263_buttons.h
```diff
@@ -257,7 +328,103 @@ SaberFett263Buttons() : PropBase() {}
...
+       	   }
+           auto_melt_on_ = true;
+           SaberBase::DoBeginLockup();
+         }
+         return true;
+#endif
+		    
+#ifdef FETT263_SWING_ON
+       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_OFF):  
+       // Due to motion chip startup on boot creating false ignition we delay Swing On at boot for 3000ms
```
profezzorn, line 381: indent two steps

fdarosa2663: indented

props/saber_fett263_buttons.h
```diff
@@ -45,12 +85,44 @@ SaberFett263Buttons() : PropBase() {}
...
+        if (!swinging_ && fusor.swing_speed() > 250) {
+          swinging_ = true;
+          Event(BUTTON_NONE, EVENT_SWING);
+        }
+        if (swinging_ && fusor.swing_speed() < 100) {
+          swinging_ = false;
+        }
+      }
+    } 
     if (!swinging_ && fusor.swing_speed() > 250) {
```
profezzorn, line 102: For this to work properly (meaning that it is only applied when the saber is OFF) it has to be in an else section that matches up with the if on line 89.

Otherwise, this part runs both when the saber is on and when it's off, which makes it impossible to set the swing speed to something higher when the saber is on.

props/saber_fett263_buttons.h
```diff
@@ -45,12 +85,44 @@ SaberFett263Buttons() : PropBase() {}
   bool swinging_ = false;
   void Loop() override {
     PropBase::Loop();
+    // Swing On gesture control this portion allows fine tuning of speed needed to ignite
+    if(!SaberBase::IsOn()) {
```
profezzorn, line 89: missing space after "if"

props/saber_fett263_buttons.h
```diff
@@ -45,12 +85,44 @@ SaberFett263Buttons() : PropBase() {}
...
       swinging_ = true;
       Event(BUTTON_NONE, EVENT_SWING);
-    }
+    }	  
+    if (auto_lockup_on_ && 
+        !swinging_ && 
+        fusor.swing_speed() > 120 && 
+        millis() - clash_impact_millis_ > FETT263_LOCKUP_DELAY &&
+        SaberBase::Lockup()) {
+          SaberBase::DoEndLockup();
```
profezzorn, line 111: too much indentation, should be two more than the if on line 106 (same for the next two lines)

props/saber_fett263_buttons.h
```diff
@@ -45,12 +85,44 @@ SaberFett263Buttons() : PropBase() {}
...
+    }	  
     if (swinging_ && fusor.swing_speed() < 100) {
       swinging_ = false;
+    }    
+    if (auto_melt_on_ &&
+        !swinging_ && 
+        fusor.swing_speed() > 60 && 
+        millis() - clash_impact_millis_ > FETT263_LOCKUP_DELAY && 
+        SaberBase::Lockup()) {
+          SaberBase::DoEndLockup();
```
profezzorn, line 123: too much indentation, should be two more spaces than the if on line 118, same for the next two lines

props/saber_fett263_buttons.h
```diff
@@ -45,14 +78,56 @@ SaberFett263Buttons() : PropBase() {}
   bool swinging_ = false;
   void Loop() override {
     PropBase::Loop();
+    // Swing On gesture control this portion allows fine tuning of speed needed to ignite
+    if(!SaberBase::IsOn() ) {
+      if (millis() - saber_off_time_ < MOTION_TIMEOUT) {
+        SaberBase::RequestMotion();
+      // Edit '250' value in line below to change swing on sensitivity, 250 ~ 400 work best in testing
```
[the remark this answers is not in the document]
profezzorn: still needs two more spaces of indentation...

profezzorn:
I'm trying to do this review as fast as I can, but it would really be a lot faster if you would just review your indentation and try to fix it proactively instead of waiting for me to add a comment for every line where the indentation is not right.

props/saber_fett263_buttons.h
fdarosa2663: that explains why some users couldn't adjust, did not notice that, fixed

fdarosa2663:
Yeah, sorry for the indentation troubles, I was getting confused at first which lines your comments refer to, I thought the entire chunks of green were to be edited which was just confusing me, I only just realized the comments refer to the final line.  And in some cases when trying to edit if I use backspace it jumps the lines to the beginning and then I have to try to line back up with space bar, Tab button is not moving two spaces so it's useless.  I "think" I have it all lined up now, fingers crossed.

props/saber_fett263_buttons.h
```diff
@@ -183,13 +255,13 @@ SaberFett263Buttons() : PropBase() {}
         swing_blast_ = true;
         hybrid_font.SB_Blast();
         return true;
-	
+
       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON):
-	if (swing_blast_) {
+        if (swing_blast_) {
 	  SaberBase::DoBlast();
```
profezzorn, line 261: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -183,13 +255,13 @@ SaberFett263Buttons() : PropBase() {}
...
 	}
-	return true;
-	
+        return true;
```
profezzorn, line 263: indented two steps too much

props/saber_fett263_buttons.h
profezzorn: still needs to be indented two steps

profezzorn:
The green areas just shows what has changed compared to the previous version of the file.
The comments always apply to the line directly above them.

props/saber_fett263_buttons.h
profezzorn: still needs to be indented two steps

props/saber_fett263_buttons.h
profezzorn: still needs to be fixed...

props/saber_fett263_buttons.h
```diff
@@ -257,7 +329,113 @@ SaberFett263Buttons() : PropBase() {}
...
+#define FETT263_SWING_ON
+#define FETT263_TWIST_OFF
+       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON | BUTTON_AUX):
+         if (!battle_mode_) {
+           battle_mode_ = true;
+           // Force sound plays when entering Battle Mode
+           hybrid_font.SB_Force();
+         } else {
+           battle_mode_ = false;
+	   // Exit Color Change sound plays when exiting Battle Mode
```
profezzorn, line 347: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -257,7 +329,113 @@ SaberFett263Buttons() : PropBase() {}
...
+       // Auto Lockup Mode
+       case EVENTID(BUTTON_NONE, EVENT_CLASH, MODE_ON):
+         if (!battle_mode_) return false;
+         clash_impact_millis_ = millis();
+         swing_blast_ = false;
+         if (!swinging_) {
+           SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
+           auto_lockup_on_ = true;
+           SaberBase::DoBeginLockup();
+	 }
```
profezzorn, line 361: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -257,7 +329,113 @@ SaberFett263Buttons() : PropBase() {}
...
+       case EVENTID(BUTTON_NONE, EVENT_STAB, MODE_ON):
+         if (!battle_mode_) return false;
+         clash_impact_millis_ = millis();
+         swing_blast_ = false;
+         if (!swinging_) {
+           if (fusor.angle1() < - M_PI / 4) {
+             SaberBase::SetLockup(SaberBase::LOCKUP_DRAG);
+           } else {
+             SaberBase::SetLockup(SaberBase::LOCKUP_MELT);
+       	   }
```
profezzorn, line 373: too much indentation, should line up with the } two lines up

props/saber_fett263_buttons.h
```diff
@@ -257,7 +329,113 @@ SaberFett263Buttons() : PropBase() {}
...
+           last_twist_ = millis();
+           battle_mode_ = true;
+         }
+         return true;
+#endif
+
+#ifdef FETT263_MULTI_PHASE
+       case EVENTID(BUTTON_NONE, EVENT_TWIST, MODE_ON | BUTTON_AUX):
+         // Delay twist events to prevent false trigger from over twisting
+	 if (millis() - last_twist_ > 2000) {
```
profezzorn, line 424: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -257,7 +329,113 @@ SaberFett263Buttons() : PropBase() {}
...
+           last_twist_ = millis();
+           next_preset();
+	 }
```
profezzorn, line 427: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -257,7 +329,113 @@ SaberFett263Buttons() : PropBase() {}
...
+           return true;
+
+       case EVENTID(BUTTON_NONE, EVENT_TWIST, MODE_ON | BUTTON_POWER):
+         // Delay twist events to prevent false trigger from over twisting
+	 if (millis() - last_twist_ > 2000) {
```
profezzorn, line 432: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -257,7 +329,113 @@ SaberFett263Buttons() : PropBase() {}
...
+           last_twist_ = millis();
+           previous_preset();
+	 }
```
profezzorn, line 435: indent two steps

props/saber_fett263_buttons.h
```diff
@@ -45,15 +85,48 @@ SaberFett263Buttons() : PropBase() {}
...
   void Loop() override {
     PropBase::Loop();
-    if (!swinging_ && fusor.swing_speed() > 250) {
-      swinging_ = true;
-      Event(BUTTON_NONE, EVENT_SWING);
-    }
-    if (swinging_ && fusor.swing_speed() < 100) {
-      swinging_ = false;
+    // Swing On gesture control this portion allows fine tuning of speed needed to ignite
+    if (!SaberBase::IsOn()) {
```
profezzorn, line 89: In general, it's easier to read code if conditions are not negated.
So, I would recommend changing this to if (!SaberBase::IsOn()) { and then switch the two code blocks below around.

profezzorn (commented):
Still a fair number of indentation things to fix.

props/saber_fett263_buttons.h
fdarosa2663: Is it possible it looks different for me? Because I'm lost- in my view "if" is indented two steps from "case" and "SaberBase" is indented two steps from "if" but this the comment view is showing them lined up so I'm totally lost on these indentations.   I don't see anything wrong with line 261 when I'm in Edit view.

props/saber_fett263_buttons.h
fdarosa2663: Sorry, I don't understand what am I changing?

profezzorn: I would recommend changing it from:

  if (!SaberBase::IsOn()) {
      CODE BLOCK A
  } else {
      CODE BLOCK B
  }

to:
  if (SaberBase::IsOn()) {
     CODE BLOCK B
  } else {
     CODE BLOCK A
  }

That makes it easier to follow the code because you don't need to think "else not on".

props/saber_fett263_buttons.h
profezzorn: What editor are you using?
Do you have tabs in the code?
Tabs can look different in different editors.
Basically, tabs are evil.

fdarosa2663: I've just been editing on github.com website and it looks all fine on my end, I use space bar because the tab jumps multiple spots.  But this one in particular when I view it on github.com in edit mode it's lined up fine but the comment isn't so I'm just lost

props/saber_fett263_buttons.h
fdarosa2663: ok, I see I have changed.  The indents are still different I look through everything in github.com and it's lined up as you explained so not sure what's going on there

props/saber_fett263_buttons.h
profezzorn: I think what we're going to do is to just fix everything *except* the indentation.
I'll fix that afterwards.
However, for next time, we're going to have to figure out a way where you can see and edit the indentation correctly.

profezzorn:
I merged it and then fixed the indentation myself.
I think tabs may have been the problem, but I'm not sure.
Maybe the github diff is doing something wrong...

PR #95, opened 2020-09-14
teensy 4.0 pindefs correction

artandtechvof:
some pin number were double defined, 
they are not actually by ProffieOS used but are a reminder not to use them in your hardware configuration
the audio shield lib will overwrite the pindefs

### DIFF
--- a/config/teensy_audio_board_rev_d.h
+++ b/config/teensy_audio_board_rev_d.h
@@ -8,13 +8,13 @@
 enum SaberPins {
   // Bottom edge (in pin-out diagram)
   // sdCardSelectPin is actually used, and can be altered if desired. 
-  // needs an external 10K pullup to 3.3V, provided on aduio shield
+  // needs an external 10K pullup to 3.3V, provided on audio shield
   sdCardSelectPin = 10,           // (See File->Example->SD->ReadWrite for
                                   // other possible values.)
 
   // these values are just a reminder not to use these pins for anything else, 
   // they definitions are not actually used, and overwritten by the AUDIO_CONTROL_SGTL5000
-  spiDataOut = 11,                 // spi out, serial flash, spi led & sd card
+  spiDataOut = 11,                // spi out, serial flash, spi led & sd card
   spiDataIn = 12,                 // spi in, serial flash & sd card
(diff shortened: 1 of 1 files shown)


config/teensy_audio_board_rev_d.h
```diff
@@ -7,20 +7,34 @@
 
 enum SaberPins {
   // Bottom edge (in pin-out diagram)
+  // sdCardSelectPin is actually used, and can be altered if desired. 
+  // needs an external 10K pullup to 3.3V, provided on aduio shield
```
profezzorn, line 11: aduio -> audio

config/teensy_audio_board_rev_d.h
```diff
@@ -7,20 +7,34 @@
...
-  spiClock = 10,                  // spi clock, flash, spi led & sd card
+  spiClock = 13,                  // spi clock, flash, spi led & sd card
+
+  // these values are just a reminder not to use these pins for anything else, 
+  // they definitions are not actually used, and overwritten by the AUDIO_CONTROL_SGTL5000
+  i2cDataPin = 18,                // I2C bus, SGTL5000 control
+  i2cClockPin = 19,               // I2C bus, SGTL5000 control
 
-//  auxPin = 15,                    // AUX button
+//  auxPin = 15,                    // AUX button  ->> this is also Volume pin on audioshield, do not use
```
profezzorn, line 26: the volume pin doesn't actually do anything, it's fine to use it.

artandtechvof: It will if you have the optional potentiometer mounted on the board, and it has a 0.1uF coupled do AGND permanently.
The stock shield, pin 15 can be re-used for other purposes

artandtechvof: Maybe make a note of that? When pot is installed do not use pin 15?

profezzorn: Yes.
I would probably state it the other way, especially since ProffieOS doesn't use the pot by default.

artandtechvof:
Spelcheck, and comment changes as suggested.

PR #97, opened 2020-09-19
add more animation types to ssd1306.h

SA-22C-smoothswing:
Adds animation playback support for the following states: saber is on, blast, clash and lockup

### DIFF
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -236,12 +236,7 @@ public:
     }
 	}
 	frame_count_++;
-  if (!SaberBase::IsOn()) {
-  	if (looped_frames_ > 1 && millis() - loop_start_ > font_config.ProffieOSImageDuration) {
-  	  STDOUT << "END OF LOOP\n";
-  	  screen_ = SCREEN_PLI;
-  	}
-  } else {
+  if (SaberBase::IsOn()) {
     if (looped_frames_ == 1 && frame_count_ == 1) {
       return font_config.ProffieOSImageDuration;
     } else if (looped_frames_ == 1 && frame_count_ > 1) {
@@ -250,7 +245,7 @@ public:
           ShowFile(&IMG_on);
           imgon_ = true;
         } else {
-      	  screen_ = SCREEN_PLI;
+          screen_ = SCREEN_PLI;
         }
       } else {
         ShowFile(&IMG_lock);
@@ -259,6 +254,11 @@ public:
       if (!SaberBase::Lockup()) {
         ShowFile(&IMG_on);
       }
+    }
+  } else {
+    if (looped_frames_ > 1 && millis() - loop_start_ > font_config.ProffieOSImageDuration) {
+      STDOUT << "END OF LOOP\n";
+      screen_ = SCREEN_PLI;
     }
   }
 
@@ -279,7 +279,7 @@ public:
   }
 
   void ShowFile(Effect* effect) {
-    if (effect) {
+    if (*effect) {
       MountSDCard();
       eof_ = true;
       file_.Play(effect);
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -244,13 +244,13 @@ public:
         ShowFile(&IMG_lock);
         return 3600000;
       } else {
-        ShowFile(&IMG_on);
+        if (!static_on_) ShowFile(&IMG_on);
       }
-    // looped image
     } else {
+      // looped image
       if (looped_frames_ == frame_count_) {
         if (!SaberBase::Lockup()) {
-          ShowFile(&IMG_on);
+          if (!static_on_) ShowFile(&IMG_on);
         }
       }
     }
@@ -286,8 +286,8 @@ public:
       loop_start_ = millis();
       frame_count_ = 0;
       SetScreenNow(SCREEN_IMAGE);
-      static_on_ = effect == &IMG_on && looped_frames_ == 1;
       eof_ = false;
+      on_played_ = effect == &IMG_on;
     }
   }
 
@@ -304,6 +304,7 @@ public:
 
   void SB_NewFont() override {
     ShowFile(&IMG_font);
+    on_played_ = false;
   }
 
   void SB_On() override {
@@ -323,7 +324,7 @@ public:
   }
 
   void SB_EndLockup() override {
-    ShowFile(&IMG_on);
+    if (!static_on_) ShowFile(&IMG_on);
   }
 
   void SB_Message(const char* text) override {
@@ -552,6 +553,8 @@ public:
       } else {
 	looped_frames_ = height / 128;
       }
+      static_on_ = on_played_ && looped_frames_ == 1;
+      STDOUT.println(static_on_);
     }
     // STDOUT << "ypos=" << ypos_ << " avail=" << f->Available() << "\n";
     if (f->Available() < sizeof(frame_buffer_)) return false;
@@ -614,6 +617,7 @@ private:
   uint8_t xor_ = 0;
   bool invert_y_ = 0;
   bool static_on_ = false;
+  bool on_played_ = false;
   uint32_t frame_buffer_[WIDTH];
   LoopCounter loop_counter_;
   char message_[32];
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -246,9 +246,18 @@ public:
       return font_config.ProffieOSImageDuration;
     } else if (looped_frames_ == 1 && frame_count_ > 1) {
       if (!SaberBase::Lockup()) {
-        ShowFile(&IMG_on);
+        if (!imgon_) {
+          ShowFile(&IMG_on);
+          imgon_ = true;
+        } else {
+      	  screen_ = SCREEN_PLI;
+        }
       } else {
         ShowFile(&IMG_lock);
+      }
+    } else if (looped_frames_ == frame_count_  && looped_frames_ > 1) {
+      if (!SaberBase::Lockup()) {
+        ShowFile(&IMG_on);
       }
     }
   }
@@ -299,14 +308,17 @@ public:
 
   void SB_On() override {
     ShowFile(&IMG_on);
+    imgon_ = true;
   }
 
   void SB_Blast() override {
     ShowFile(&IMG_blst);
+    imgon_ = false;
   }
 
   void SB_Clash() override {
     ShowFile(&IMG_clsh);
+    imgon_ = false;
   }
 
   void SB_BeginLockup() override {
@@ -315,6 +327,7 @@ public:
 
   void SB_EndLockup() override {
     ShowFile(&IMG_on);
+    imgon_ = true;
   }
 
   void SB_Message(const char* text) override {
@@ -604,6 +617,7 @@ private:
   uint16_t i;
   uint8_t xor_ = 0;
   bool invert_y_ = 0;
+  bool imgon_ = false;
   uint32_t frame_buffer_[WIDTH];
   LoopCounter loop_counter_;
   char message_[32];
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -229,10 +229,10 @@ public:
 	}
 	if (eof_) {
 	  // STDOUT << "EOF " << frame_count_ << "\n";
-    if (!SaberBase::IsOn()) {
-	   screen_ = SCREEN_PLI;
-	    if (frame_count_ == 1) return font_config.ProffieOSImageDuration;
-	    return FillFrameBuffer();
+    if (!SaberBase::IsOn() || static_on_) {
+      screen_ = SCREEN_PLI;
+      if (frame_count_ == 1) return font_config.ProffieOSImageDuration;
+      return FillFrameBuffer();
     }
 	}
 	frame_count_++;
@@ -240,20 +240,18 @@ public:
     // Single frame image
     if (looped_frames_ == 1) {
       if (frame_count_ == 1) return font_config.ProffieOSImageDuration;
-      if (frame_count_ > 1) {
-        if (SaberBase::Lockup()) {
-          ShowFile(&IMG_lock);
-        } else {
-          screen_ = SCREEN_PLI;
-          if (!imgon_) {
-            ShowFile(&IMG_on);
-          }
-        }
-      }
-      // looped image
-    } else if (looped_frames_ == frame_count_) {
-      if (!SaberBase::Lockup()) {
+      if (SaberBase::Lockup()) {
+        ShowFile(&IMG_lock);
+        return 3600000;
+      } else {
         ShowFile(&IMG_on);
+      }
+    // looped image
+    } else {
+      if (looped_frames_ == frame_count_) {
+        if (!SaberBase::Lockup()) {
+          ShowFile(&IMG_on);
+        }
       }
     }
   } else {
@@ -284,11 +282,11 @@ public:
       MountSDCard();
       eof_ = true;
       file_.Play(effect);
-      imgon_ = effect == &IMG_on;
       frame_available_ = false;
       loop_start_ = millis();
       frame_count_ = 0;
       SetScreenNow(SCREEN_IMAGE);
+      static_on_ = effect == &IMG_on && looped_frames_ == 1;
       eof_ = false;
     }
   }
@@ -615,7 +613,7 @@ private:
   uint16_t i;
   uint8_t xor_ = 0;
   bool invert_y_ = 0;
-  bool imgon_ = false;
+  bool static_on_ = false;
   uint32_t frame_buffer_[WIDTH];
   LoopCounter loop_counter_;
   char message_[32];
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -4,6 +4,10 @@
 // Images/animations
 IMAGE_FILESET(font);
 IMAGE_FILESET(boot);
+IMAGE_FILESET(on);
+IMAGE_FILESET(clsh);
+IMAGE_FILESET(blst);
+IMAGE_FILESET(lock);
 
 struct Glyph {
   int32_t skip : 7;
@@ -183,7 +187,7 @@ public:
 	xor_ = 0;
 	invert_y_ = false;
 	return 1000;
-	
+
       case SCREEN_STARTUP:
 	memset(frame_buffer_, 0, sizeof(frame_buffer_));
         // DrawText("==SabeR===", 0,15, Starjedi10pt7bGlyphs);
@@ -231,8 +235,16 @@ public:
 	}
 	frame_count_++;
 	if (looped_frames_ > 1 && millis() - loop_start_ > 5000) {
-	  STDOUT << "END OF LOOP\n";
-	  screen_ = SCREEN_PLI;
+    if (!SaberBase::IsOn()) {
+      STDOUT << "END OF LOOP\n";
+      screen_ = SCREEN_PLI;
+    } else {
+      if (!SaberBase::Lockup()) {
+        ShowFile(&IMG_on);
+      } else {
+        ShowFile(&IMG_lock);
+      }
+    }
 	}
 
 	if (font_config.ProffieOSAnimationFrameRate > 0.0) {
@@ -280,6 +292,36 @@ public:
     }
   }
 
+  void SB_On() override {
+    if (IMG_on) {
+      ShowFile(&IMG_on);
+    }
+  }
+
+  void SB_Blast() override {
+    if (IMG_blst) {
+      ShowFile(&IMG_blst);
+    }
+  }
+
+  void SB_Clash() override {
+    if (IMG_clsh) {
+      ShowFile(&IMG_clsh);
+    }
+  }
+
+  void SB_BeginLockup() override {
+    if (IMG_lock) {
+      ShowFile(&IMG_lock);
+    }
+  }
+
+  void SB_EndLockup() override {
+    if (IMG_on) {
+      ShowFile(&IMG_on);
+    }
+  }
+
   void SB_Message(const char* text) override {
     strncpy(message_, text, sizeof(message_));
     message_[sizeof(message_)-1] = 0;
@@ -309,10 +351,10 @@ public:
     Send(DISPLAYOFF);                    // 0xAE
     Send(SETDISPLAYCLOCKDIV);            // 0xD5
     Send(0x80);                                  // the suggested ratio 0x80
-    
+
     Send(SETMULTIPLEX);                  // 0xA8
     Send(HEIGHT - 1);
-    
+
     Send(SETDISPLAYOFFSET);              // 0xD3
     Send(0x0);                                   // no offset
     Send(SETSTARTLINE | 0x0);            // line #0
@@ -334,7 +376,7 @@ public:
     Send(0x40);
     Send(DISPLAYALLON_RESUME);           // 0xA4
     Send(NORMALDISPLAY);                 // 0xA6
-    
+
     Send(DEACTIVATE_SCROLL);
 
     Send(DISPLAYON);                     //--turn on oled panel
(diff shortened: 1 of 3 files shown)


display/ssd1306.h
```diff
@@ -231,8 +235,16 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
 	frame_count_++;
 	if (looped_frames_ > 1 && millis() - loop_start_ > 5000) {
-	  STDOUT << "END OF LOOP\n";
-	  screen_ = SCREEN_PLI;
+    if (!SaberBase::IsOn()) {
+      STDOUT << "END OF LOOP\n";
+      screen_ = SCREEN_PLI;
+    } else {
+      if (!SaberBase::Lockup()) {
+        ShowFile(&IMG_on);
```
profezzorn, line 243: I think we need to have some different behavior if "on" is looped or if it's not looped.
If it's not looped, we should just play it once I think.
If it is looped, we should change the 5000 above so that it keeps looping forever.

Right now, I think what happens is that it will play "on" over and over again, regardless of whether it is looped or not. And if it is looped, it will stop the loop after 5 seconds, only to re-start it again.

display/ssd1306.h
```diff
@@ -280,6 +292,36 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
     }
   }
 
+  void SB_On() override {
+    if (IMG_on) {
```
profezzorn, line 296: might be easier if we put this test inside ShowFile

SA-22C-smoothswing:
So it appears that the framerate of an animated bmp is variable and based on its total frames?

from line # 241:
	if (looped_frames_ > 1) {
	  return 1000 / looped_frames_;

Is this intentional?  Shouldn't we be setting a static framerate?

profezzorn:
The frame rate can be set in the config file if I remember correctly.
For the fallback I tried to do what CFX is doing, although it is not exactly documented, so I don't know if I got it right or not.

profezzorn:
(the font config file that is, not the proffieos config file..)

SA-22C-smoothswing:
Ok that's helpful.  What I'm seeing is that the frames per second on the display slows down when the saber is on compared to when it's off.  Is this because of SD card contention: playing a wav file AND a bmp at the same time?

The reason why I need to know is that I'm trying to time the end of bmp display to the length of the file but I'm getting inconsistent results and it appears to be tied to the display just being slower to refresh/redraw when the saber is on vs off.

profezzorn:
That is most likely why.
The display code is written to be as gentle as possible, so it transmits a small number of bytes to the display, then yields to let the motion chip use the i2c bus for a while. This means that it can take a while to transmit a frame sometimes, because it can take a while before control returns to the ssd1306 class after yielding.  It's possible that using HFLINK can fix or reduce this problem, not sure if it's that simple though.

display/ssd1306.h
SA-22C-smoothswing: done.

display/ssd1306.h
SA-22C-smoothswing: Ok, I've got that all sorted now.

Looped animations for for 'on' will play continuously, single frames will play for 5 seconds.  Same for lockup (as long as the saber is in lockup, that is.)

clash and blast animations will play either for the duration of a single loop OR for 5 seconds in the case of a static image.

SA-22C-smoothswing:
Ok, I've got this working the way I think is correct.

Behaviour:
When the prop is off, all bmp files regardless of looping or no will play for ProffieOSImageDuration (defaults to 5000) and I've added that parameter to the font config file in hybrid_font.h

When prop is on the following files will play:
on.bmp - if the bmp is looping, on.bmp will always play.  If it's static, it will display for ProffieOSImageDuration and then revert to PLI
blst.bmp / clsh.bmp - if the bmp is looping, it will play for ONE loop then revert to on.bmp.  If static, it will play for ProffieOSImageDuration and then revert to on.bmp
lock.bmp - will play on a loop while lockup is active, then reverting to on.bmp.

If you think it's appropriate we could also play the looping blst/clsh for the ProffieOSImageDuration rather than just one loop but since those events are so quick, I thought a single loop made more sense.

display/ssd1306.h
```diff
@@ -274,10 +303,31 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
-    if (IMG_font) {
-      // Overrides message from below..
-      ShowFile(&IMG_font);
-    }
+    ShowFile(&IMG_font);
+  }
+
+  void SB_On() override {
+    ShowFile(&IMG_on);
+    imgon_ = true;
```
profezzorn, line 311: If we really just need to know if it is IMG_on or not, we can just check effect == &IMG_on in ShowFile instead.
If there are potentially other effects that needs this to be true, then it should probably be an argument to ShowFile()

display/ssd1306.h
```diff
@@ -230,9 +234,19 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
 	  return FillFrameBuffer();
 	}
 	frame_count_++;
-	if (looped_frames_ > 1 && millis() - loop_start_ > 5000) {
-	  STDOUT << "END OF LOOP\n";
-	  screen_ = SCREEN_PLI;
+  if (!SaberBase::IsOn()) {
+	  if (looped_frames_ > 1 && millis() - loop_start_ > 5000) {
```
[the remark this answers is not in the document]
SA-22C-smoothswing: yeah, no matter what editor I use, git seems to mess up the tabs.

display/ssd1306.h
```diff
@@ -230,9 +234,19 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
 	  return FillFrameBuffer();
 	}
 	frame_count_++;
-	if (looped_frames_ > 1 && millis() - loop_start_ > 5000) {
-	  STDOUT << "END OF LOOP\n";
-	  screen_ = SCREEN_PLI;
+  if (!SaberBase::IsOn()) {
```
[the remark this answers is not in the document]
SA-22C-smoothswing: refactored.

display/ssd1306.h
SA-22C-smoothswing: fixed.  I don't see the need for other effects, so just setting the bool in showfile

display/ssd1306.h
```diff
@@ -252,14 +266,16 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
   void ShowFile(Effect* effect) {
-    MountSDCard();
-    eof_ = true;
-    file_.Play(effect);
-    frame_available_ = false;
-    loop_start_ = millis();
-    frame_count_ = 0;
-    SetScreenNow(SCREEN_IMAGE);
-    eof_ = false;
+    if (effect) {
```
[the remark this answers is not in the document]
SA-22C-smoothswing: fixed.

SA-22C-smoothswing:
ok, all issues except perhaps indentation should be resolved.

display/ssd1306.h
```diff
@@ -252,14 +278,21 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
-    frame_available_ = false;
-    loop_start_ = millis();
-    frame_count_ = 0;
-    SetScreenNow(SCREEN_IMAGE);
-    eof_ = false;
+    if (*effect) {
+      MountSDCard();
+      eof_ = true;
+      file_.Play(effect);
+      if (effect == &IMG_on) {
```
profezzorn, line 285: this if statement can be written as: imgon_ = effect == &IMG_on;

display/ssd1306.h
```diff
@@ -225,15 +229,37 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
 	}
 	if (eof_) {
 	  // STDOUT << "EOF " << frame_count_ << "\n";
-	  screen_ = SCREEN_PLI;
-	  if (frame_count_ == 1) return 5000;
-	  return FillFrameBuffer();
+    if (!SaberBase::IsOn()) {
```
profezzorn, line 232: definitely something wrong with indentation still.
If there are tabs in here, please remove them.
If that is not the problem, I'm going to have to fix the indentation after I merge.
(Assuming of course that it's not the diff tool that is just showing it weird...)

sound/hybrid_font.h
```diff
@@ -201,9 +204,9 @@ class HybridFont : public SaberBase {
     if (delta_micros > 1000000) delta_micros = 1;
     if (swing_speed > swingThreshold) {
       float delta = delta_micros * 0.000001;
-#ifdef ENABLE_SPINS      
+#ifdef ENABLE_SPINS
```
profezzorn, line 207: Why are the spin changes in here? Need to sync maybe?

display/ssd1306.h
```diff
@@ -225,15 +229,37 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
 	frame_count_++;
-	if (looped_frames_ > 1 && millis() - loop_start_ > 5000) {
-	  STDOUT << "END OF LOOP\n";
-	  screen_ = SCREEN_PLI;
-	}
+  if (SaberBase::IsOn()) {
+    if (looped_frames_ == 1 && frame_count_ == 1) {
+      return font_config.ProffieOSImageDuration;
+    } else if (looped_frames_ == 1 && frame_count_ > 1) {
+      if (!SaberBase::Lockup()) {
```
profezzorn, line 243: I think we should invert this if statement. (and swap the two code blocks below)
Same for the one on the next line.

display/ssd1306.h
```diff
@@ -225,15 +229,37 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
+	    return FillFrameBuffer();
+    }
 	}
 	frame_count_++;
-	if (looped_frames_ > 1 && millis() - loop_start_ > 5000) {
-	  STDOUT << "END OF LOOP\n";
-	  screen_ = SCREEN_PLI;
-	}
+  if (SaberBase::IsOn()) {
+    if (looped_frames_ == 1 && frame_count_ == 1) {
```
profezzorn, line 240: I think I would like to change the structure a bit, and add some comments, something like:

    if (looped_frames == 1) {
        // Single frame
        if (frame_count == 1) return font_config.ProffieOSImageDuration;
        if (SaberBase::Lockup()) {
        } else {
        }
     } else {
        // Looped
     } 

There shouldn't be a need to check looped_frames in so many places.

display/ssd1306.h
```diff
@@ -225,15 +229,37 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
+      return font_config.ProffieOSImageDuration;
+    } else if (looped_frames_ == 1 && frame_count_ > 1) {
+      if (!SaberBase::Lockup()) {
+        if (!imgon_) {
+          ShowFile(&IMG_on);
+        } else {
+          screen_ = SCREEN_PLI;
+        }
+      } else {
+        ShowFile(&IMG_lock);
```
profezzorn, line 250: What happens if there are no IMG_lock file(s)?
We should still play IMG_on in that case, right?
(And if there is no IMG_on, we should go back to PLI?)

profezzorn (commented):
getting better...

display/ssd1306.h
SA-22C-smoothswing: Any tabs were present when I pulled the file from github, my editor is configured to use spaces.

sound/hybrid_font.h
SA-22C-smoothswing: I think it's whitespace issues, rather than actual new code.

display/ssd1306.h
SA-22C-smoothswing: done

display/ssd1306.h
SA-22C-smoothswing: Should be a bit more simplified now.

display/ssd1306.h
SA-22C-smoothswing: Added a screen_ = SCREEN_PLI before invoking ShowFile, which covers off a missing on.bmp.  Missing lock.bmp is already handled by the check within ShowFile for the effect existing.

display/ssd1306.h
```diff
@@ -225,15 +229,39 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
 	frame_count_++;
-	if (looped_frames_ > 1 && millis() - loop_start_ > 5000) {
-	  STDOUT << "END OF LOOP\n";
-	  screen_ = SCREEN_PLI;
-	}
+  if (SaberBase::IsOn()) {
+    // Single frame image
+    if (looped_frames_ == 1) {
+      if (frame_count_ == 1) return font_config.ProffieOSImageDuration;
+      if (frame_count_ > 1) {
```
profezzorn, line 243: When we get here, frame_count_ must be greater than one, so we don't need to check again.

display/ssd1306.h
```diff
@@ -225,15 +229,39 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
+          ShowFile(&IMG_lock);
+        } else {
+          screen_ = SCREEN_PLI;
+          if (!imgon_) {
+            ShowFile(&IMG_on);
+          }
+        }
+      }
+      // looped image
+    } else if (looped_frames_ == frame_count_) {
```
profezzorn, line 254: While not incorrect, I think it would be better if this was:

} else {
   // looped image
   if (looped_frames_ == frame_count_) {
   }
}

The reason for this is that the else  and the if are testing different concepts (looped vs. non-looped) and (done vs not done)  so joining them together on the same level can make it confusing.

display/ssd1306.h
```diff
@@ -225,15 +229,39 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
+  if (SaberBase::IsOn()) {
+    // Single frame image
+    if (looped_frames_ == 1) {
+      if (frame_count_ == 1) return font_config.ProffieOSImageDuration;
+      if (frame_count_ > 1) {
+        if (SaberBase::Lockup()) {
+          ShowFile(&IMG_lock);
+        } else {
+          screen_ = SCREEN_PLI;
+          if (!imgon_) {
```
profezzorn, line 248: This, and all of the logic in here needs comments to make it obvious what is going on.

display/ssd1306.h
```diff
@@ -274,10 +305,27 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
+  void SB_Clash() override {
+    ShowFile(&IMG_clsh);
+  }
+
+  void SB_BeginLockup() override {
+    ShowFile(&IMG_lock);
+  }
+
+  void SB_EndLockup() override {
+    ShowFile(&IMG_on);
```
profezzorn, line 328: It seems to me that the "on" image is not looping, then we only want to play it once when turning on.

display/ssd1306.h
profezzorn: I'm actually a little confused, because I don't know how we get here if imgon_ is false.

display/ssd1306.h
```diff
@@ -225,15 +229,39 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
 	}
 	frame_count_++;
-	if (looped_frames_ > 1 && millis() - loop_start_ > 5000) {
-	  STDOUT << "END OF LOOP\n";
-	  screen_ = SCREEN_PLI;
-	}
+  if (SaberBase::IsOn()) {
+    // Single frame image
+    if (looped_frames_ == 1) {
+      if (frame_count_ == 1) return font_config.ProffieOSImageDuration;
```
profezzorn, line 242: maybe if lockup is true, we should just return a large number here, like an hour or something?
That way we never have to worry about re-starting a lockup image.

display/ssd1306.h
SA-22C-smoothswing: i changed the logic for detecting static on.bmp, it's much cleaner now.

display/ssd1306.h
SA-22C-smoothswing: I'm struggling to come up with a good way to do a check to see if a on.bmp is static when looking to play the on.bmp after an effect (blast, clash, lockup)  Any tips on how you'd do it?

profezzorn: I'm thinking we need another bool. (Maybe called on_is_looped_)
When we calculate if the current frame is looped or not, we then do:

   if (is_on_) { on_is_looped_ = looped_frames_ > 1; }

We would need to reset on_is_looped_ in SB_NewFont()

display/ssd1306.h
```diff
@@ -225,15 +229,37 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
+      if (SaberBase::Lockup()) {
+        ShowFile(&IMG_lock);
+        return 3600000;
+      } else {
+        ShowFile(&IMG_on);
+      }
+    // looped image
```
profezzorn, line 249: this comment should go after the } else {

display/ssd1306.h
```diff
@@ -225,15 +229,37 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
+    } else {
+      if (looped_frames_ == frame_count_) {
```
profezzorn, line 251: I'm not sure why we need this line (and the following two)

If we are at the end of an animation, and not in lockup, we start IMG_on.
But wouldn't we already be playing IMG_on? And if it is looped, it would restart automatically.
So why bother?

display/ssd1306.h
profezzorn: Of course, this makes the assumption that a looping ON image means that we need to resume it after lockup, and that a non-looping ON image means that we don't.  Not entirely sure if that's the assumption we should make though...

SA-22C-smoothswing:
My working assumption was that looping on.bmp would always play, except when interrupted by blst, clsh or lock, and then resume.  We'd only want the static on.bmp to play once and revert to PLI.

What do you think would be the right approach here?

display/ssd1306.h
SA-22C-smoothswing: this is how we start IMG_on.  Without these lines, an animated blst or clsh will play forever.

display/ssd1306.h
SA-22C-smoothswing: My working assumption was that looping on.bmp would always play, except when interrupted by blst, clsh or lock, and then resume. We'd only want the static on.bmp to play once and revert to PLI.

What do you think would be the right approach here?

display/ssd1306.h
profezzorn: That will make the loop play exactly once.
However, if that's what the user wants, they can just make a non-looped animation.
Maybe we should make it so that the play duration can be configured per effect?
That makes a lot of sense for single-frame images as well, you could specify an image to show for "clash", and specify that it should be shown for 200ms or something....

display/ssd1306.h
profezzorn: That's basically my assumption too. I just wanted to check my assumptions. :)

SA-22C-smoothswing:
I tried using a bool called static_on_ but the non-looped on.bmp still plays once after a lockup/clash or blast.

I'm clearly doing something wrong, but not sure what.

profezzorn:
static_on_ isn't a particularly good name, because animated and single frame should basically behave the same, and "animated" and "static" sounds like different things.

display/ssd1306.h
```diff
@@ -252,14 +278,17 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
...
+    if (*effect) {
+      MountSDCard();
+      eof_ = true;
+      file_.Play(effect);
+      frame_available_ = false;
+      loop_start_ = millis();
+      frame_count_ = 0;
+      SetScreenNow(SCREEN_IMAGE);
+      eof_ = false;
+      on_played_ = effect == &IMG_on;
```
profezzorn, line 290: on_played_ sounds like "on" has played at any time in the past.
In fact, I think it might be better to just have a pointer to the effect and call it current_effect_.

display/ssd1306.h
```diff
@@ -506,6 +553,8 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
       } else {
 	looped_frames_ = height / 128;
       }
+      static_on_ = on_played_ && looped_frames_ == 1;
```
profezzorn, line 556: The current code sets static_on_ to false if on_played_ is false, which I don't think is what you want.

Try something like:
    if (current_effect_ == &IMG_on) {
       looped_on_ = looped_frames_ > 1;
    }

display/ssd1306.h
SA-22C-smoothswing: Ok, for per-effect durations, would that go in config.ini?

So something like

ProffieOSBootImageDuration
ProffieOSFontImageDuration
ProffieOSClashImageDuration
ProffieOSBlastImageDuration
ProffieOSOnImageDuration

SA-22C-smoothswing: Or do we somehow link it to the duration in the blade style for each effect?

profezzorn: The blade style durations are going to be all wrong I think. Better to just add config variables for these things. (Unless we could find some way to put them in the file themselves, which would be even better.)

display/ssd1306.h
```diff
@@ -567,6 +619,8 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
   uint16_t i;
   uint8_t xor_ = 0;
   bool invert_y_ = 0;
+  bool looped_on_ = false;
+  Effect* current_effect_;
```
profezzorn, line 623: I think this needs to be volatile, because it's used in both interrupts and non-interrupts.

display/ssd1306.h
SA-22C-smoothswing: added duration variables for all effect types

display/ssd1306.h
SA-22C-smoothswing: bool looped_on_ marked as volatile

SA-22C-smoothswing:
Ok, added distinct display durations for effect types.  There is one bug that I can't figure out: the boot.bmp doesn't display for the right interval if it's static: it goes straight to PLI.

sound/hybrid_font.h
```diff
@@ -55,12 +60,22 @@ class FontConfigFile : public ConfigFile {
   float ProffieOSSlashAccelerationThreshold;
   // For OLED displays, this specifies the frame rate of animations.
   float ProffieOSAnimationFrameRate;
-#ifdef ENABLE_SPINS  
+  // for OLED displays, the time a static BMP or loop will play when saber is off
+  float ProffieOSFontImageDuration;
+  // for OLED displays, the time a font.bmp or boot.bmp will play
```
profezzorn, line 65: This comment seems like it belongs two lines up...

profezzorn:
> Ok, added distinct display durations for effect types. There is one bug that I can't figure out: the boot.bmp doesn't display for the right interval if it's static: it goes straight to PLI after one frame.

Was this working before?
I've looked very very carefully at your changes, and I haven't seen anything that I think would change how long boot.bmp is displayed.

SA-22C-smoothswing:
> > Ok, added distinct display durations for effect types. There is one bug that I can't figure out: the boot.bmp doesn't display for the right interval if it's static: it goes straight to PLI after one frame.
> 
> Was this working before?
> I've looked very very carefully at your changes, and I haven't seen anything that I think would change how long boot.bmp is displayed.

You know, I don't think it was working before either.  Still, from how the image code is written, I can't figure out why the boot image isn't behaving the same as the font image.

SA-22C-smoothswing:
All requested changes/corrections made.

profezzorn:
I think it's time to merge this and mop up any remaining issues afterwards.

PR #98, opened 2020-09-20
motion/mpu6050:    added ASYNC read option, for ProffieOS5.x on Teensy4.0 

artandtechvof:
As requested, made the modification of the mpu605o a sperate pull-request from the Teensy4.0 integration.

    added ASYNC read option, for ProffieOS5.x on Teensy4.0 
    cleaned up the original read-rouitne a little bit (indentation and non relevant comments deleted)
    modified:   motion/mpu6050.h

Switching between the 2 is now coupled to the #define USE_TEENSY4
Maybe this should be more generic, and use a separate define line #define mpu6050_ASYNC in the hardware config?

### DIFF
--- a/motion/mpu6050.h
+++ b/motion/mpu6050.h
@@ -100,7 +100,7 @@ public:
 
   MPU6050() : I2CDevice(0x68) {}
 
-#ifdef ÀSYNC_READ_MOTION // use Wire.h and ASYNC to read MPU6050
+#ifdef ASYNC_READ_MOTION // use Wire.h and ASYNC to read MPU6050
   void Loop() override {
     STATE_MACHINE_BEGIN();
 
--- a/motion/mpu6050.h
+++ b/motion/mpu6050.h
@@ -100,6 +100,109 @@ public:
 
   MPU6050() : I2CDevice(0x68) {}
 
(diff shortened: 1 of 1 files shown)


motion/mpu6050.h
```diff
@@ -100,6 +100,109 @@ class MPU6050 : public I2CDevice, Looper, StateMachine {
 
   MPU6050() : I2CDevice(0x68) {}
 
+#ifdef USE_TEENSY4 //use Wire.h and ASYNC to read MPU6050
```
profezzorn, line 103: I think we should replace this with something like "USE_ASYNC_MOTION" or something and then define that in the teensy4 config file instead.

Also, there should be a space after //

motion/mpu6050.h
```diff
@@ -100,6 +100,109 @@ class MPU6050 : public I2CDevice, Looper, StateMachine {
...
+  }
+
+  uint8_t databuffer[14];
+  int status_reg;
+  uint32_t last_temp_;
+  uint32_t last_event_;
+  bool first_motion_;
+  bool first_accel_;
+
+#else //non async, motion interrupt pin is not required for this method of reading mpu6050
```
profezzorn, line 204: space after //

artandtechvof:
Updates as requested

PR #102, opened 2020-10-02
SubBladeWithStride for KR Strips

lmetty:
Added new SubBladeWithStride wrapper with an additional 'stride' parameter to adjust the step between leds.
Primarily intended to be used with KR Pixel Strips, allowing the odd and even pixels to be mapped as independent sub blades.

Usage example:

BladeConfig blades[] = {
 { 0, SubBladeWithStride(0, 262, 2, WS281XBladePtr<264, bladePin, Color8::GRB, PowerPINS<bladePowerPin2, bladePowerPin3> >()),
    SubBladeWithStride(1, 263, 2, NULL)
  , CONFIGARRAY(presets) },
};

The new 3rd parameter is the 'stride' which is essentially the step between leds, a value of 1 would be the same as a regular continuous blade but a value of 2 would only set half of the leds by skipping every other led.

So, in the example above, the two sub blades map to leds 0-262 and 1-263 respectively, although these ranges overlap they specify a stride of 2 and are offset by 1 so they are actually interleaved with no clashing leds, one blade sets the even pixels (0,2,4,6,etc...) and the other sets the odd pixels (1,3,5,7,etc...).

### DIFF
--- a/blades/sub_blade.h
+++ b/blades/sub_blade.h
@@ -195,4 +195,48 @@ class BladeBase* SubBladeReverse(int first_led, int last_led, BladeBase* blade) 
   ret->SetupSubBlade(blade, last_led, last_led + 1 - first_led);
   return ret;
 }
+
+class SubBladeWrapperWithStride : public SubBladeWrapper {
+public:
+  void SetupStride(int stride) {
+    stride_ = stride;
+  }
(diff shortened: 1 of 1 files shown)


profezzorn:
Looks pretty good, nice job.

PR #107, opened 2020-10-14
Added gesture Length to Push and Thrust to avoid false detection

fdarosa2663:
Per your suggestion I've added Gesture Length (ms) to both Push and Thrust to avoid false detections.  I've also included separate defines for each to allow the user to modify as needed in the config.  We'll need the users who experienced the false detections to test as none of my sabers experienced the false detections.

### DIFF
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -416,13 +416,13 @@ SaberFett263Buttons() : PropBase() {}
           Off();
           saber_off_time_ = millis();
         }
-#endif
-#ifndef FETT263_BM_DISABLE_OFF_BUTTON
+#else
          Off();
          saber_off_time_ = millis();
+         battle_mode_ = false;
 #endif
-#ifndef FETT263_BATTLE_MODE_ALWAYS_ON
-          battle_mode_ = false;
+#ifdef FETT263_BATTLE_MODE_ALWAYS_ON
+          battle_mode_ = true;
 #endif
         }
         return true;
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -411,11 +411,15 @@ SaberFett263Buttons() : PropBase() {}
 #endif
         if (!swinging_) {
           swing_blast_ = false;
+#ifdef FETT263_BM_DISABLE_OFF_BUTTON
+        if (!battle_mode_) {
+          Off();
+          saber_off_time_ = millis();
+        }
+#endif
 #ifndef FETT263_BM_DISABLE_OFF_BUTTON
-          if (!battle_mode_) {
-            Off();
-            saber_off_time_ = millis();
-          }
+         Off();
+         saber_off_time_ = millis();
 #endif
 #ifndef FETT263_BATTLE_MODE_ALWAYS_ON
           battle_mode_ = false;
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -254,12 +254,12 @@ SaberFett263Buttons() : PropBase() {}
           mss.y * mss.y + mss.z * mss.z > 100 &&
           fusor.swing_speed() < 30 &&
           fabs(fusor.gyro().x) < 10) {
-        if (millis() - push_begin_millis_ > 25) {
+        if (millis() - push_begin_millis_ > 10) {
           Event(BUTTON_NONE, EVENT_PUSH);
           push_begin_millis_ = millis();
-        } else {
-          push_begin_millis_ = millis();
-        }
+        } 
+      } else {
+        push_begin_millis_ = millis();
       }
 
     } else {
@@ -278,9 +278,9 @@ SaberFett263Buttons() : PropBase() {}
         if (millis() - thrust_begin_millis_ > 50) {
           Event(BUTTON_NONE, EVENT_THRUST);
           thrust_begin_millis_ = millis();
-        } else {
-          thrust_begin_millis_ = millis();
-        }
+        } 
+      } else {
+        thrust_begin_millis_ = millis();
       }
     }
   }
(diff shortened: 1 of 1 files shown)


props/saber_fett263_buttons.h
```diff
@@ -110,10 +110,16 @@
 // FETT263_THRUST_ON_NO_BM
 // To enable Thrust On Ignition control but not activate Battle Mode (works with THRUST_ON_PREON only)
 //
+// FETT263_THRUST_LENGTH 50
```
profezzorn, line 113: I don't really think you need defines for everything.
You don't really need to change it, but too many defines are just going to confuse users and make it harder to keep things backwards compatible. People who really want to tweak things can just fiddle with the code directly.

props/saber_fett263_buttons.h
```diff
@@ -251,10 +265,18 @@ SaberFett263Buttons() : PropBase() {}
 
       // EVENT_PUSH
       if (fabs(mss.x) < 3.0 &&
-          mss.y * mss.y + mss.z * mss.z > 120 &&
+          mss.y * mss.y + mss.z * mss.z > 100 &&
           fusor.swing_speed() < 30 &&
           fabs(fusor.gyro().x) < 10) {
-        Event(BUTTON_NONE, EVENT_PUSH);
+          if (!push_begin_) {
```
profezzorn, line 271: I don't think this works that well.
Basically, when pushing is detected, it starts a timer, and when that timer expires, the next time pushing is detected, a push event is generated.

I think what you want is something like this:

    if ( DETECT PUSH ) {
       if (millis() - push_begin_millis_ > FETT263_PUSH_LENGTH) {
         Event(BUTTON_NONE, EVENT_PUSH);
         push_begin_millis_ = millis();
       }
     } else {
        push_begin_millis_ = millis();
     }

Basically, push_begin_millis_ is set to millis every time a push is NOT detected. That way, millis() - push_begin_millis_ will only become 25 if push was detected every time we get to this code for 25 milliseconds.

Same thing goes for thrust below.

props/saber_fett263_buttons.h
```diff
@@ -411,12 +441,18 @@ SaberFett263Buttons() : PropBase() {}
...
           saber_off_time_ = millis();
+        }
+#else
+         Off();
+         saber_off_time_ = millis();
+         battle_mode_ = false;
 #endif
-#ifndef FETT263_BATTLE_MODE_ALWAYS_ON
-          battle_mode_ = false;
+#ifdef FETT263_BATTLE_MODE_ALWAYS_ON
```
profezzorn, line 454: Did you mean to change this back again?
Which is it supposed to be?

fdarosa2663: this was tied to the previous change that was merged to handle the button changing to false.

props/saber_fett263_buttons.h
fdarosa2663: thanks, I was thinking the push_begin_millis_ would be continuously overwritten if I didn't include a bool, I've updated both the same, it's working for me but the target for this is the users with the speaker vibrations so will need their testing to confirm

props/saber_fett263_buttons.h
fdarosa2663: OK, wasn't sure if it made it easier or harder but I've removed.

props/saber_fett263_buttons.h
```diff
@@ -251,10 +251,15 @@ SaberFett263Buttons() : PropBase() {}
 
       // EVENT_PUSH
       if (fabs(mss.x) < 3.0 &&
-          mss.y * mss.y + mss.z * mss.z > 120 &&
+          mss.y * mss.y + mss.z * mss.z > 100 &&
           fusor.swing_speed() < 30 &&
           fabs(fusor.gyro().x) < 10) {
-        Event(BUTTON_NONE, EVENT_PUSH);
+        if (millis() - push_begin_millis_ > 25) {
```
profezzorn, line 257: This is not what I suggested, the else is in the wrong place.
As written, this will do roughly the opposite to what we want.
push_begin_millis_ will be set to millis() every time that the push detection triggers.
push events will be generated if it's been more than 25 ms since the last push detection, but it will not do anything to filter out very short events.

fdarosa2663: sorry, fixed. did have to lower for push as it's a very short movement as opposed to thrust

props/saber_fett263_buttons.h
```diff
@@ -251,10 +251,15 @@ SaberFett263Buttons() : PropBase() {}
 
       // EVENT_PUSH
       if (fabs(mss.x) < 3.0 &&
-          mss.y * mss.y + mss.z * mss.z > 120 &&
+          mss.y * mss.y + mss.z * mss.z > 100 &&
           fusor.swing_speed() < 30 &&
           fabs(fusor.gyro().x) < 10) {
-        Event(BUTTON_NONE, EVENT_PUSH);
+        if (millis() - push_begin_millis_ > 10) {
```
profezzorn, line 257: At 10 ms, this loop will only execute a few times, and it might not be worth doing this extra step.
Wish we could test for "speed" instead of "mss", that would make all of this much better.

profezzorn:
This is ready to merge, but git says it has conflicts that must be resolved first.

fdarosa2663:
How do I resolve conflicts, it's not giving me an option

props/saber_fett263_buttons.h
fdarosa2663: Did we try speed?  It's working for me currently but I can test another method, with the 25 it was inconsistent because the gesture is meant to be a short movement so I had to lower to have it work.  but if we can improve I'm open to trying other options.

fdarosa2663:
looks like indents were the conflicts, should be ok unless you want to test out some other changes to Push first, I'm open to suggestions

PR #112, opened 2020-10-17
timers added to Stab, Thrust, Twist ignitions

fdarosa2663:
Added timers to reduce false gesture ignition from in.wav during retraction, will need some users to test once it's merged.  Also, per TRA PM I set the timers to abs() to account for millis() reset.  If there's a better way I am open, but I "think" the reset on millis() for a few timers was the inconsistency that was hard to replicate.

### DIFF
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -416,13 +416,13 @@ SaberFett263Buttons() : PropBase() {}
           Off();
           saber_off_time_ = millis();
         }
-#endif
-#ifndef FETT263_BM_DISABLE_OFF_BUTTON
+#else
          Off();
          saber_off_time_ = millis();
+         battle_mode_ = false;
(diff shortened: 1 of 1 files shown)


props/saber_fett263_buttons.h
```diff
@@ -260,7 +260,7 @@ SaberFett263Buttons() : PropBase() {}
           mss.y * mss.y + mss.z * mss.z > 70 &&
           fusor.swing_speed() < 30 &&
           fabs(fusor.gyro().x) < 10) {
-        if (millis() - push_begin_millis_ > 5) {
+        if (abs(millis() - push_begin_millis_) > 5) {
```
profezzorn, line 263: no, remove all abs() and then I will review again.

fdarosa2663: I have removed

PR #113, opened 2020-10-21
Update saber_fett263_buttons.h

fdarosa2663:
Added define to enable Force Push to be active full time instead of just during Battle Mode per user request

### DIFF
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -206,6 +206,54 @@
 #error You cannot define both FETT263_FORCE_PUSH_ALWAYS_ON and FETT263_FORCE_PUSH
 #endif
 
+#ifdef FETT263_SWING_ON
+#define SWING_GESTURE
+#endif
+
+#ifdef FETT263_SWING_ON_PREON
+#define SWING_GESTURE
+#endif
+
+#if defined(FETT263_SWING_ON_NO_BM) && !defined(SWING_GESTURE)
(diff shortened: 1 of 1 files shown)


props/saber_fett263_buttons.h
```diff
@@ -734,6 +734,20 @@ SaberFett263Buttons() : PropBase() {}
         return true;
 
 #endif
+        
+#ifdef FETT263_FORCE_PUSH_ALWAYS_ON
```
profezzorn, line 738: This is so similar to the previous if statement.
I think you should write it as:

#ifdef FETT263_FORCE_PUSH_ALWAYS_ON
#define FORCE_PUSH_CONDITION true
#define FETT263_FORCE_PUSH
#else
#define FORCE_PUSH_CONDITION battle_mode_
#endif

#ifdef FETT263_FORCE_PUSH
          case EVENTID(BUTTON_NONE, EVENT_PUSH, MODE_ON):
￼        if (FORCE_PUSH_CONDITION && millis() - last_push_ > 2000) {
￼          if (SFX_push) {
￼            hybrid_font.PlayCommon(&SFX_push);
￼          } else {
￼            hybrid_font.DoEffect(EFFECT_FORCE, 0);
￼          }
￼          last_push_ = millis();
￼        }
￼        return true;
￼#endif

fdarosa2663: thanks, I have updated

PR #116, opened 2020-10-26
Added controls

fdarosa2663:
I've added an additional Multi-Blast mode control per user idea, in Battle Mode if you do a Blast with Aux button and swing within 3 seconds it will automatically enter Multi-Blast mode. 
And I've added control for Start or Stop Track while blade is on per user request using hold PWR with blade pointing straight up, this enables users to control tracks while blade is on.

### DIFF
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -11,9 +11,8 @@
 // Retract (OFF) - click PWR while ON (disabled during swinging, Twist Off available with define)
 // Play/Stop Music Track - hold and release PWR while OFF or hold and release PWR while ON and pointing blade straight up
 // Blast - click AUX while ON
-// Multi-Blast Mode - hold and release AUX while ON or in Battle Mode, click AUX and swing within 3 seconds to enter Mode,
-//                    while in Multi-Blast Mode every swing will trigger a Blast effect.
-//                    click AUX or lockup, clash, stab, melt, drag or any button presses automatically exits Mode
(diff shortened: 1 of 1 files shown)


props/saber_fett263_buttons.h
```diff
@@ -9,10 +9,11 @@
 // Ignite (ON) - click PWR while OFF (Swing On, Twist On and Stab On available with defines)
 // Muted Ignition (ON) - double click PWR while OFF
 // Retract (OFF) - click PWR while ON (disabled during swinging, Twist Off available with define)
-// Play Music Track - hold and release PWR while OFF
+// Play/Stop Music Track - hold and release PWR while OFF or hold and release PWR while ON and pointing blade straight up
 // Blast - click AUX while ON
-// Multi-Blast Mode - hold and release AUX while ON to enter mode, Swing to initiate Blasts, click Aux to exit mode
-//                    lockup, clash, stab, melt, drag or any button presses automatically exits mode
+// Multi-Blast Mode - hold and release AUX while ON or in Battle Mode, click AUX and swing within 3 seconds to enter Mode,
```
profezzorn, line 14: Are you sure this is a good idea?
It seems like it could be pretty annoying if that's not what you want.

fdarosa2663: Valid point, I've removed if there's a need will revisit in a future OS, the previous control was easy and intuitive enough, the additional control wasn't necessary.
I'm keeping the Start or Stop track with Hold PWR while pointing up so there's a control for tracks while the blade is ON, that's something I see a need for.

PR #118, opened 2020-11-01
Update ssd1306.h for 180 degrees optional rotation

Aleyan82:
Adding ability to rotate 180 degrees the OLED mapping as a define.
Code by Brian Conner


NoSloppy:
Actually, your comments should just swap out. The first section is normal operation (if flip NOT defined) , then the #else is the flipped version.

Aleyan82 (commented):
Comments on the function are inverted between normal and flipped operation

Aleyan82 (commented):
Functions have been mistakenly inverted.

PR #127, opened 2021-01-23
OLED idle.bmp and Aurebesh

NoSloppy:
Added idle.bmp playback if it exists in place of constant battery meter. Battery still shows on boot until the first time blade on/off so that Mass Storage can still work. This shouldn't bog anything down (I think) and works fine for users who don't have the file in the font.

Also added an Aurebesh font option for SCREEN_MESSAGE on a #define AUREBESH.
I've attached a nifty idle.bmp file I made if you want to try it with this,
and I think the font looks good!
What do you think?
[idle.bmp.zip](https://github.com/profezzorn/ProffieOS/files/5860235/idle.bmp.zip)


display/ssd1306.h
```diff
@@ -203,6 +206,8 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
 
       case SCREEN_PLI:
         memset(frame_buffer_, 0, sizeof(frame_buffer_));
+      // Need non-SDcard content at first so Mass Storage can work.
+          if (!just_booted_) ShowFile(&IMG_idle, font_config.ProffieOSOnImageDuration);
```
profezzorn, line 210: It seems weird that we don't do an } else { or something here to avoid doing a showfile AND draw a battery bar.

display/ssd1306.h
```diff
@@ -211,11 +216,20 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
 
       case SCREEN_MESSAGE:
         memset(frame_buffer_, 0, sizeof(frame_buffer_));
+    // Aurbesh Font option.
+    #ifdef AUREBESH
```
profezzorn, line 220: we can avoid a little duplication here by doing something like:

#ifdef AURABESH
   const Glyph* font = Aurabsh10pt7Glyphs;
#else
   const Glyph* font = Starjedi10pt7bGlyphs;
#endif

profezzorn: How about we use "USE_AUREBESH_FONT" instead of just "AURBESH" ?

display/ssd1306.h
```diff
@@ -312,6 +328,8 @@ class SSD1306 : public I2CDevice, Looper, StateMachine, SaberBase, private Audio
 
   void SB_On() override {
     ShowFile(&IMG_on, font_config.ProffieOSOnImageDuration);
+    // Make idle.bmp wait until first time blade on
```
profezzorn, line 331: but why?

display/Aurebesh10Font.h
```diff
@@ -0,0 +1,1507 @@
+const uint8_t Aurebesh10pt7bChar0[] = {
```
profezzorn, line 1: What is the licenses of the aurabesh font?
Are we allowed to use it in ProffieOS?
Is the license compatible with GPL?

display/ssd1306.h
```diff
@@ -42,6 +43,8 @@ const uint32_t BatteryBar16_data[] = {
 const Glyph BatteryBar16 = { 16, 0, 0, GLYPHDATA(BatteryBar16_data) };
 
 #include "StarJedi10Font.h"
+#include "Aurebesh10Font.h"
+//#include "StarJediRounded10Font.h"
```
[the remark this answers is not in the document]
NoSloppy: Agreed, it's a left over.
I was seeing if the other variant looked any better.

display/ssd1306.h
NoSloppy: It works perfectly like this in my testing. Unless you mean in the background it's ALSO drawing battery bar?

display/ssd1306.h
NoSloppy: That is more descriptive, yes.

display/ssd1306.h
NoSloppy: This is echoed in the comment above where we need non-SD card content initially so that Mass Storage option will work. Otherwise, the board never releases the SD card for mounting to the desktop. Plus, I wasn't looking to completely hide the battery bar. It's nice to have it at boot to see where the battery level is at.

display/Aurebesh10Font.h
NoSloppy: The one I converted is by BobaFonts, actually named Aurek-Besh. 
https://www.fontspace.com/boba-fonts
It looks like it maybe the same source as StarJedi?
Licencing is Personal Use, defined there as "typically refers to any use that is not meant to generate profit."
Something specifically labeled GPL would be this version:
https://fonts2u.com/aurebesh-af-canon.font
It seems there's a lot out there, so considering it's re-formatted using fontconvert as well, I don't think there's an issue? I'd be happy to investigate further if need be.

profezzorn: Where did you see the "personal use" license? Because that could be a problem.
From what I can see, both StarJedi and Aurek-Besh were made by the same person, and they seem to be licensed as freeware with no restrictions., which would mean that everything is fine.

display/ssd1306.h
profezzorn: I didn't get that part from the comments (which means that the comments needs to be improved.)
Also, I think we might need the SD card to be released if you plug it into USB after turning it on and off too.
Maybe we just play idle immediately on boot but stop doing it as soon as USB is connected IF mass storage is on...

display/Aurebesh10Font.h
NoSloppy: Oh. It was on the FAQ page of that site, https://www.fontspace.com/help#license-3

NoSloppy: But the font's page specifically says Licenced as Freeware

display/ssd1306.h
NoSloppy: I'm not sure how to use Glyph* font in place of the existing way it's coded, sorry. Is it used as an alias where the  DrawText line calls for the specified glyphs?

display/ssd1306.h
NoSloppy: Also, the whole naming convention of "idle.bmp" is somewhat conflicting with the OffType OFF_IDLE, no? I realized that once I got down to the screen sleeping part.

NoSloppy:
What can I do next to move this forward?

display/ssd1306.h
NoSloppy: Fixed.

display/ssd1306.h
profezzorn: You just use "font" instead of Aurabsh10pt7Glyphs/Starjedi10pt7bGlyphs

display/ssd1306.h
profezzorn: I think idle.bmp is fine.
The other option is just "off.bmp" I guess.

display/Aurebesh10Font.h
profezzorn: I think the license is fine then.

display/ssd1306.h
NoSloppy: done.

display/ssd1306.h
NoSloppy: > I didn't get that part from the comments (which means that the comments needs to be improved.)
Updated - added "so Mass Storage can work" there as well.

> Also, I think we might need the SD card to be released if you plug it into USB after turning it on and off too.
Just typing reset in serial monitor or toggling a kill key/switch remounts the SD card. Do you feel that is too much to ask end users?
> Maybe we just play idle immediately on boot but stop doing it as soon as USB is connected IF mass storage is on...
I really think the battery meter is still nice to have at boot, no? But stopping idle.bmp playing on USB connection is cool, yes. I'm just not familiar enough with the code for detecting USB communication. Something like:
if (USBD_Connected() BECOMES_TRUE && MASS STORAGE IS ON... MTP SOMETHING?**) then stop the loop.
I feel like it's a double negative and I'm on the wrong course.

profezzorn: > Just typing reset in serial monitor or toggling a kill key/switch remounts the SD card. Do you feel that is too much to ask end users?

Yes, it is inconvenient. Also, the reset command tends to cause problems if used incorrectly, so I don't want to rely on everybody using that.

I don't idle.bmp should start after boot. It could be nice to have an option to show the battery meter first, but some people might not want that if it doesn't go with the design they have in mind....

For the condition, I think all we need is:

#ifdef USB_CLASS_MSC
   if (USBD_Connected()) { /* stop loop here */ }
#endif

I don't think we need to worry about MTP since MTP works without unmounting the SD card.

display/ssd1306.h
NoSloppy: Oh, this is generating this error when "font" is used in place of the the Glyphs :

In file included from /ProffieOS.ino:1627:
sketch/display/ssd1306.h: In member function 'int SSD1306::FillFrameBuffer()':
sketch/display/ssd1306.h:242:12: error: jump to case label
  242 |       case SCREEN_IMAGE:
      |            ^~~~~~~~~~~~
sketch/display/ssd1306.h:227:20: note:   crosses initialization of 'const Glyph* font'
  227 |       const Glyph* font = Starjedi10pt7bGlyphs;
      |                    ^~~~
exit status 1
Error compiling for board Proffieboard V2.

profezzorn: Usually the way to fix this is to put a { after the case and a } somewhere before the "break".
This creates a scope for the "font" variable which doesn't cross any "case" jump targets.

display/ssd1306.h
NoSloppy: I think there's another piece to the puzzle I don't have. The piece you provided above looks simple enough, but I think it needs expansion for it to work on both connect and disconnect USB. The closest I can get to having USB trigger the battery meter on/off instead of idle.bmp works like: With no Mass Storage option, all is well. idle.bmp all when expected. Turning Mass Storage on, when boot/upload with USB connected, it shows batt, SD card mounts to computer fine, pulling usb does idle again. But having it stop idle.bmp when reconnecting USB is eluding me. I have USB detection in the loop for SCREEN_PLI section, (the seems to be working part), but something similar must have to go in the SCREEN_IMAGE loop, as idle.bmp would need to get interrupted there when USB is connected. I have tried the conditions in both area, variants of it, etc... and inexplicably I see the battery meter when SB_Off happens even though the conditions are not met, and/or I see a blip of idle.bmp play but then it goes back to battery meter. Tough to word. Is there another condition, such as when the SD card gets mounted/ unmounted that could be used for detection? 

The commented out parts is where I've been unsuccessful and playing around with the reconnect part. The SCREEN_PLI section seems to handle unplugging correctly. And why would SB_OFF be showing battery meter at all if no USB connected?
Commiting update now.

display/ssd1306.h
NoSloppy: Fixed

display/ssd1306.h
NoSloppy: Ok. update. I think I found why I'm going crazy here. 
It seems that USB is not being detected reliably as connected or not.
I added some prints to the void Loop to monitor the status via a serial connection while I plugged in/out USB cable.
It is reporting connected although it is not, with a single report of truth (not connected) approx every 1 minute.
This is what's throwing off my attempts above, and I'm not sure if it's a buffer thing, or if we'd need to prioritize the USBD_Connected() polling somehow?
Here's the results of the monitoring. This is all completely with USB unplugged:

22:02:28.360 -> --------------------while(true)
22:02:28.360 -> --------------------USBD_Connected
22:02:29.337 -> --------------------while(true)
22:02:29.337 -> --------------------USBD NOT Connected
22:02:30.351 -> --------------------while(true)
22:02:30.351 -> --------------------USBD_Connected
22:02:31.368 -> --------------------while(true)
22:02:31.368 -> --------------------USBD_Connected

A summary of just the NOT connected prints shows these irregular intervals:
21:59:49.209 -> --------------------USBD NOT Connected
22:01:09.269 -> --------------------USBD NOT Connected
22:02:29.337 -> --------------------USBD NOT Connected
22:03:49.413 -> --------------------USBD NOT Connected
22:04:20.175 -> --------------------USBD NOT Connected
22:04:57.214 -> --------------------USBD NOT Connected
22:05:40.254 -> --------------------USBD NOT Connected

Thoughts?

profezzorn: That is weird and unexpected.
Can I see the code you used to print that?

NoSloppy: Yes, it's updated in the latest commit.
Right under where Sidpay is Initialized, I found that that loop is running contantly in tandem with the Screen switch, so I stuck it in there  ¯\_(ツ)_/¯

    while (true) {
      STDOUT << "--------------------while(true)\n";
      if (USBD_Connected()) {
        STDOUT << "--------------------USBD_Connected\n";
      } else {
        STDOUT << "--------------------USBD NOT Connected\n";
      }
      millis_to_display_ = FillFrameBuffer();
      frame_start_time_ = millis();
      lock_fb_ = true;
      frame_available_ = false;

NoSloppy: FYI I never get NOT connected message when it's in it's own if condition like this instead:

  while (true) {
      STDOUT << "--------------------while(true)\n";
      if (USBD_Connected()) {
        STDOUT << "--------------------USBD_Connected\n";
      } 
      if (!USBD_Connected()) {
        STDOUT << "--------------------USBD NOT Connected\n";
      }
      millis_to_display_ = FillFrameBuffer();
      frame_start_time_ = millis();

profezzorn: That doesn't make any sense at all.

NoSloppy: Agreed. Been pulling my hair out for a week on this.
When USB is unplugged, the single instance of NOT connected triggers idle.bmp, which plays once completely until it finishes its frames. But during that playback, USBD_Connected() has reverted back to true in the background for some reason, so at END OF LOOP, when we go back to SCREEN_PLI, we now get DrawBatteryBar instead until the next single instance of NOT connected...and repeat the cycle.

profezzorn: How about you just make it work and show the idle.bmp all the time and I'll figure out how to make it not do it when USB is connected?

NoSloppy: Well sure that's super easy at this point. To clarify, this would be intentionally submitting something that breaks mass storage in lieu of you figuring out that part, as opposed to just reverting back to where I had idle.bmp just waiting until first ignition?

NoSloppy:
The new I2C updates cause the idle.img to freeze when I2C goes to sleep :(

profezzorn:
I think the sleeping problem was fixed in the latest update, wasn't it?
Obviously, we shouldn't put i2c to sleep if it's still being used.

NoSloppy:
Yes, working good as of now. This was left with odd behavior detecting USB connected/disconnected, and you were going to maybe look into that.
I am still running "my" version where the just_booted_ check allows the battery meter to show until first ignition, then idle.bmp from then until power cycle, with just resetting to mount SD if Mass Storage on. Works for me, but just let me know what else I could do for this version here to get merged.

profezzorn:
I'm afraid this PR has quite a few conflicts now.
Sorry about that.

PR #132, opened 2021-03-13
add sum.h function

fdarosa2663:


fdarosa2663:
I created a new fork and PR, not sure why the changes wouldn't match up they were the same in ino.  This one shows no conflicts.

PR #137, opened 2021-03-18
Update saber_sa22c_buttons.h

NoSloppy:
Edits to volume menu allowing for audio wavs to be used for up, down, min and max, with beeps if they don't exist.
Also, with #define VOLUME_MENU_CYCLE, when you reach the maximum volume, you can get back to minimum by pressing volume up again, and vice versa. 
Basically it's a loop now instead of a closed end range.
Checked with Matthew McGeary, he has no issues with it.

### DIFF
--- a/props/saber_sa22c_buttons.h
+++ b/props/saber_sa22c_buttons.h
@@ -180,16 +180,20 @@
 
 #define FORCE_PUSH_CONDITION battle_mode_
 
-EFFECT(dim); // for EFFECT_POWERSAVE
-EFFECT(battery); // for EFFECT_BATTERY_LEVEL
-EFFECT(bmbegin); // for Begin Battle Mode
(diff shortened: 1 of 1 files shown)


profezzorn:
I have no idea why anybody would want a sound menu that goes in a loop. Seems crazy, but hey, to each their own, right?  The code looks pretty good, so I'm going to merge it.

NoSloppy:
Thanks! Haha well, maybe crazy, but I had several people ask if it was possible to have it do "round robin", which I didn't understand, but this is what they meant.
I found it similar in concept to say an onscreen keyboard on Netflix to enter your email address. Wrap around through the edge of the screen is a shorter path sometimes.

NoSloppy:
Missed an update on one ChangeVolume(true); -> VolumeUp();  Sorry. Glad you found it.

PR #138, opened 2021-03-18
updates for file_reader and current_preset

fdarosa2663:
additions to enable save float values and SetStyle/GetStyle for current_preset

### DIFF
--- a/common/current_preset.h
+++ b/common/current_preset.h
@@ -317,13 +317,6 @@ public:
     LOCK_SD(false);
   }
 
-void SetStyle(int blade, const char* style) {
-#define SET_STYLE_N(N) case N: current_style##N = style; break;
-  switch (blade) {
-    ONCEPERBLADE(SET_STYLE_N)
-  }
-}
-
 LSPtr<char> SetStyle(int blade, LSPtr<char> style) {
 #define SET_STYLE_N(N) case N: std::swap(current_style##N, style); break;
   switch (blade) {
--- a/common/current_preset.h
+++ b/common/current_preset.h
@@ -316,6 +316,30 @@ public:
     if (!Load(preset)) Set(preset);
     LOCK_SD(false);
   }
+
+void SetStyle(int blade, const char* style) {
+#define SET_STYLE_N(N) case N: current_style##N = style; break;
+  switch (blade) {
+    ONCEPERBLADE(SET_STYLE_N)
+  }
+}
+
+LSPtr<char> SetStyle(int blade, LSPtr<char> style) {
+#define SET_STYLE_N(N) case N: std::swap(current_style##N, style); break;
+  switch (blade) {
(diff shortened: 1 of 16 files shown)


common/current_preset.h
```diff
@@ -316,6 +316,30 @@ class CurrentPreset {
     if (!Load(preset)) Set(preset);
     LOCK_SD(false);
   }
+
+void SetStyle(int blade, const char* style) {
```
profezzorn, line 320: I think this one is dangerous and we should not have it unless we really need it for some reason.

common/file_reader.h
```diff
@@ -410,6 +410,17 @@ class FileReader {
     itoa(v, value, 10);
     write_key_value(key, value);
   }
+  void write_key_value_float(const char* key, float f) {
+    int  a_ = (int)floorf(f);
+    int  b_ = ((int)floorf(f * 10)) % 10;
+    int  c_ = ((int)floorf(f * 100)) % 10;
+    char new_value[10];
```
profezzorn, line 417: This needs to be bigger.
An int can take 10 bytes, then '.' and 2 decimals, plus a zero.
Just make it 16 bytes.

common/file_reader.h
```diff
@@ -410,6 +410,17 @@ class FileReader {
     itoa(v, value, 10);
     write_key_value(key, value);
   }
+  void write_key_value_float(const char* key, float f) {
+    int  a_ = (int)floorf(f);
```
profezzorn, line 414: variables with underscores are meant for class-level variables.
Also, we don't need variables for this as all as we can just inline the floorf() call in the itoa below.

common/file_reader.h
```diff
@@ -410,6 +410,17 @@ class FileReader {
     itoa(v, value, 10);
     write_key_value(key, value);
   }
+  void write_key_value_float(const char* key, float f) {
```
profezzorn, line 413: Add a comment that this always uses two decimals.

common/current_preset.h
fdarosa2663: this one is used throughout Edit Mode for styles, it was something you provided when we first began.

common/file_reader.h
fdarosa2663: I have made changes

common/file_reader.h
fdarosa2663: comment added

common/file_reader.h
fdarosa2663: I have made 16 bytes

common/current_preset.h
profezzorn: I think we should remove this one and use the one below instead.

fdarosa2663: ok, removed the void SetStyle(), will need to implement and test on Edit Mode version to see what changes are needed in prop.

PR #142, opened 2021-03-25
Add TrCenterWipeX, TrCenterWipeInX

fdarosa2663:


transitions/center_wipe.h
```diff
@@ -0,0 +1,88 @@
...
+// starts at the POSTION and extends up and down the blade
+// in the specified number of milliseconds.
+template<class MILLIS, class POSITION = Int<16384>>
+class TrCenterWipeX : public TransitionBaseX<MILLIS> {
+public:
+  void run(BladeBase* blade) {
+    pos_.run(blade);
+    TransitionBaseX<MILLIS>::run(blade);
+    center_ = (pos_.getInteger(0) * blade->num_leds()) >> 7;
+    fade_top_ = this->update(256 * blade->num_leds() - center_);
```
profezzorn, line 24: center_, fade_top_ and fade_bottom_ is only used in this function, so they should be local variables.

transitions/center_wipe.h
```diff
@@ -0,0 +1,88 @@
...
+  auto getColor(const A& a, const B& b, int led) -> decltype(MixColors(a,b,1,1)) {
+        int mix = (Range(bottom_, top_) & Range(led << 8, (led << 8) + 256)).size();
+    return MixColors(a, b, mix, 8);
+  }
+private:
+  POSITION pos_;
+  uint32_t fade_top_;
+  uint32_t fade_bottom_;
+  uint32_t center_;
+  uint32_t top_;
```
profezzorn, line 40: replace top_ and bottom_ with: Range range_;

transitions/center_wipe.h
fdarosa2663: fixed

transitions/center_wipe.h
fdarosa2663: I could not get it to compile when trying to replace, how should it be done?

profezzorn: Replace top_ and bottom_ with:

   Range range_;

Make top and bottom local variables, and then at the end of Run(), do:

   range_ = Range(bottom, top);

in getColor, replace Range(botom, top) with range_.

same below

transitions/center_wipe.h
profezzorn: did you forget to upload or something? It doesn't look fixed.

fdarosa2663: yeah, looks like I pushed a previous, working on the change you suggested but still getting compile error, sending in PM on TRA as it's odd.

transitions/center_wipe.h
fdarosa2663: Range range_; added and tested, thank you.

profezzorn:
I think this is ready to be merged, but there is a conflict in ProffieOS.ino

PR #144, opened 2021-03-28
Update hybrid_font.h

NoSloppy:
This is not where I wanted to do Spins auto-blast, should probably be in BC prop file, but it only gets detected here...?
Adds optional uses of:
#define ENABLE_AUTO_SPINS_BLAST
also
#define LOW_BATT_ONCE

### DIFF
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -245,7 +245,16 @@ public:
               if (SFX_spin) {
                 swing_player_ = PlayPolyphonic(&SFX_spin);
               }
-              angle_ -= font_config.ProffieOSSpinDegrees;
+#ifdef ENABLE_AUTO_SPINS_BLAST
+              if (millis() - last_blast_ < 2000) {
+                SaberBase::DoBlast();
+                last_blast_ = millis();
+                  STDOUT.println("Entering Auto Spin Blast mode");
+              } else {
+                STDOUT.println("Not in Auto Spin Blast mode");
+              }
+#endif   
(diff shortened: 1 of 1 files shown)


profezzorn:
I think both of these features blongs in the prop where the these things are detected, not in hybrid_font.
Please re-open if you disagree for some reason.

NoSloppy:
> Please re-open if you disagree for some reason.
I do not disagree at all, I know they belong in the prop. I am just not seeing this with an untrained eye how to accomplish it that way. 
Could you help me understand how to detect spins in the prop?
I could only see a place in hybrid_font as it's a sub-state of swings, so I put it where I saw &SFX_spin was being played.
There's no EVENT_SPINS, else I'd just have that be a case EVENTID in the prop.
I tried something like 
if (swing_player_ = PlayPolyphonic(&SFX_spin)) {
but I don't think that player carries over to another file, nor did angle_ trying 
if (angle_ > font_config.ProffieOSSpinDegrees) {

Also, low batt is only in hybrid_font. Would it "override" that whole void by putting this whole thing in the prop? Is that how override works? 
  void SB_LowBatt() override {
    // play the font's low battery sound if it exists  
    if (!low_batt_notify_) {             
      if (SFX_lowbatt) {
        PlayCommon(&SFX_lowbatt);
      } else {
  #ifdef ENABLE_AUDIO
        talkie.Say(talkie_low_battery_15, 15);
  #endif
      }
  #ifdef LOW_BATT_ONCE           // BC add for limiting Low Batt notifiation to once
      low_batt_notify_ = true;
  #endif    
    }
  }

PR #145, opened 2021-03-28
Update voice_data.h

NoSloppy:
Added female voice to match my custom menu wavs, but I'm not sure if it's ok to outright replace the male voice. Commented out unused data to save some memory (?)


profezzorn:
1) There is no need to comment anything out. Unused entries does not get included in the final output and thus  use no memory.
2) Where did this voice data come from? What is the license?
3) We'll want a define that controls which entries to use instead of just replacing the entries.

NoSloppy:
Same Talkie Library source. I added the info and ROM number.
Also added define to use female voice as an option.

NoSloppy:
Hey Fredrik,, so I know you said if it's not called for it's not considered, but trying to compile this version now, I get errors referencing words that are not used anywhere:
sketch/sound/talkie.h:721:11: error: 'spMILLION' was not declared in this scope
  721 |       Say(spMILLION);
      |           ^~~~~~~~~
sketch/sound/talkie.h:726:11: error: 'spTHOUSAND' was not declared in this scope
  726 |       Say(spTHOUSAND);

profezzorn:
Just use the thousand/million outtside of the ifdef. If they aren't used it doesn't matter.
If they are used it sounds a little funny, but still works.

NoSloppy:
yes, fixed and separate pull request sent, thanks.

PR #146, opened 2021-03-28
Update saber_base.h. add SWAP and MONOFORCE

NoSloppy:
Add SWAP and MONOFORCE - just extra features that can be used with BC prop file. 
SWAP - just another EFFECT to use to trigger EffectSequence to avoid being locked into using an existing effect like blast, clash, or tying up colorchange mode to do the swap, nulling access to the color wheel.
Used with BC prop file to "swap" blade styles within the same preset, or anything within EffectSequence really. Uses custom swap.wavs.
MONOFORCE is simply force effect that intentionally triggers a monophonic wav for sounds where the hum is silenced (seismic charge, iceblade, open for future conceptual use) Currently does not trigger a blade animation as invoking DoForce would collide with regular force sound.

### DIFF
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -36,6 +36,8 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(POWERSAVE)                    \
     DEFINE_EFFECT(BATTERY_LEVEL)                \
     DEFINE_EFFECT(FAST_ON)                      \
+    DEFINE_EFFECT(SWAP)            		\
+    DEFINE_EFFECT(MONOFORCE)       		\
     /* Blaster effects */                       \
     DEFINE_EFFECT(STUN)				\
     DEFINE_EFFECT(FIRE)				\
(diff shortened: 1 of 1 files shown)


common/saber_base.h
```diff
@@ -36,6 +36,8 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(POWERSAVE)                    \
     DEFINE_EFFECT(BATTERY_LEVEL)                \
     DEFINE_EFFECT(FAST_ON)                      \
+    DEFINE_EFFECT(SWAP)            		\
```
profezzorn, line 39: Why not use USER1 instead of SWAP?

common/saber_base.h
```diff
@@ -194,6 +196,7 @@ public:                                                         \
   static void DoClash() { DoEffectR(EFFECT_CLASH); }
   static void DoBlast() { DoEffectR(EFFECT_BLAST); }
   static void DoForce() { DoEffectR(EFFECT_FORCE); }
+  static void DoMonoForce() { DoEffectR(EFFECT_MONOFORCE); }
```
profezzorn, line 199: why does MONOFORCE get a Do* function but not SWAP?

common/saber_base.h
```diff
@@ -36,6 +36,8 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(POWERSAVE)                    \
     DEFINE_EFFECT(BATTERY_LEVEL)                \
     DEFINE_EFFECT(FAST_ON)                      \
+    DEFINE_EFFECT(SWAP)            		\
+    DEFINE_EFFECT(MONOFORCE)       		\
```
profezzorn, line 40: Why not USER2 instead of MONOFORCE?

common/saber_base.h
NoSloppy: I think because MONOFORCE can do a blade effect and sound, while SWAP is just event used as a trigger with a sound.

common/saber_base.h
NoSloppy: Didn't even see that , and was overthinking the sound vs blade effects...it works very easily now. Thanks.

PR #152, opened 2021-04-09
add get_min_file() and get_directory()

fdarosa2663:


sound/effect.h
```diff
@@ -218,6 +218,10 @@ class Effect {
     return ret;
   }
 
+  size_t get_min_file() { return min_file_;}
```
profezzorn, line 221: space after semicolon
This function should be "const": size_t get_min_file() const { return min_file_; }
Also applies to get_directory below.

fdarosa2663: thanks, updated.

PR #153, opened 2021-04-11
update saber_base for define Add ENABLE_ALL_MENU_OPTIONS define and COLOR_CHANGE_ZOOMED

fdarosa2663:


fdarosa2663:
Just pinging to remind

common/saber_base.h
```diff
@@ -76,6 +76,16 @@ DEFINE_ALL_EFFECTS();
 constexpr size_t NUMBER_OF_EFFECTS = (size_t)EffectTypeHelper::NUMBER_OF_EFFECTS;
 #undef DEFINE_EFFECT
 
+#ifdef ENABLE_ALL_MENU_OPTIONS
```
profezzorn, line 79: Can't this go in your prop file?

common/saber_base.h
```diff
@@ -277,7 +287,8 @@ public:                                                         \
   enum ColorChangeMode {
     COLOR_CHANGE_MODE_NONE,
     COLOR_CHANGE_MODE_STEPPED,
-    COLOR_CHANGE_MODE_SMOOTH
+    COLOR_CHANGE_MODE_SMOOTH,
+    COLOR_CHANGE_MODE_ZOOMED,
```
profezzorn, line 291: Seems like this needs an implementation somewhere?
Is it just in your prop file?

common/saber_base.h
fdarosa2663: I had tried but the DYNAMIC_BLADE_DIMMING code below was missing causing compile error, it fixed when I moved to here.

common/saber_base.h
fdarosa2663: I have the corresponding code in prop_base but you asked me to submit 1 file at a time

common/saber_base.h
profezzorn: You just need to put it before the #include "prop_base.h" line.

common/saber_base.h
profezzorn: it's supposed to be one *feature* at a time, not one file at a time.

common/saber_base.h
fdarosa2663: OK, I thought I tried that but I'll do it again and re-test tomorrow.

common/saber_base.h
fdarosa2663: ah, sorry misunderstood.  ok let me test the define tomorrow and I'll get the corresponding code for prop_base added

common/saber_base.h
fdarosa2663: removed

common/saber_base.h
fdarosa2663: submitted prop_base addition

props/prop_base.h
```diff
@@ -958,6 +958,25 @@ class PropBase : CommandParser, Looper, protected SaberBase {
         }
         break;
       }
+      case SaberBase::COLOR_CHANGE_MODE_ZOOMED: {
+#define ZOOM_ANGLE (M_PI * 2 / 2000)
```
profezzorn, line 962: For 2000 steps per revolution, we need code that can calculate how many steps to go instead of just going one step at a time.

props/prop_base.h
```diff
@@ -958,6 +958,25 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+          current_tick_angle_ += ZOOM_ANGLE;
+          if (current_tick_angle_ > M_PI) current_tick_angle_ -= M_PI * 2;
+          STDOUT << "TICK+\n";
+          SaberBase::UpdateVariation(1);
+        }
+        if (a < -ZOOM_ANGLE * 2/3) {
+          current_tick_angle_ -= ZOOM_ANGLE;
+          if (current_tick_angle_ < M_PI) current_tick_angle_ += M_PI * 2;
+          STDOUT << "TICK-\n";
+          SaberBase::UpdateVariation(-1);
```
profezzorn, line 976: This can make the variation negative.
For "ticked" variations that's not an issue, but I'm not sure how well it works for smooth variations.
(The  code for COLOR_CHANGE_SMOOTH keeps the values in the 0 - 32768 range.)

props/prop_base.h
fdarosa2663: COLOR_CHANGE_ZOOMED is used while the PWR button is held after using COLOR_CHANGE_SMOOTH to allow user to dial in color better, it's not a standalone mode.  When PWR is released we ToggleColorChangeMode() to then SaveVariation.  It's intended to go from SMOOTH to STEPPED once the user has nearly the right color to allow for fine-tuning before saving.  I did a lot of testing to get the 2000 value so that it both gave you the ability to still fine-tune but also made it easy to stay in range of selected color with button depressed.

props/prop_base.h
fdarosa2663: It was based on STEPPED, should we apply a clamp in SaberBase::UpdateVariation to prevent negative value?

props/prop_base.h
profezzorn: I have no problem with the 2000 value, but I don't think the ticked code is very suitable.
Instead I think we should do something like:

float a = fusor.angle2() - current_tick_angle_; 
if (a > M_PI) a-=M_PI*2;
if (a < -M_PI) a+=M_PI*2;
int steps = (int)floor(fabs(a) / ZOOM_ANGLE - 0.3);
if (steps < 0) steps = 0;
if (a < 0) steps = -steps;
current_tick_angle_ += ZOOM_ANGLE * steps;
SaberBase::SetVariation(0x7fff & (SaberBase::GetVariation() + steps));

It should be pretty similar, but be more consistent if you turn back and forth I think.

fdarosa2663: ok, I will test this and make changes, thanks.

fdarosa2663: SaberBase::SetVariation(0x7fff & (SaberBase::GetCurrentVariation() + steps));

this is not working as intended but unable to get it work properly.  With ZOOMED you are in COLOR_CHANGE_SMOOTH to rotate through all colors, then you press the Pwr button and it switches to ZOOMED as long as Pwr is pressed letting you fine tune, when you release that color is then saved.  The above is resetting the color when Pwr is pressed back to original and then adjusting instead of allowing adjustment in the color you had rotated to before pressing the button.  I tried switching to UpdateVariation but it is not working so not sure how to modify this to work as original code did.

fdarosa2663: fixed, should be good now.

PR #155, opened 2021-04-16
"Font directory" spoken as "font folder" instead

NoSloppy:
Not sure if you're into this, but it seems most average users don't call it a directory, and/or have difficulty understanding the talkie.
Completely doesn't matter, but maybe?
The rest of the code still uses talkie_font_directory_15.


profezzorn:
How does this actually sound?

NoSloppy:
> How does this actually sound?


https://user-images.githubusercontent.com/53964195/116773623-6b428c00-aa24-11eb-8542-f0cdf5845dcb.mov

NoSloppy:
Like that

profezzorn:
Hmm, I'm not sure I like it, "font folder" just doesn't sound as accurate as "font directory" to me.
Also, it takes more space.

PR #157, opened 2021-04-21
V5.x

sledulous:
Config defines for the following:

Degrees of rotation between color change "ticks"

Motion Chip Timeout/Or make motion chip timeout tied to idle time off?

FindBladeID upon preset change.

:)


profezzorn:
Too many changes in one PR.
Also, no change should go directly into the V5.x branch.

PR #158, opened 2021-04-22
Define

sledulous:
Add config define for Motion Chip Timeout, or have controlled by idle time off?


profezzorn:
There is something wrong here, because there are 84 changed files and this PR is still targeting the v5.x branch.

PR #160, opened 2021-04-23
Add DYNAMIC_CLASH_THRESHOLD

fdarosa2663:


props/prop_base.h
```diff
@@ -152,7 +162,13 @@ class PropBase : CommandParser, Looper, protected SaberBase {
       SetMute(false);
     }
   }
-
+	
+#ifdef DYNAMIC_CLASH_THRESHOLD
+  static float clash_threshold_;
```
profezzorn, line 167: no need for these three to be static

props/prop_base.h
```diff
@@ -1560,4 +1590,8 @@ class PropBase : CommandParser, Looper, protected SaberBase {
   LoopCounter accel_loop_counter_;
 };
 
+#ifdef DYNAMIC_CLASH_THRESHOLD
```
profezzorn, line 1593: if not static, this can be removed

props/prop_base.h
```diff
@@ -152,7 +162,13 @@ class PropBase : CommandParser, Looper, protected SaberBase {
       SetMute(false);
     }
   }
-
+	
+#ifdef DYNAMIC_CLASH_THRESHOLD
+  static float clash_threshold_;
+  static float GetCurrentClashThreshold() { return clash_threshold_; }
+  static void SetClashThreshold(float clash_threshold) { clash_threshold_ = clash_threshold; }
+#endif
```
profezzorn, line 170: In here, we could also add:

#undef CLASH_THRESHOLD_G
#define CLASH_THRESHOLD_G clash_threshold_

This can simplify some of the code below.

props/prop_base.h
fdarosa2663: updated

props/prop_base.h
fdarosa2663: removed

props/prop_base.h
fdarosa2663: added and remove ifdef in DoAccel()

props/prop_base.h
```diff
@@ -609,6 +637,7 @@ class PropBase : CommandParser, Looper, protected SaberBase {
     // If we're spinning the saber, require a stronger acceleration
     // to activate the clash.
     if (v > CLASH_THRESHOLD_G + fusor.gyro().len() / 200.0) {
+#endif
```
profezzorn, line 640: Did you test this?
I don't think it compiles with this extra #endif in here....

fdarosa2663: sorry, local copy had it removed, missed on github.com change, fixed now.

PR #161, opened 2021-04-23
Add AlphaMixL<>

fdarosa2663:


styles/alpha.h
```diff
@@ -38,4 +38,6 @@ class AlphaL {
   }
 };
 
+template<class MIX, class... COLORS> using AlphaMixL = AlphaL<Mix<MIX, COLORS...>, MIX>;
```
profezzorn, line 41: Needs comment.

fdarosa2663: added

styles/alpha.h
```diff
@@ -38,4 +38,7 @@ class AlphaL {
   }
 };
 
+// To enable Gradient/Mixes constricted within Bump<> and SmoothStep<> layers
```
profezzorn, line 41: Maybe add an example to make it clear how to use it?

fdarosa2663: sure, added

profezzorn:
I had to revert this change because some of the tests failed.
I think you need to add an #include "mix.h" to prevent that.

PR #164, opened 2021-04-23
Add Ramp<>

fdarosa2663:


functions/ramp.h
```diff
@@ -0,0 +1,17 @@
+#ifndef FUNCTIONS_RAMP_H
+#define FUNCTIONS_RAMP_H
+
+//Returns led as value between 0 ~ 32768
+
+class Ramp {
```
profezzorn, line 6: Actually, when I think about it, I think we should call it RampF<>
It makes it easy to see that it's a function, and it's much less likely to collide with something.

fdarosa2663: updated

functions/ramp.h
```diff
@@ -0,0 +1,17 @@
+#ifndef FUNCTIONS_RAMP_H
+#define FUNCTIONS_RAMP_H
+
+//Returns led as value between 0 ~ 32768
```
profezzorn, line 4: Needs a better description/comment.

fdarosa2663: added with use example

PR #171, opened 2021-04-23
make chdir() virtual and "list_current_tracks"

fdarosa2663:


props/prop_base.h
```diff
@@ -1506,6 +1506,17 @@ class PropBase : CommandParser, Looper, protected SaberBase {
       return true;
     }
 
+    if (!strcmp(cmd, "list_current_tracks")) {
```
profezzorn, line 1509: Shouldn't this be a part of "list_tracks" ?

fdarosa2663: We added this to limit the tracks found during Edit Mode to only the current font (or common) folder. "list_tracks" was returning all tracks on the SD so this was added.

profezzorn: So if the user wants to pick a different track, how would they do it?

fdarosa2663: we set it so they can pick any track within the current font, this was due to many duplicates, etc if tracks were repeated in fonts and/or too many track folders overall bogging the menu down to be read.  If a user wants to select a track that is within another font they will Edit Font first, then Edit Track, then Edit the Font back.  This was the lesser of two evils to make the Edit Tracks work better.  It will be part of the documentation that tracks need to be in either the font or common folder to be selected.

profezzorn: Would it make sense to have a function that lists all tracks, but lists the "current" ones first?

fdarosa2663: I think it's better overall to keep this function, particularly with more and more fonts having character based tracks included, if users really wanted to make all tracks visible in Edit Mode they can choose to put them (or shared) tracks within the common folder.  In addition, the "list_current_tracks" is also used in the new Track Player so for the Rotate and Random player functions it works much better having the selection limited solely to the current font. Since users can set up the folders how they want having this as an option covers users that only want specific tracks per font whereas without it that wouldn't be an option and the use in the Track Player would be impacted without this, again if you want themed tracks by font to only play.  I believe it is necessary and beneficial.

fdarosa2663: moved "list_current_tracks" to prop file with override

PR #173, opened 2021-04-25
Add LinearSectionF<>

fdarosa2663:


functions/linear_section.h
```diff
@@ -0,0 +1,51 @@
...
+    Range led_range(led * 32768, led * 32768 + 32768);
+     int black_mix = 0;
+      black_mix = (Range(start_, end_) & led_range).size();
+    return black_mix;
+  }
+private:
+  POSITION pos_;
+  FRACTION fraction_;
+  
+  uint32_t start_;
```
profezzorn, line 46: replace start_ and end_ with: Range range_;

functions/linear_section.h
```diff
@@ -0,0 +1,51 @@
...
+      int pos = pos_.getInteger(0);
+      start_ = clampi32((pos - fraction / 2) * num_leds_, 0, 32768 * num_leds_);
+      end_ = clampi32((pos + fraction / 2) * num_leds_, 0, 32768 * num_leds_);
+    }
+    num_leds_ *= 32768;
+    return ret;
+  }
+  int getInteger(int led) {
+    Range led_range(led * 32768, led * 32768 + 32768);
+     int black_mix = 0;
```
profezzorn, line 38: no need for this variable, just return the value.

functions/linear_section.h
```diff
@@ -0,0 +1,51 @@
...
+      black_mix = (Range(start_, end_) & led_range).size();
+    return black_mix;
+  }
+private:
+  POSITION pos_;
+  FRACTION fraction_;
+  
+  uint32_t start_;
+  uint32_t end_;
+  uint32_t num_leds_;
```
profezzorn, line 48: num_leds_ is never used in getInteger(), so it can be a local variable in run()

functions/linear_section.h
```diff
@@ -0,0 +1,51 @@
...
+template<class POSITION, class FRACTION>
+class LinearSectionF {
+public:
+  FunctionRunResult run(BladeBase* base) {
+    pos_.run(base);
+    FunctionRunResult ret = RunFunction(&fraction_, base);
+    
+    num_leds_ = base->num_leds();
+    int fraction = fraction_.getInteger(0);
+    if (fraction == 32768) {
```
profezzorn, line 22: This optimization doesn't work for LinearSectionF and should be removed.
(Just because fraction is 1.0, doesn't mean that it covers the whole blade...)

functions/linear_section.h
fdarosa2663: updated

functions/linear_section.h
fdarosa2663: updated

functions/linear_section.h
fdarosa2663: updated

functions/linear_section.h
fdarosa2663: removed

functions/linear_section.h
```diff
@@ -0,0 +1,40 @@
...
+  }
+  int getInteger(int led) {
+    Range led_range(led * 32768, led * 32768 + 32768);
+    return (Range(range_) & led_range).size();
+  }
+private:
+  POSITION pos_;
+  FRACTION fraction_;
+  Range range_;
+  uint32_t start;
```
profezzorn, line 35: start and end should be local variables inside run()

functions/linear_section.h
```diff
@@ -0,0 +1,40 @@
...
+  uint32_t end;
+  uint32_t num_leds;
```
profezzorn, line 37: num_leds should have an underscore at the end: num_leds_

functions/linear_section.h
fdarosa2663: updated

functions/linear_section.h
fdarosa2663: fixed

functions/linear_section.h
```diff
@@ -0,0 +1,36 @@
...
+// creates a "block" of pixels at POSITION taking up FRACTION of blade
+
+class BladeBase;
+template<class POSITION, class FRACTION>
+class LinearSectionF {
+public:
+  FunctionRunResult run(BladeBase* base) {
+    pos_.run(base);
+    FunctionRunResult ret = RunFunction(&fraction_, base);
+    uint32_t num_leds_;
```
profezzorn, line 19: since this is now a local variable, it should not have any _ at the end.
Also, you can merge this with the next line into:
int num_leds = base->num_leds();

Also, I don't think this compiles right now because some of the stuff below uses "num_leds" which doesn't exist right now.

fdarosa2663: local had num_leds_, fixed on git and combined

PR #174, opened 2021-04-25
Add Fire<> alias

fdarosa2663:


styles/fire.h
```diff
@@ -152,5 +152,12 @@ StyleAllocator StyleFirePtr() {
   return &factory;
 }
 
+// Base Fire, no Clash, Lockup or Off state
```
profezzorn, line 155: Need better explanation/comment.

fdarosa2663: Updated

styles/fire.h
```diff
@@ -152,5 +152,12 @@ StyleAllocator StyleFirePtr() {
   return &factory;
 }
 
+// Simplified Fire, negates Clash, Lockup or Off FireConfig<> to use Fire with other Layered effects
+template<class COLOR1, class COLOR2, int DELAY=0, int SPEED=2, int BASE=0, int RAND=2000, int COOLING=5> 
+using Fire = StyleFire<COLOR1, COLOR2, DELAY, SPEED,
```
profezzorn, line 157: Just "Fire" is too generic a name.
Maybe StaticFire or Simplefire?

fdarosa2663: StaticFire works, updated

PR #175, opened 2021-04-25
RawHID support for Teensy based sabers

artandtechvof:
Extended the usb_rawhid class provided by Teensyduino an made  a SerialHID object /common/_usb_rawhid.h

Could not get it to be an actual Print*, therefore had to add an dedicated serial Parser in ProffieOS.ino and modified common/stdout.h to switch between serialHID and stream objects.

Only when Serial type: raw HID is selected, will the dode be actual used by the compiler.

Any suggestions are welcome to implement it better, before integrating this into ProffieOS

### DIFF
--- a/ProffieOS.ino
+++ b/ProffieOS.ino
@@ -20,7 +20,7 @@
 
 // You can have multiple configuration files, and specify which one
 // to use here.
-
+#define CONFIG_FILE "config/aat_teensy_singleblade_config.h"
 // #define CONFIG_FILE "config/default_proffieboard_config.h"
 // #define CONFIG_FILE "config/default_v3_config.h"
 // #define CONFIG_FILE "config/crossguard_config.h"
@@ -29,7 +29,7 @@
 // #define CONFIG_FILE "config/owk_v2_config.h"
 // #define CONFIG_FILE "config/test_bench_config.h"
 // #define CONFIG_FILE "config/toy_saber_config.h"
-#define CONFIG_FILE "config/proffieboard_v1_test_bench_config.h"
+// #define CONFIG_FILE "config/proffieboard_v1_test_bench_config.h"
 // #define CONFIG_FILE "config/td_proffieboard_config.h"
 // #define CONFIG_FILE "config/teensy_audio_shield_micom.h"
 // #define CONFIG_FILE "config/proffieboard_v2_ob4.h"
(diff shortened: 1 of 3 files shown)


ProffieOS.ino
```diff
@@ -1496,6 +1495,142 @@ private:
 
 StaticWrapper<Parser<SerialAdapter>> parser;
 
+#ifdef TEENSYDUINO
+#include "common/_usb_rawhid.h"
+#if defined(RAWHID_INTERFACE)
+class RawHIDParser: Looper, StateMachine  {
```
profezzorn, line 1501: Why do we need to duplicate the whole parser?
Isn't it enough to use the existing parser with a different adapter?

common/_usb_rawhid.h
```diff
@@ -0,0 +1,129 @@
+/* Teensyduino Core Library
```
profezzorn, line 1: So this looks like it was copied from the teensyduino core library.
Why?
Can't we just use the one that was in there?
If not, we need comments that explains why and what was changed.

common/stdout.h
```diff
@@ -2,7 +2,9 @@
 #define COMMON_STDOUT_H
 
 #include "monitoring.h"
-
+#ifdef TEENSYDUINO
+#include "_usb_rawhid.h"
```
profezzorn, line 6: Why is the filename starting with an _?

common/stdout.h
```diff
@@ -23,12 +25,38 @@ class ConsoleHelper : public Print {
       stdout_output != default_output;
   }
   size_t write(uint8_t b) override {
+#ifdef TEENSYDUINO
+#if defined(RAWHID_INTERFACE)	
+	size_t ret = 1;
+    if (stdout_output) {
+		ret = stdout_output->write(b);
+	} else {
+		SerialHID.write(b);
```
profezzorn, line 34: Why can't we just make SerialHID extend Print?
Alternatively; why can't we just wrap SerialHID in something that extends Print?

artandtechvof: That would be the preferred way of implementing it.
I tried to inherit both From Print() and Stream() but did not get it to work (my programming skills failed me to understand why it did not work)
The _usb_rawhid.h inherits from the Teensuiduino usb_rawhid_class RawHID, I tried to extend and managed to get it working the way it is implemented in this PR. The _usb_rawhid.h is my attempt to make the RawHID objetc comply with Stream() to be able to use it in PorffieOS as a serial output, hence my new object _usb_rawusb_class SerialHID

This works;
`class rawHIDAdapter {
public:
  static void begin() { /* nothing todo */ }
  // static bool Connected() { return !!WebUSBSerial; }
  static bool Connected() { return true; }
  static bool AlwaysConnected() { return true; }
  static Stream& stream() { return SerialHID; }
  static const char* response_header() { return "-+=BEGIN_OUTPUT=+-\n"; }
  static const char* response_footer() { return "-+=END_OUTPUT=+-\n"; }
};`

When I try to use it the way it should;
     stdout_output = &SA::stream();
or;
     stdout_output = &SerialHID; //reroute output to SerialHID
I get the following error;
   ...\ProffieOS.ino.cpp.o: In function `RawHIDParser::Loop()':
   ...\ProffieOS/ProffieOS.ino:1572: undefined reference to `SerialHID'
collect2.exe: error: ld returned 1 exit status

I am sure that this could be implemented different to make the SerialHID behave as a Stream* or Print* Although receiving data might prove difficult to get right, and still requires its own parser? 

So far I am quite happy that I got it to work, but realise that it is not worthy integrating into the main OS the way it is written right now. I would need some help to get this fixed, as my coding-skills are not enough to see why this is not working the way it should...

profezzorn: There are a few options for how to make this work....

1. Start with a copy of usb_seremu.{h,c} from the teensy library, then just modify that to talk to the rawhid endpoint. This would make it work exactly like a serial port, so it would be compatible with everything.
2. Don't copy any code, just write a new piece of code in ProffieOS that extends Stream and calls the rawhid recv/send functions to send/receive data. This new class could inherit from Looper to do the polling required. I think this might be slightly more complicated, because there may be some buffere management involved.

PR #176, opened 2021-04-26
Talkie for Auto, Stun, Kill

NoSloppy:
Addressing // TODO: would rather do a Talkie to speak the mode we're in after mode sound

### DIFF
--- a/props/blaster.h
+++ b/props/blaster.h
@@ -35,7 +35,7 @@ public:
     MODE_AUTO
   };
 
-  BlasterMode blaster_mode = MODE_KILL;
+  BlasterMode blaster_mode = MODE_STUN;
 
   virtual void SetBlasterMode(BlasterMode to_mode) {
     if (!auto_firing_) {
--- a/props/blaster.h
+++ b/props/blaster.h
@@ -10,6 +10,7 @@ EFFECT(clipout);
 EFFECT(empty);
 EFFECT(full);
 EFFECT(jam);
+EFFECT(mode);
 EFFECT(plioff);
 EFFECT(plion);
 EFFECT(range);
@@ -19,7 +20,9 @@ EFFECT(unjam);
 EFFECT(mdstun);
 EFFECT(mdkill);
 EFFECT(mdauto);
-
+// For mode sounds, specific "mdstun", "mdkill", and "mdauto" may be used.
+// If just a single "mode" sound for all switches exists, that will be used.
+// If no mode sounds exist in the font, a talkie version will speak the mode on switching.
 class Blaster : public PropBase {
 public:
   Blaster() : PropBase() {}
@@ -32,7 +35,7 @@ public:
     MODE_AUTO
   };
 
-  BlasterMode blaster_mode = MODE_STUN;
+  BlasterMode blaster_mode = MODE_KILL;
 
   virtual void SetBlasterMode(BlasterMode to_mode) {
     if (!auto_firing_) {
@@ -228,7 +231,7 @@ public:
   }
 
   // Make clash do nothing except unjam if jammed.
-  void Clash(bool stab) override {
+  void Clash(bool stab, float strength) override {
     if (is_jammed_) {
       is_jammed_ = false;
       SaberBase::DoEffect(EFFECT_UNJAM, 0);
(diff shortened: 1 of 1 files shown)


props/blaster.h
```diff
@@ -306,8 +306,19 @@ class Blaster : public PropBase {
 	  hybrid_font.PlayCommon(&SFX_mode);
 	  return;
 	}
-	// TODO: would rather do a Talkie to speak the mode we're in after mode sound
-	beeper.Beep(0.05, 2000.0);
+	if (blaster_mode == MODE_STUN) {
+	  talkie.Say(spSTUN); return;
```
profezzorn, line 310: Fonts should have the opportunity to override these effects.

props/blaster.h
```diff
@@ -306,8 +306,19 @@ class Blaster : public PropBase {
 	  hybrid_font.PlayCommon(&SFX_mode);
 	  return;
 	}
-	// TODO: would rather do a Talkie to speak the mode we're in after mode sound
-	beeper.Beep(0.05, 2000.0);
+	if (blaster_mode == MODE_STUN) {
```
profezzorn, line 309: use a switch()

props/blaster.h
NoSloppy: Are you suggesting we should check for a font version of the spoken words? 
Since there already is a stun.wav that is the blaster sound, something such as "spstun" or similar?

profezzorn: How about mdstun, mdauto, mdkill ?

NoSloppy: Changed to mdstun, mdauto, mdkill. Removed plain "mode" as these replace that with specific sounds for each mode, yes?

props/blaster.h
NoSloppy: Done.

props/blaster.h
```diff
@@ -301,22 +303,32 @@ class Blaster : public PropBase {
...
       case EFFECT_PLI_OFF: hybrid_font.PlayCommon(&SFX_plioff); return;
-	
+  
+    }
+  }
+
+  void SayMode() {
+    switch(blaster_mode) {
+      case MODE_STUN:
+      if (SFX_mdstun) hybrid_font.PlayCommon(&SFX_mdstun); return;
```
profezzorn, line 321: Did you test it?
(Hint; thjis doesn't work.)

NoSloppy: I was dumb and lazy and just commented out the font files checks to avoid a song and dance to delete the files.
That being said, it's corrected and tested properly now.
Why wouldn't it fall through to the talkie if no sounds were found? I was just trying to condense the if clause to one line for neatness/efficiency.

props/blaster.h
```diff
@@ -301,22 +307,44 @@ class Blaster : public PropBase {
...
       case EFFECT_UNJAM: hybrid_font.PlayCommon(&SFX_unjam); return;
       case EFFECT_PLI_ON: hybrid_font.PlayCommon(&SFX_plion); return;
       case EFFECT_PLI_OFF: hybrid_font.PlayCommon(&SFX_plioff); return;
-	
+  
+    }
+  }
+
+  void SayMode() {
+    switch(blaster_mode) {
```
profezzorn, line 323: space after switch

props/blaster.h
```diff
@@ -301,22 +307,44 @@ class Blaster : public PropBase {
...
+      case MODE_STUN:
+        if (SFX_mdstun) {
+          hybrid_font.PlayCommon(&SFX_mdstun);
+          return;
```
profezzorn, line 327: This return is redundant.

props/blaster.h
```diff
@@ -301,22 +307,44 @@ class Blaster : public PropBase {
...
+        } else {
+          talkie.Say(spSTUN);
+        }
+      return;
```
profezzorn, line 331: this should be "break", not "return"

props/blaster.h
```diff
@@ -301,22 +307,44 @@ class Blaster : public PropBase {
...
+      case MODE_KILL:
+        if (SFX_mdkill) {
+          hybrid_font.PlayCommon(&SFX_mdkill);
+          return;
```
profezzorn, line 335: redundant

props/blaster.h
```diff
@@ -301,22 +307,44 @@ class Blaster : public PropBase {
...
+        } else {
+          talkie.Say(spKILL);      
+        }
+      return;
```
profezzorn, line 339: break

props/blaster.h
```diff
@@ -301,22 +307,44 @@ class Blaster : public PropBase {
...
+      case MODE_AUTO:
+        if (SFX_mdauto) {
+          hybrid_font.PlayCommon(&SFX_mdauto);
+          return;
```
profezzorn, line 343: redundant

props/blaster.h
```diff
@@ -301,22 +307,44 @@ class Blaster : public PropBase {
...
+        } else {
+          talkie.Say(spAUTOFIRE);       
+        }
+      return;
```
profezzorn, line 347: break

props/blaster.h
```diff
@@ -10,13 +10,15 @@ EFFECT(clipout);
 EFFECT(empty);
 EFFECT(full);
 EFFECT(jam);
-EFFECT(mode);
```
profezzorn, line 13: I wonder if the "mode" sound would be preferable to the talkie effects?
At the very least, maybe should add a comment at the beginning of the file telling people that they can copy the mode sound to mdstun/mdkill/mdauto if they prefer one sound over the talkie effects.

NoSloppy: How about an else if to just use "mode" for all 3 as backup
check for mdkill
if not check for mode
if not play talkie version?

profezzorn: that would also work, and may be preferable.

props/blaster.h
```diff
@@ -30,7 +35,7 @@ class Blaster : public PropBase {
     MODE_AUTO
   };
 
-  BlasterMode blaster_mode = MODE_STUN;
+  BlasterMode blaster_mode = MODE_KILL;
```
profezzorn, line 38: This was not part of the PR description.

props/blaster.h
NoSloppy: This whole section fixed and tested good.

props/blaster.h
NoSloppy: Oops. Sorry. Ok do have it default start on kill?

profezzorn: Let's not change it unless there is a reason to change it.

PR #178, opened 2021-04-27
Add TrSparkX (or rename)

fdarosa2663:
Creating a variation on TrWaveX that does not fade, for use in SparkTip type Wipe transitions to create better SparkTip ignition/retraction versions as well as blast/clash effects where the Fade is not desired.  

The FADOUT_MS in TrWaveX affects certain uses even when set to higher value to negate fading because it keeps transition active for duration so TrSparkX bases transition on WAVE_MS (SPARK_MS) and does not fade effect over length of the blade.

Naming is flexible if TrSparkX isn't good.


transitions/wave.h
```diff
@@ -57,4 +57,51 @@ class TrWaveX : public TransitionBaseX<FADEOUT_MS> {
   }
 };
    
+template<class COLOR,
```
profezzorn, line 60: Needs description

fdarosa2663: added description

transitions/wave.h
```diff
@@ -56,5 +56,59 @@ class TrWaveX : public TransitionBaseX<FADEOUT_MS> {
...
+    spark_size_.run(blade);
+    spark_center_.run(blade);
+    spark_ms_.run(blade);
+    color_.run(blade);
+    if (this->restart()) {
+      center_ = spark_center_.getInteger(0);
+      size_ = spark_size_.getInteger(0);
+    }
+    TransitionBaseX<SPARK_MS>::run(blade);
+    mix_ = this->update(32768);
```
profezzorn, line 84: It seems like this will fade the mix *up* rather than down.

fdarosa2663: Ah, yup was watching for the fade out but slowing down it became visible, removed mix_, just setting 32768 fixed

transitions/wave.h
```diff
@@ -56,5 +56,57 @@ class TrWaveX : public TransitionBaseX<FADEOUT_MS> {
...
+  SPARK_MS spark_ms_;
+  COLOR color_;
+public:
+  template<class A, class B>
+  auto getColor(const A& a, const B& b, int led) -> decltype(MixColors(a,color_.getColor(led),1,1)) {
+    int dist = std::abs(center_ - led * 32768 / num_leds_);
+    int N = std::abs(dist - offset_) * size_ >> 15;
+    int mix;
+    if (N <= 32) {
+      mix = blast_hump[N] * 32768 >> 8;
```
profezzorn, line 105: * 32768 >> 8   can be replaced   with << 7

transitions/wave.h
```diff
@@ -56,5 +56,57 @@ class TrWaveX : public TransitionBaseX<FADEOUT_MS> {
...
+    spark_center_.run(blade);
+    spark_ms_.run(blade);
+    color_.run(blade);
+    if (this->restart()) {
+      center_ = spark_center_.getInteger(0);
+      size_ = spark_size_.getInteger(0);
+    }
+    TransitionBaseX<SPARK_MS>::run(blade);
+    num_leds_ = blade->num_leds();
+    offset_ = (millis() - this->start_millis()) * 32768 / spark_ms_.getInteger(0);
```
profezzorn, line 85: I think we can replace this with offset_ = update(32768);

transitions/wave.h
```diff
@@ -56,5 +56,57 @@ class TrWaveX : public TransitionBaseX<FADEOUT_MS> {
...
+private:
+  SPARK_CENTER spark_center_;
+  int center_;
+  
+  SPARK_SIZE spark_size_;
+  int size_;
+
+  int num_leds_;
+  int offset_;
+  SPARK_MS spark_ms_;
```
profezzorn, line 96: we don't need spark_ms_

transitions/wave.h
fdarosa2663: Ah, thanks.

transitions/wave.h
fdarosa2663: updated

transitions/wave.h
fdarosa2663: removed

PR #185, opened 2021-05-02
Fix TrWipeSparkTip, TrWipeInSparkTip

fdarosa2663:
Replace TrWaveX with TrSparkX in alias to fix fading on "SparkTip" ignitions, template remains the same so it's backwards compatible.


profezzorn:
Backwards compatible means that it also WORKS the same.
Altering how an existing alias or template works should be done very carefully.
Are you *sure* this is what *everybody* wants?

fdarosa2663:
Yes, I've had many users note that if you try to use the Spark versions over 500 ms that the spark disappears too quickly since we introduced in previous OS.  The Spark versions have not worked correctly with Wave since their introduction because of the FADE but there wasn't another option, hence the creation of the TrSpark.

PR #188, opened 2021-05-03
Fix compile error.

NoSloppy:
DoStab() needed a strength value.
Also added no Spoken battery level if in Volume menu.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -706,11 +706,13 @@ public:
     // 2 and 3 button
     case EVENTID(BUTTON_AUX, EVENT_FIRST_HELD_LONG, MODE_OFF):
   #endif 
-    talkie.SayDigit((int)floorf(battery_monitor.battery()));
-    talkie.Say(spPOINT);
-    talkie.SayDigit(((int)floorf(battery_monitor.battery() * 10)) % 10);
-    talkie.SayDigit(((int)floorf(battery_monitor.battery() * 100)) % 10);
-    talkie.Say(spVOLTS);
+    if (!mode_volume_) {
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -706,11 +706,13 @@ class SaberBCButtons : public PropBase {
     // 2 and 3 button
     case EVENTID(BUTTON_AUX, EVENT_FIRST_HELD_LONG, MODE_OFF):
   #endif 
-    talkie.SayDigit((int)floorf(battery_monitor.battery()));
-    talkie.Say(spPOINT);
-    talkie.SayDigit(((int)floorf(battery_monitor.battery() * 10)) % 10);
-    talkie.SayDigit(((int)floorf(battery_monitor.battery() * 100)) % 10);
-    talkie.Say(spVOLTS);
+    if (!mode_volume_) {
```
profezzorn, line 709: Not sure if this was meant to be here, it doesn't seem related to the compile error below.

NoSloppy: Probably my fault. I *did* include the change in the pull request comment as I thought that's what you'd asked for?
Would you prefer PRs to be topically separate, and not combine multiple fixes all in one?
Just trying to avoid bothering you so much with small fixes. 
If that's the case, no problem just let me know.
For example, I went through and cleaned all trailing spaces from lines...should that even be an update worth PRing?

profezzorn: Generally topically separate is better.
However, including it in the description should be enough.
I tend to skip very quickly to the actual changes, so I missed reading the description in this case. (I only read the title)

PR #194, opened 2021-05-16
Fix compile issue

NoSloppy:
I don't think this was right.

### DIFF
--- a/props/saber_shtok_buttons.h
+++ b/props/saber_shtok_buttons.h
@@ -144,7 +144,7 @@ SaberShtokButtons() : PropBase() {}
 // Multi-Blaster Deflection mode
       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON | BUTTON_POWER):
         swing_blast_ = true;
-        hybrid_font.SB_Blast();
+        hybrid_font.DoBlast();
         return true;
 
       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON):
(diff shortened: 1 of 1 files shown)


props/saber_shtok_buttons.h
```diff
@@ -144,7 +144,7 @@ SaberShtokButtons() : PropBase() {}
 // Multi-Blaster Deflection mode
       case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON | BUTTON_POWER):
         swing_blast_ = true;
-        hybrid_font.SB_Blast();
+        hybrid_font.DoBlast();
```
profezzorn, line 147: This is not right.
Either we call SaberBase::DoBlast() to notify all saber bases of a blast, or we call hybrid_font.SB_Effect(EFFECT_BLAST, (200 + random(700))/1000.0f)); to only play the blast sound and nothing else.

NoSloppy: It's Dmitry's prop. I just came across it helping someone out. It's SaberBase::DoBlast(); according the general consensus in the other props, so I'll do that...for him?

profezzorn: It's better than not compiling...

NoSloppy: Well, turns out it should have been hybrid_font.SB_Effect(EFFECT_BLAST, 0);
When entering the multiblast mode, it just plays the sound.
He has his own pull request in for updates to this file anyway and it's correct in that version, so.....

PR #195, opened 2021-05-18
Add clash_impact.h

fdarosa2663:


functions/clash_impact.h
```diff
@@ -0,0 +1,35 @@
...
+    return value_;
+  }
+
+private:
+  MIN min_;
+  MAX max_;
+  int value_;
+};
+
+template<int MIN = 2, int MAX = 14>
```
profezzorn, line 32: min and max really should be floats, but floats don't work with templates...
I think we just need to multiply them by 100.
So if you want MIN to be 2.5, you set it to 250.

We could maybe call it min_cents/max_cents to make it clear that it is counted in hundredths.

functions/clash_impact.h
```diff
@@ -0,0 +1,35 @@
...
+using ClashImpact = ClashImpactX<Int<MIN>, Int<MAX>>;
```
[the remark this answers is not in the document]
fdarosa2663: changed

functions/clash_impact.h
fdarosa2663: updated

functions/clash_impact.h
```diff
@@ -0,0 +1,35 @@
...
+// returned value: INTEGER
+
+template<class MIN, class MAX>
+class ClashImpactF {
+public:
+  void run(BladeBase* base) {
+    min_cents_.run(base);
+    max_cents_.run(base);
+    float i = SaberBase::GetClashStrength();
+    value_ = clampi32(((i - (min_cents_.getInteger(0) / 100)) / (max_cents_.getInteger(0) / 100)) * 32768, 0, 32768);
```
profezzorn, line 19: By dividing with 100, we are throwing away the decimals.
Also, we should be dividing by max - min, not just the maximum.

I think what we want is this:

int min = min_cents_.getInteger();
int max = max_cents_.getInteger();
value_ = clampi32((i - min) * 32768 / (max - min), 0, 32768);

fdarosa2663: I've updated, had to modify though as min and max were 100 times larger than i

functions/clash_impact.h
```diff
@@ -0,0 +1,37 @@
...
+#include "../common/saber_base.h"
+
+// Usage: ClashImpact<MIN, MAX>
+// MIN is minimum value Clash is detected
+// MAX is maximum impact to return 32768
+// Returns 0-32768 based on impact strength of clash
+// returned value: INTEGER
+
+template<class MIN, class MAX>
+class ClashImpactF {
```
profezzorn, line 13: Actually, this should be called ClashImpactFX, and ClashImpact at the bottom should be called ClashImpactF<>

functions/clash_impact.h
```diff
@@ -0,0 +1,37 @@
...
+public:
+  void run(BladeBase* base) {
+    min_cents_.run(base);
+    max_cents_.run(base);
+    float i = SaberBase::GetClashStrength();
+    float min = min_cents_.getInteger(0) / 100;
```
profezzorn, line 19: This doesn't work. We're still dividing two integers and THEN converting the result to a float.
Also, since we're multiplying by min and dividing my max-min below, if both numbers are 100 times bigger, they will cancel out, so there is no need to divide by 100.

fdarosa2663: i is between 2.5 ~ 16, we made min_cents_ and max_ cents_ * 100 to allow for say 3.5 as a min or 14.2 as a max by using Int<350> or Int<1420> respectively.  So when I subtract min from i without dividing min by 100 it is always negative so there is no effect when using the function from here:

value_ = clampi32((i - min)  

This results in the Function always returning essentially 0 and not working as intended.  I switched and am now multiplying i * 100 instead of dividing min and max by 100 and it works.  Dividing by 100 was also working as expected though since the min and max were intentionally multiplied by 100 in the template to allow users to get the decimal values in using Int<>.  

Also, had to remove the subtraction of min from max for the range, otherwise the top end is lost and (max -  min) creates 32768 at lower than actual max value.  We're subtracting min from i to create a floor for 0 in the function, otherwise the lightest clash is not returning near zero as intended because of actual CLASH_THRESHOLD_G.  The min is to shift the clash strength down to CLASH_THRESHOLD_G essentially with the threshold being zero, max is meant as the range you want your hardest hit to return 32768 which should be 16, but we're multiplying by 100 so it becomes 1600, subtracting min from this caused the range to not be full value so I removed the subtraction. 

With the function we want the MIN to be the lowest clash you want detected being = 0 and the MAX to be the highest clash you want detected to be = 32768 since the clash strength range is actually CLASH_THRESHOLD_G to @16, I used 200 and 1600 (2.0 and 16.0) as the range by default but users can adjust if needed.

I added comments in top section to clarify for users.

functions/clash_impact.h
fdarosa2663: updated

PR #197, opened 2021-05-28
Fett263 Prop for OS6

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -285,75 +442,1428 @@ EFFECT(faston); // for EFFECT_FAST_ON
...
+
+  FormEntry clash_rec[32];
+ 
+};
+#endif
+
+#ifdef FETT263_EDIT_MODE_MENU
+// Edit Length
+
+class NewLength {
```
profezzorn, line 588: What is this claass for, please add a comment.

props/saber_fett263_buttons.h
```diff
@@ -285,75 +442,1428 @@ EFFECT(faston); // for EFFECT_FAST_ON
...
+  }
+  static int GetNewLength() {
+    return length_;
+  }
+};
+
+int NewLength::length_ = 0;
+
+template<class BASE, class LIGHTUP, class BLACK = BLACK>
+class LengthEdit : protected NewLength {
```
profezzorn, line 603: I'm not sure I understand the point of inheriting a class with only static members.

props/saber_fett263_buttons.h
```diff
@@ -285,75 +442,1428 @@ EFFECT(faston); // for EFFECT_FAST_ON
...
+        }
+      }
+    }
+    if (say_type_ != SAY_NONE && !wav_player->isPlaying()) {
+      switch (say_type_) {
+        case SAY_BATTERY:
+          switch (play_number_) {
+            case PLAY_FIRST:
+              if (say1_ == 0) {
+                PlayMenuSound("mzero.wav");
```
profezzorn, line 1163: Why isn't zero a part of SFX_mnum?

props/saber_fett263_buttons.h
```diff
@@ -285,75 +442,1428 @@ EFFECT(faston); // for EFFECT_FAST_ON
...
+        } else {
+          SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
+          SaberBase::DoBeginLockup();
+        }
+#endif
+        }
+      }
+    }
+    if (say_type_ != SAY_NONE && !wav_player->isPlaying()) {
+      switch (say_type_) {
```
profezzorn, line 1158: This state machine is pretty hard to read.
I think it would be better to split it out into a separate class that inherits looper and StateMachine, then we can just do:

   while (wav_player && wav_player->isPlaying()) YIELD();

to wait until the sound is done playing.

props/saber_fett263_buttons.h
```diff
@@ -285,75 +442,1428 @@ EFFECT(faston); // for EFFECT_FAST_ON
...
+      off_event_ = false;
+      restart_ = true;
+      last_rotate_ = millis();
+    }
+    if (restart_ && millis() - last_rotate_ > calc_ + 1000) {
+      restart_ = false;
+      FastOn();
+    }
+#endif
+    if (announce_mode_ && !wav_player->isPlaying()) {
```
profezzorn, line 1337: also check if wav_player is not set

profezzorn (commented):
This review is going to take a very very very long time.
There is so many things going on here.
I've scanned though the file, but it's going to have to read it many times just to understand what's going on with all of it, and there will be many, many, many, many comments.

Here is the first batch.

PR #205, opened 2021-06-04
Add idle.bmp loop

NoSloppy:
Has quirky behavior I can't iron out.
SCREEN_MESSAGE seems to ignore ProffieOSFontImageDuration when idle.bmp is playing.

### DIFF
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -135,7 +135,7 @@ public:
       case SCREEN_PLI:
 	Clear();
         if (!just_booted_) {
-          if (&IMG_idle) {
+          if (IMG_idle) {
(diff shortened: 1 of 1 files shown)


NoSloppy (commented):
Line 147 is what doesn't seem to work. 
I have changed the return to 10000 and still the message only appears for a blip and then idle.bmp plays.
I can't see any reason why the message doesn't stay visible for the time specified in the return.
font.bmp files show for the time set in ProffieOSFontImageDuration no problem.

NoSloppy:
Closing due to modding the new improved ssd1306.h instead, where this works....mostly.

PR #207, opened 2021-06-14
Fixed compile issue re: NO_BLADE_NO_GEST_ONOFF and usage edits.

NoSloppy:
Someone ate my PR....anyway,
In case NO_BLADE_NO_GEST_ONOFF is used without BLADE_DETECT_PIN defined.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -315,6 +315,10 @@ Turn off blade -        Hold POW and wait until blade is off,
 #define THRUST_GESTURE
 #endif
 
+#if defined(NO_BLADE_NO_GEST_ONOFF) && !defined(BLADE_DETECT_PIN)
+#error Using NO_BLADE_NO_GEST_ONOFF requires a BLADE_DETECT_PIN to be defined 
+#endif
+
 EFFECT(dim);        // for EFFECT_POWERSAVE
 EFFECT(battery);    // for EFFECT_BATTERY_LEVEL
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -1039,6 +1040,7 @@ class SaberBCButtons : public virtual PropBase {
   bool battle_mode_ = false;
   bool max_vol_reached_ = false;
   bool min_vol_reached_ = false;
+  bool blade_detected_ = true;
```
profezzorn, line 1043: This creates two variables with the same name which is almost never a good idea.
Unless you have a terribly good reason for this, please solve it some other way.

NoSloppy:
Do you think it's okay that I leave it up to the end user based on the
instructions up top that a blade detect pin needs to be defined to use it?
I'm fine with that. I just happened to compile with only one of them
defined and got errors and was trying to keep it foolproof. Now I see where
it could be a problem this way. I went this route to avoid adding a whole
bunch of #ifndef blade detect pin lines in the middle of all the code
so.... Just let it go?

profezzorn:
I'm thinking the easy thing to do would be something like:

#if defined(NO_BLADE_NO_GEST_ONOFF) && !defined(BLADE_DETECT_PIN)
#error You need blade detect (BLADE_DETECT_PIN) to use
NO_BLADE_NO_GEST_ONOFF
#endif

NoSloppy:
Oh ok that's great, thank you. Now I know.

PR #209, opened 2021-06-16
Add style size printout?

NoSloppy:
Would this be helpful to just include officially?
Handy when managing memory and info is always good in my opinion.


NoSloppy:
This breaks parts of WebUSB.

PR #215, opened 2021-07-05
saber_BC_buttons Match preon/in pstoff/out files.

NoSloppy:
Tested and working well.


profezzorn:
I'm not a fan of this change, because it will make fonts work differently for your prop than it does for other props, basically fracturing font making. (Some fonts will need your prop in order to work as intended, which isn't great for the community.)

I think this should be implemented globally and have a config file option for turning it on/off instead.

NoSloppy:
> I think this should be implemented globally and have a config file option for turning it on/off instead.
Agreed. I just wasn't sure if you'd want it globally.
Cool. New PR submitted.

PR #216, opened 2021-07-06
prop_base.h - matching preons/outs and pstoffs/ins

NoSloppy:
Not sure if chosen spot in PropBase class is an issue, but tested working.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -110,6 +110,38 @@ public:
     return SaberBase::IsOn() || on_pending_;
   }
 
+// Match preon.wav to out.wav
+  virtual void SelectIgnitionPair() {
+    if (!SFX_preon.files_found()) return;
+
+    int preonCount = SFX_preon.files_found();
+    int outCount = SFX_out.files_found();
+    int pairIgnition;
+
+    // If we don't have the same number of preon/out files, then don't bother getting a matched pair.
+    if (preonCount == outCount) {
+        pairIgnition = rand() % preonCount;
+        SFX_preon.Select(pairIgnition);
+        SFX_out.Select(pairIgnition);
+    }
+  }
+
+// Match pstoff.wav to in.wav
+  virtual void SelectRetractionPair() {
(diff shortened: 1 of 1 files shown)


profezzorn:
Actually, I think I want to take a look at rewriting how preon works a bit.
Depending on that turns out, I will decide if this is the right way to do it or not.

NoSloppy:
Oh ok. Curious how else they could work.

NoSloppy:
With Fett263's "dual mode" this could interfere. How about I just move the matching number preon and out files back into my prop?

props/prop_base.h
```diff
@@ -110,6 +110,38 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+    }
+  }
+
+// Match pstoff.wav to in.wav
+  virtual void SelectRetractionPair() {
+    if (!SFX_pstoff.files_found()) return;
+
+    int pstoffCount = SFX_pstoff.files_found();
+    int inCount = SFX_in.files_found();
+    int pairIgnition;
```
profezzorn, line 135: This should be combined with line 139.

props/prop_base.h
```diff
@@ -110,6 +110,38 @@ class PropBase : CommandParser, Looper, protected SaberBase {
     return SaberBase::IsOn() || on_pending_;
   }
 
+// Match preon.wav to out.wav
+  virtual void SelectIgnitionPair() {
+    if (!SFX_preon.files_found()) return;
+
+    int preonCount = SFX_preon.files_found();
+    int outCount = SFX_out.files_found();
+    int pairIgnition;
```
profezzorn, line 119: should be combined with line 123

props/prop_base.h
NoSloppy: Both cleaned up. Also moved pstoff above Off().

props/prop_base.h
NoSloppy: Done.

PR #218, opened 2021-07-13
blaster.h had typo - shorts_fired

NoSloppy:


profezzorn:
But what if I have a gun that fires shorts?

PR #224, opened 2021-07-25
Sound library updates/additions/improvements

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -687,103 +689,20 @@ class SoundLibrary {
...
   void SayTwistIgnition() { Play("mtwiston.wav"); }
   void SayTwistRetraction() { Play("mtwstoff.wav"); }
-  void SayUp() { Play("mup.wav"); }
+  void SayUp() { Play("mup.wav"); } // Sound for increase
   void SayVolumeMenuEnd() { Play("vmend.wav"); }
-  void SayVolumeUp() { Play("volup.wav"); }
-  void SayZoom() { Play("mzoom.wav"); } // Zoom in for color menu "zooming in"
-  void SayVolumeDown() { Play("voldown.wav"); }
+  void SayVolumeUp() { Play("volup.wav"); } // Sound for increasing volume
+  void SayZoom() { Play("mzoom.wav"); } // Sound for color menu "zooming in"
```
profezzorn, line 698: If that's what it says, we should probably rename this function to SayZoomingIn()
The intention is that the function name should be as close to the actual sound as possible.

props/saber_fett263_buttons.h
```diff
@@ -687,103 +689,20 @@ class SoundLibrary {
...
+  void SayVolumeDown() { Play("voldown.wav"); } // Sound for decreasing volume
   void SayGestureMenu() { Play("mgestsub.wav"); }
   void SaySettingsMenu() { Play("msetsub.wav"); }
   void SayStyleSettings() { Play("stylstm.wav"); }
 
   void SayColor(int n) {
```
profezzorn, line 704: replace int with "SayColorName" (hopefully it won't lead to a bunch of errors)

props/saber_fett263_buttons.h
```diff
@@ -466,6 +466,37 @@ enum SayType {
   SAY_MILLIS,
 };
 
+EFFECT(clrlst); // spoken color names for SAY_COLOR_LIST
+enum ColorNumber {
```
profezzorn, line 470: Add a comment specifying that new colors should be added to the end and that the numbers should never change.

props/saber_fett263_buttons.h
```diff
@@ -973,34 +892,36 @@ int IntEdit::int_edit_ = 0;
...
+  { { 46260, 33410, 0 }, COLOR_GOLD }, // Gold
+  { Yellow::color(), COLOR_YELLOW },
+  { GreenYellow::color(), COLOR_GREENYELLOW },
+  { Green::color(), COLOR_GREEN },
+  { Aquamarine::color(), COLOR_AQUAMARINE },
+  { Cyan::color(), COLOR_CYAN },
+  { DeepSkyBlue::color(), COLOR_DEEPSKYBLUE },
+  { DodgerBlue::color(), COLOR_DODGERBLUE },
+  { Blue::color(), COLOR_BLUE },
+  { { 7710, 15420, 51400 }, COLOR_ICEBLUE }, // IceBlue
```
profezzorn, line 911: The comments aren't actually helpful anymore since the COLOR_ICEBLUE actually says what the color is.

props/saber_fett263_buttons.h
```diff
@@ -2627,8 +2548,15 @@ SaberFett263Buttons() : PropBase() {}
       style_parser.GetArgument(current_preset_.GetStyle(blade_num_), set_num_ + 2, argspace);
       calc_ = strtol(argspace, NULL, 0);
       IntEdit::SetIntValue(calc_);
-      sound_library_.SaySelect();
       if (NUM_BLADES == 1) blade_num_ = 1;
+      switch (set_num_) {
```
profezzorn, line 2552: Add a comment explaining why we only use SaySelect() some of the time.

props/saber_fett263_buttons.h
```diff
@@ -2993,9 +2916,9 @@ SaberFett263Buttons() : PropBase() {}
           dial_ += direction;
           if (dial_ < 0) dial_ = NELEM(color_list_) - 1;
           if (dial_ > (int)NELEM(color_list_) - 1) dial_ = 0;
-          ShowColorStyle::SetColor(Color16(color_list_[dial_]));
+          ShowColorStyle::SetColor(Color16(color_list_[dial_].color));
 #ifdef FETT263_SAY_COLOR_LIST
-          sound_library_.SayColor(dial_);
+          sound_library_.SayColor(color_list_[dial_].color_number - 1);
```
profezzorn, line 2921: The -1 should be inside SayColor() I think.

props/saber_fett263_buttons.h
fdarosa2663: SayColorName was removed, we're using SFX_clrlst.Select() instead.

props/saber_fett263_buttons.h
fdarosa2663: updated

props/saber_fett263_buttons.h
fdarosa2663: Comment added

props/saber_fett263_buttons.h
fdarosa2663: removed

props/saber_fett263_buttons.h
fdarosa2663: added

props/saber_fett263_buttons.h
fdarosa2663: updated

props/saber_fett263_buttons.h
profezzorn: Right, it's called ColorName now.
Replace "int" with "ColorName".

fdarosa2663: ok, updated

PR #225, opened 2021-07-25
Fix CC_COLOR_LIST

fdarosa2663:
Use ShowColorAllBladeTemplate<> for CC_COLOR_LIST and remove RotateColorsX<Variation,XXX> from preview styles to accurately show preview colors.


profezzorn:
I find this hard to review.
There seem to be a bunch of changes thrown in that fixes something, but I don't know what it is and the description doesn't really help.

Could you update the description of this PR to include all the things that is being changed or fixed here?

props/saber_fett263_buttons.h
```diff
@@ -1418,8 +1425,12 @@ SaberFett263Buttons() : PropBase() {}
     itoa(Color16(color_source).g, new_color + strlen(new_color), 10);
     strcat(new_color, ",");
     itoa(Color16(color_source).b, new_color + strlen(new_color), 10);
-    current_preset_.SetStyle(blade,style_parser.SetArgument(current_preset_.GetStyle(blade), effect + 2, new_color));
-    if (color_mode_ != CC_COLOR_LIST) color_mode_ = NONE;
+    if (color_mode_ == CC_COLOR_LIST) {
+      ONCEPERBLADE(NEW_COLOR_ALL_BLADES);
```
profezzorn, line 1429: There is really no need to use ONCEPERBLADE here. Since your macro doesn't use ## to concatenate N with anything you can just use a loop from 1 to NUM_BLADES instead.

fdarosa2663:
The bulk is all tied to fixing CC_COLOR_LIST implementation and making colors appear consistently by removing RotateColorsX<> in preview templates:  
1. Switching from UpdateStyle() on each dial turn to using ShowColorAllBladeTemplate<> to preview the colors. Had to change entering, dial, exiting and reverting for CC_COLOR_LIST to use show_color_all_ and NewColor()
2. Line 1170~ Removed "RotateColorsX<Variation," from the previews because Variation should not be used with RgbArg or colors could be displayed differently from Color List which could become confusing.  CC_COLOR_LIST removes ColorWheel from RgbArg styles so the Variation is not applicable to those styles anymore anyway.
3. Added a separate define for SayColor specific to CC_COLOR_LIST.

Others were minor fixes for errors found during testing:
4.  Line 3630 - Fixed incorrect MenuPrompt for the Style Settings SubMenu.
5.  Line 2769 - Fixed bug in MenuDial for Blade selection if first turn was -1
6.  Lines 4109, 4117 added missing return true (was triggering Force Effect on button event)

props/saber_fett263_buttons.h
fdarosa2663: replaced with while()

props/saber_fett263_buttons.h
```diff
@@ -4290,8 +4309,12 @@ SaberFett263Buttons() : PropBase() {}
             handles_color_change |= current_config->blade##N->current_style() && current_config->blade##N->current_style()->IsHandled(HANDLED_FEATURE_CHANGE_TICKED);
             ONCEPERBLADE(USES_COLOR_CHANGE)
             if (!handles_color_change) {
-              current_preset_.Save();
               color_mode_ = CC_COLOR_LIST;
+              show_color_all_.Start();
+              for (int i = 1; i <= NUM_BLADES; i++) {
+                ShowColorStyle::SetColor(GetColorArg(i, BASE_COLOR_ARG));
+                if (style_parser.UsesArgument(current_preset_.GetStyle(i), BASE_COLOR_ARG + 2)) break;
```
profezzorn, line 4316: This seems weird to me.
It breaks as soon as it finds a style that doesn't use the given argument.
Since the blades could be in any order, it seems odd to have one style change how other styles work.

fdarosa2663: Yes, I didn't think we could have multiple colors in ShowColorAllBladesTemplate<> so I have it use the first color argument it finds.  Is that not correct?  How would we have the blades individually show their current color is they are all using show_color_all_?

profezzorn: Hmm, I think the code is probably correct, just confusing.
I think it should probably look like this:

for (....) {
  if (  ... UsesArgument( )) {
     ShowColorStyle::SetColor(....)
     break;
  }
}

fdarosa2663: ok, updated, thanks.

fdarosa2663:
Had to move GetColorArg() function outside the EDIT_MODE #ifdef since it's used in CC_COLOR_LIST to prevent compile error if EDIT_MODE_MENU not used.

PR #227, opened 2021-07-29
Edit Length fix

fdarosa2663:
enabled blade length preview for blades < 10 pixels
fixed bug for blade_length > max_length_ -or- < 1


fdarosa2663:
added a small fix for TrackPlayer, there is no track_num_ 0, so fixed to prevent.

profezzorn:
Would have been nice if it was split into two CLs...

PR #229, opened 2021-08-01
fix saber_BC_buttons - factor in CONFIG_STARTUP_DELAY for SWING_ON delay

NoSloppy:
because motion chip causing EVENT: swing at delayed boot.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -265,6 +265,10 @@ Turn off blade -        Hold POW and wait until blade is off,
 #undef PROP_TYPE
 #define PROP_TYPE SaberBCButtons
 
+#ifndef CONFIG_STARTUP_DELAY
+#define CONFIG_STARTUP_DELAY 0
+#endif
+
 #ifndef MOTION_TIMEOUT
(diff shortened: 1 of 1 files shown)


NoSloppy:
Hmmm. not working... posted on the Crucible.

props/saber_BC_buttons.h
```diff
@@ -265,6 +265,10 @@ Turn off blade -        Hold POW and wait until blade is off,
 #undef PROP_TYPE
 #define PROP_TYPE SaberBCButtons
 
+#ifndef CONFIG_STARTUP_DELAY
+#define CONFIG_STARTUP_DELAY 0
```
profezzorn, line 269: This is not actually how CONFIG_STARTUP_DELAY works, and doing it this way could cause a problem.
If CONFIG_STARTUP_DELAY isn't defined, it means that it will use a default from the arduino-proffieboard plugin, not zero.

NoSloppy: I see. Well omission is an easy edit if that's the correct move here?

profezzorn:
This should be fixed now.

NoSloppy:
> This should be fixed now.

Thank you.

PR #231, opened 2021-08-02
Update IntArg enum for additional options

fdarosa2663:
per PM on theCrucible, expanding the enum options for IntArg


fdarosa2663:
Re-ordering enum will submit new update

PR #242, opened 2021-09-04
Add batterylevel, percentage, and female numbers to voice_data

NoSloppy:
I had a few requests for % being spoken in lieu of voltage.
As part of the update, both talkie genders get "percent" and "The battery level is", and female talkie now has the rest of the numbers as well.


common/battery_monitor.h
```diff
@@ -97,6 +97,7 @@ BatteryMonitor() : reader_(batteryLevelPin,
       float v = battery();
       STDOUT.println(v);
 #ifdef ENABLE_AUDIO
+      talkie.Say(spBATTERYLEVEL);
```
profezzorn, line 100: Some of these changes will use up some memory for everybody, and it has very little utility, so I don't think it's a good idea.

Maybe change this PR to only have the changes in your prop file and then I can merge it.

NoSloppy: Sure. I just edited the file back to original state instead of rebasing etc....

sound/voice_data.h
```diff
@@ -28,14 +28,34 @@ const uint8_t spSIX[] PROGMEM = {0x0E,0xD8,0xAE,0xDD,0x03,0x0E,0x38,0xA6,0xD2,0x
...
 const uint8_t spPOINT[] PROGMEM = {0xa0,0xfe,0xa1,0xc3,0xd5,0x25,0x4a,0xc0,0xba,0xad,0x30,0x93,0xc8,0xc9,0x09,0xb6,0x45,0x5c,0xec,0x14,0x27,0x98,0x63,0xc8,0xa6,0xda,0xdc,0x68,0x4e,0x20,0x9a,0xec,0xf0,0x53,0x5c,0xc5,0x4e,0xca,0x2b,0x4c,0x21,0x55,0xd3,0x94,0xac,0x38,0x5b,0x0d,0x2b,0x95,0x35,0xe2,0xa6,0xc9,0xbd,0xdd,0x56,0x89,0x0b,0xc7,0x9c,0x2d,0xd9,0x2e,0xd6,0xd2,0x77,0xda,0x64,0xbb,0x20,0x63,0xf3,0x68,0x4f,0x6d,0x80,0x25,0x5d,0x05,0x70,0x62,0x18,0x02,0x56,0x4a,0x03,0x00,0x80,0x07};
-const uint8_t spGOOD[] PROGMEM = {0x0A,0x28,0xCD,0x34,0x20,0xD9,0x1A,0x45,0x74,0xE4,0x66,0x24,0xAD,0xBA,0xB1,0x8C,0x9B,0x91,0xA5,0x64,0xE6,0x98,0x21,0x16,0x0B,0x96,0x9B,0x4C,0xE5,0xFF,0x01};
```
profezzorn, line 32: Why is this removing GOOD and MORNING?

NoSloppy: Well, I had just added them in as a personal perk, but stopped using them. I suppose they could stay since they don't take up any memory if unused and are then there if somebody ever wanted them?

NoSloppy: Now I remember. There's no Male voice equivalent, so I thought it best to just remove.

PR #243, opened 2021-09-04
add NO_REPEAT_RANDOM option to SelectFloat()

fdarosa2663:
Open to improvements but this implementation is working to randomize the up/down selection if SelectFloat() tries to repeat file as a start point.


sound/effect.h
```diff
@@ -231,9 +231,31 @@ class Effect {
     selected_++;
     if (selected_ == (int)files_found()) selected_ = 0;
   }
+#ifdef NO_REPEAT_RANDOM
+  int last_selected_;
```
profezzorn, line 235: No, we are not going to add a new variable, we're going to fix how last_ works instead.

sound/effect.h
```diff
@@ -231,9 +231,31 @@ class Effect {
...
+#endif
   void SelectFloat(float value) {
     int f = files_found();
     selected_ = clamp(floorf(f * value), 0, f - 1);
+#ifdef NO_REPEAT_RANDOM
+    if (selected_ == last_selected_) {
+      if (f == 2) {
+        selected_ = rand() % num_files_;
```
profezzorn, line 243: just se selected_ to -1 and let RandomFile do the picking.

sound/effect.h
```diff
@@ -231,9 +231,31 @@ class Effect {
...
+      } else {
+        if (selected_ == (int)files_found()) {
```
profezzorn, line 245: This is never going to be true, since selected_ is in the range 0 ... f - 1

sound/effect.h
```diff
@@ -231,9 +231,31 @@ class Effect {
...
+          selected_ -= 1;
```
profezzorn, line 246: We could get in trouble here if there is only one file.

sound/effect.h
```diff
@@ -231,9 +231,31 @@ class Effect {
...
+        } else {
+          int n = rand() % 2;
+          if (selected_ >= 1 && n == 2) {
```
profezzorn, line 249: rand() % 2 will never be equal to 2

sound/effect.h
```diff
@@ -231,9 +231,31 @@ class Effect {
...
     selected_ = clamp(floorf(f * value), 0, f - 1);
+#ifdef NO_REPEAT_RANDOM
+    if (selected_ == last_selected_) {
+      if (f == 2) {
+        selected_ = rand() % num_files_;
+      } else {
+        if (selected_ == (int)files_found()) {
+          selected_ -= 1;
+        } else {
+          int n = rand() % 2;
```
profezzorn, line 248: no need for a variable here, just put rand() % 2 in the next expression.

sound/effect.h
fdarosa2663: no problem, will need last_ fixed as it rewrites between Select currently.

sound/effect.h
fdarosa2663: done

sound/effect.h
fdarosa2663: my mistake, fixed

sound/effect.h
fdarosa2663: added "f > 1" to upper if statement

sound/effect.h
fdarosa2663: fixed

sound/effect.h
fdarosa2663: fixed

sound/effect.h
profezzorn: so fix it?

profezzorn:
I implemented this myself a bit differently.

PR #244, opened 2021-09-04
Add Accent Swing selection using SelectFloat()

fdarosa2663:
Open to improvement, this is the selection of swng/swing files based on swing acceleration using SelectFloat().


props/saber_fett263_buttons.h
```diff
@@ -3898,6 +3898,14 @@ SaberFett263Buttons() : PropBase() {}
       LOCK_SD(false);
       return true;
     }
+    if (!strcmp(cmd, "get_gesture") && arg) {
```
profezzorn, line 3901: This is from my code, not sure how it got into your PR, but that needs to be fixed before merging.

sound/hybrid_font.h
```diff
@@ -231,13 +240,24 @@ class HybridFont : public SaberBase {
         }
         if (!swing_player_) {
           if (!swinging_) {
-            if (rss > slashThreshold && SFX_slsh) {
-              swing_player_ = PlayPolyphonic(&SFX_slsh);
-            } else if (SFX_swng) {
-              swing_player_ = PlayPolyphonic(&SFX_swng);
+            if (font_config.ProffieOSMaxSwingAcceleration > font_config.ProffieOSMinSwingAcceleration) {
+              float s = (rss - font_config.ProffieOSMinSwingAcceleration) / font_config.ProffieOSMaxSwingAcceleration;
+              if (SFX_swng) {
```
profezzorn, line 245: I think we should reorganize the code a little so that it first picks the right Effect, then assigns that to an Effect* variable. Then we'll do the if statement that decides if we're doing a SelectFloat() based on the rss, and then we'll just call PlayPolyphonic. Should simplify the code a bit.

props/saber_fett263_buttons.h
fdarosa2663: GitHub was acting up trying to Fetch and Merge, I think I fixed now.

sound/hybrid_font.h
fdarosa2663: Updated.

profezzorn: I think we should use exactly the same logic for selecting the effect regardless of whether ProffieOSMaxSwingAceleration is set or not.  It may lead to some weird things if you try to use swings and slashes at the same time, but it might also be useful...

profezzorn: My thinking is that nobody should have both accent slashes and swing selection configured at the same time, so we might as well make the code as simple as possible with the assumption that only one of those will be active.

fdarosa2663: Sorry, maybe I'm missing something I have it so that if ProffieOSMaxSwingAcceleration is set (> Min) slsh isn't used at all but if it 's not then slsh is via the "else" but only if it exists.  Is there a better way to do this, it's escaping me at present?

fdarosa2663: Nevermind, I think I understand was looking at it backwards? Is this what you meant?

PR #250, opened 2021-09-15
Update BC_buttons. preon skip, consolidate revert CC.

NoSloppy:


props/saber_BC_buttons.h
```diff
@@ -437,6 +437,53 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     SaberBase::DoEffect(EFFECT_FAST_ON, 0);
   }
 
+  // Go to the next Preset skipping Preon effect with FastOn.
```
profezzorn, line 440: Let's move these functions to PropBase instead of duplicating it in every prop file.

props/saber_BC_buttons.h
```diff
@@ -748,12 +804,17 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     case EVENTID(BUTTON_POWER, EVENT_FIRST_SAVED_CLICK_SHORT, MODE_OFF):
       // No power on without exiting Vol Menu first
       if (!mode_volume_) {
-        On();
+      // Bypass preon if pointing up  
+        if (fusor.angle1() >  M_PI / 3) {
+          SaberBase::TurnOn();
```
profezzorn, line 809: This skips several steps that On() does, not sure if any of those are actually required though....

NoSloppy: That's was the idea. Skip preon and no font.wav played, just out.wav on the live switch of presets (and I guess FastOn too if that exists.) Not sure what faston.wav is supposed to be if we're already playing a normal ignition sound, but this works great for me.

props/saber_BC_buttons.h
NoSloppy: Sounds good. It IS right from Fett263_buttons.
Is that some of the last ironing-out for OS6, or did you mean for me to actually PR it into prop_base?

profezzorn: I meant for you to use a PR to move it into prop_base

props/saber_BC_buttons.h
profezzorn: You might want to respect NoOnOff() at least.

NoSloppy: Updated to just call FastOn() instead now.

props/saber_BC_buttons.h
NoSloppy: > I meant for you to use a PR to move it into prop_base

Ok. So in doing so, there's a macro conflict. It compiles, but with warnings of "already defined" for SET_BLADE_STYLE. While looking for ways to sort it out, I spoke with Fett263 since it was his code I copied.
He said this ties into Edit Mode and the whole breaking-things-out complication at the moment.
Maybe duplicating in each prop for now is easiest? As long as it's not "wrong", this works great for me.

profezzorn: Just rename the define, or just add #undef SET_BLADE_STYLE after it's been used.

NoSloppy: Renamed the define, updated prop_base in #263

props/saber_BC_buttons.h
```diff
@@ -437,6 +429,53 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     SaberBase::DoEffect(EFFECT_FAST_ON, 0);
   }
 
+  // Go to the next Preset skipping Preon effect with FastOn.
+  virtual void next_preset_fast() {
```
NoSloppy: @profezzorn I don't see this anywhere to remove it, so the "move to prop_base" is a move.
I click edit file and this section is not there.

NoSloppy:
Standing by to see results of other PRs to prop_base.

PR #256, opened 2021-09-28
Added usage comments to WavLen.h

NoSloppy:


functions/wavlen.h
```diff
@@ -1,6 +1,14 @@
 #ifndef FUNCTIONS_WAVELEN_H
 #define FUNCTIONS_WAVELEN_H
 
+// Usage: WavLen<EFFECT>
```
profezzorn, line 4: You can also use just WavLen<>.
If you use that inside a TransitionEffectL, then it will automatically use the right effect.

NoSloppy: oh wow. ok even better.

PR #265, opened 2021-10-10
Fix Multi-Blast check for Battle Mode

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -4155,8 +4155,10 @@ SaberFett263Buttons() : PropBase() {}
             return true;
           } else {
             SaberBase::DoBlast();
-            check_blast_ = true;
-            last_blast_millis_ = millis();
+	    if (battle_mode_) {
```
profezzorn, line 4158: This looks weirdly indented, but it might just be github being weird.

props/saber_fett263_buttons.h
```diff
@@ -4209,14 +4211,13 @@ SaberFett263Buttons() : PropBase() {}
         if (swing_blast_) {
           SaberBase::DoBlast();
           return true;
-        }
-        if (check_blast_ && battle_mode_) {
-          if (!swing_blast_ && millis() - last_blast_millis_ > 2000) {
-            swing_blast_ = true;
-            hybrid_font.PlayCommon(&SFX_blstbgn);
-            SaberBase::DoBlast();
+        } else {
```
profezzorn, line 4214: There is really no need for an else here since the return on the previous line prevents the processor from going here anyways, even without the else.

If you do want the else, then you might want to remove the "return true" on the previous line since it is now redundant. The "return true" on line 4223 can do the job instead.

profezzorn (commented):
These comments are optional, so I'm going to merge the code.
If you want to address these comments, just send another PR.

PR #270, opened 2021-10-11
Move "Fast" and UpdateFont/Style functions to prop_base

fdarosa2663:
Moving shared functions from my prop to prop_base


props/prop_base.h
```diff
@@ -435,12 +432,57 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+  }
+
+  // Update Font / Save Style in Edit Mode, skips Preon effect (except for Preon Editing previews) using FastOn
+  void UpdateFont(int preset_num, bool preon) {
+    TRACE(PROP, "start");
+    bool on = BladeOff();
+    SaveColorChangeIfNeeded();
+    // First free all styles, then allocate new ones to avoid memory
+    // fragmentation.
+    UnsetCurrentStyles();
```
profezzorn, line 474: Not sure why we're doing UnsetCurrentStyles() / SetBladeStyle() if we're only changing the font?

props/prop_base.h
```diff
@@ -435,12 +432,57 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+    SetBladeLength();
+    SetBladeStyle();
+    TRACE(PROP, "end");
+  }
+
+  // Update Font / Save Style in Edit Mode, skips Preon effect (except for Preon Editing previews) using FastOn
+  void UpdateFont(int preset_num, bool preon) {
```
profezzorn, line 468: I'm not sure why you would every want the "preon" to play if you're updating the font?

props/prop_base.h
```diff
@@ -379,37 +373,39 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
-        strcpy(message, "Preset: ");
-        itoa(current_preset_.preset_num + 1,
-             message + strlen(message), 10);
-        strcat(message, "\n");
-        strncat(message + strlen(message),
-                current_preset_.font.get(), sizeof(message) - strlen(message));
-        message[sizeof(message) - 1] = 0;
-        SaberBase::DoMessage(message);
-      }
+    current_preset_.SetPreset(current_preset_.preset_num);
```
profezzorn, line 389: This is wrong and will make SetPreset() not work.
In the previous code, this set current_preset_.SetPreset(preset_num)
in the current code, this uses the *current* preset number, which means the preset will never change.

props/prop_base.h
profezzorn: I don't think UpdateFont() should take a preset number as an argument. That's just an artifact from SetPreset()
Same for UpdateStyle.

props/prop_base.h
```diff
@@ -379,37 +373,39 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+  }
+
+  void AnnouncePreset() {
```
profezzorn, line 392: should be private

props/prop_base.h
fdarosa2663: To clarify further, the initial use for UpdateFont was the Edit Font function but as the IntArg and other pieces were added I used UpdateFont in other places because it was skipping the NEWFONT and PREON sounds and effects which didn't make sense during certain edits (except for Preon related edits) so rather than make more functions that did the same thing UpdateFont was used.  It was explained in the comment on the function but wanted to give more background to explain further.

props/prop_base.h
profezzorn: Wouldn't it be enough to just do Off(); On(); in this case?

props/prop_base.h
```diff
@@ -435,12 +416,68 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+    }
+    SetBladeLength();
+    SetBladeStyle();
     if (on) On();
     if (announce) SaberBase::DoNewFont();
     TRACE(PROP, "end");
   }
 
+  // Update Blade Style (no On/Off for use in Edit Mode)
+  void UpdateStyle(int preset_num) {
```
profezzorn, line 440: I don't think UpdateStyle should take a number, it should just work on the current preset.

props/prop_base.h
```diff
@@ -435,12 +416,68 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+    // Update Font & Style, skips Preon effect using FastOn (for use in Edit Mode and "fast" preset changes)
+  void UpdatePreset(int preset_num) {
+    TRACE(PROP, "start");
+    bool on = BladeOff();
+    SaveColorChangeIfNeeded();
+    // First free all styles, then allocate new ones to avoid memory
+    // fragmentation.
+    UnsetCurrentStyles(preset_num);
+    SetBladeStyle();
+    if (on) {
```
profezzorn, line 460: if (on) FastOn();

props/prop_base.h
```diff
@@ -435,12 +416,68 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+    // First free all styles, then allocate new ones to avoid memory
+    // fragmentation.
+    UnsetCurrentStyles(preset_num);
+    SetBladeLength();
+    SetBladeStyle();
+    TRACE(PROP, "end");
+  }
+
+    // Update Font & Style, skips Preon effect using FastOn (for use in Edit Mode and "fast" preset changes)
+  void UpdatePreset(int preset_num) {
```
profezzorn, line 452: UpdatePreset is a terrible name. It's confusingly similar to SetPreset, but doesn't clarify what the difference is or when it should be used.

I think a function that is used in next_preset_fast() should probably be called SetPresetFast(int preset_num). I'm not sure if it makes sense to use that when just changing font though.

props/prop_base.h
```diff
@@ -435,12 +416,68 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+  // Update Preon IntArg in Edit Mode
+  void UpdatePreon(int preset_num) {
+    TRACE(PROP, "start");
+    bool on = BladeOff();
+    SaveColorChangeIfNeeded();
+    // First free all styles, then allocate new ones to avoid memory
+    // fragmentation.
+    UnsetCurrentStyles(preset_num);
+    SetBladeStyle();
+    if (on) {
```
profezzorn, line 475: if (on) On();

props/prop_base.h
```diff
@@ -457,6 +502,22 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+  // Go to the previous Preset skipping NewFont and Preon effects using FastOn.
+  void previous_preset_fast() {
+#ifdef SAVE_PRESET
+    SaveState(current_preset_.preset_num - 1);
+#endif
+    UpdatePreset(current_preset_.preset_num - 1);
+  }
+
+  // Go to first Preset.
+  void first_preset() {
```
profezzorn, line 514: Why would anybody need this?
Also, if this is a "fast" function, it ought to be called "first_preset_fast".

props/prop_base.h
```diff
@@ -379,37 +373,23 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
-    // First free all styles, then allocate new ones to avoid memory
-    // fragmentation.
+    return on;
+  }
+
+  void UnsetCurrentStyles(int preset_num) {
 #define UNSET_BLADE_STYLE(N) \
     delete current_config->blade##N->UnSetStyle();
     ONCEPERBLADE(UNSET_BLADE_STYLE)
     current_preset_.SetPreset(preset_num);
```
profezzorn, line 389: This doesn't belong in a function called UnsetCurrentStyles.
Also, in the Update* functions, we probably don't want to call this at all if we can avoid it as it can cause audio glitches.

props/prop_base.h
```diff
@@ -379,37 +373,23 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
-        strcat(message, "\n");
-        strncat(message + strlen(message),
-                current_preset_.font.get(), sizeof(message) - strlen(message));
-        message[sizeof(message) - 1] = 0;
-        SaberBase::DoMessage(message);
-      }
-    }
+  }
 
+  void SetBladeLength() {
```
profezzorn, line 392: Hmm, this probably don't belong in SetPreset() at all.
We should probably do this separately.

props/prop_base.h
```diff
@@ -419,8 +399,9 @@ class PropBase : CommandParser, Looper, protected SaberBase {
 #else
 #define WRAP_BLADE_SHORTERNER(N)
 #endif
+  }
 
-
+  void SetBladeStyle() {
```
profezzorn, line 404: This name is a bit confusing, because "setting the blade style" would seem like it means setting the style string in the preset. Maybe ConfigureBladeStyles, or AllocateBladeStyles.

If we go with AllocateBladeStyles, then the function above should probably be called FreeBladeStyles()

props/prop_base.h
```diff
@@ -435,12 +416,68 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
 
+  // Select preset (font/style)
+  virtual void SetPreset(int preset_num, bool announce) {
+    TRACE(PROP, "start");
+    bool on = BladeOff();
+    SaveColorChangeIfNeeded();
+    // First free all styles, then allocate new ones to avoid memory
+    // fragmentation.
+    UnsetCurrentStyles(preset_num);
+    if (announce) {
```
profezzorn, line 429: if (announce) AnnouncePreset();

props/prop_base.h
fdarosa2663: changed

props/prop_base.h
fdarosa2663: fixed

props/prop_base.h
fdarosa2663: If I recall this is run even during Update* to prevent memory fragmentation when we need to parse/reload the arguments.

fdarosa2663: moved current_preset_.SetPreset(preset_num) to main functions

props/prop_base.h
fdarosa2663: renamed AllocateBladeStyles and FreeBladeStyles

props/prop_base.h
profezzorn: Actually never mind.
I looked at this functoin, and it doesn't actually *do* anything except for restoring the savestate.
Everything else is done in AllocatedBladeStyles(). Breaking this out from AllocateBladeStyles doesn't really work, so just put together with AllocateBladeStyles().

props/prop_base.h
profezzorn: Keep it in your prop for now.

props/prop_base.h
```diff
@@ -435,12 +412,64 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+  }
 
+  // Select preset (font/style)
+  virtual void SetPreset(int preset_num, bool announce) {
+    TRACE(PROP, "start");
+    bool on = BladeOff();
+    SaveColorChangeIfNeeded();
+    // First free all styles, then allocate new ones to avoid memory
+    // fragmentation.
+    FreeBladeStyles(preset_num);
```
profezzorn, line 424: This is weird, because FreeBladeStyles() does current_preset_.SetPreset(), but then we do it again two lines down...

props/prop_base.h
```diff
@@ -379,37 +373,22 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
 #endif
     bool on = SaberBase::IsOn();
     if (on) Off();
-    SaveColorChangeIfNeeded();
-    // First free all styles, then allocate new ones to avoid memory
-    // fragmentation.
+    return on;
+  }
+
+  void FreeBladeStyles(int preset_num) {
```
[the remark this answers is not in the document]
fdarosa2663: removed

props/prop_base.h
fdarosa2663: FreeBladeStyles() only does UNSET_BLADE_STYLE

props/prop_base.h
```diff
@@ -435,12 +412,64 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+    // fragmentation.
+    FreeBladeStyles(preset_num);
+    current_preset_.SetPreset(preset_num);
+    AllocateBladeStyles();
+    if (on) FastOn();
+    TRACE(PROP, "end");
+  }
+
+  // Update Preon IntArg in Edit Mode
+  void UpdatePreon(int preset_num) {
```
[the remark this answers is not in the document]
fdarosa2663: removed

PR #273, opened 2021-10-16
Set CLASH_THRESHOLD_G max value to MAX_CLASH in menu

fdarosa2663:


fdarosa2663:
Because MAX_CLASH can be edited via Edit Mode I used saved_gesture_control.maxclash to determine the maximum value for Clash Threshold.

PR #282, opened 2021-10-20
add check for tracks within font / common for TrackPlayer mode

fdarosa2663:


profezzorn:
Have you tested this?
From what I can tell, this doesn't have an override for "list_current_tracks", which means that it will list all available tracks, not just the ones in the current font. It also doesn't have an override for StartOrStopTrack(), which I think means it will only play the one track listed in the preset, right?

profezzorn:
Looks like I missed some stuff when I decided that it might not work.
I'm going to merge it, and if it doesn't work, we'll fix it later... :)

PR #285, opened 2021-10-20
BC_buttons - Consolidate Revert Colorchange code

NoSloppy:


NoSloppy:
It's only better by seeming a little cleaner to me? Either way works.

PR #287, opened 2021-10-20
BC_Buttons - Bypass preon if pointing up, NewFont too if blade is on …

NoSloppy:
…(next_preset_fast)


props/saber_BC_buttons.h
```diff
@@ -753,23 +762,58 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     return true;
 
 // Blaster Deflection
+    // original sa22c mode if ENABLE_AUTO_SWING_BLAST is not defined
```
profezzorn, line 765: I think you need to sync before I merge, as this seems to undo some changes I just merged.

NoSloppy: Ok, i think this is right now. Made the changes here too so it doesn't conflict?
Although it looks like I hit GO twice :/

PR #289, opened 2021-10-20
BC_buttons - use sound_library for speech things.

NoSloppy:


NoSloppy:
So I see in serial monitor `WARNING, wav player still referenced!`
I am not sure about how `wav_player.Free();` should be done in and amongst this file now.

PR #292, opened 2021-10-21
ssd1306 - Add idle.bmp, a Work In Progress..   *see notes

NoSloppy:
Well this is a start? I'm sure it's sloppy coding, but it works as expected ... mostly.
Working correctly whether idle.bmp exists in the font or not. 
At boot, it shows when blade is off until first ignition. Then conditionals:
If Mass storage is turned on,
  If USB is connected, it shows battery meter so the SD can be available to the PC. 
  If USB is disconnected, we can access the SD and play idle.bmp instead when the blade is off.
This is working!
However, as currently written, battery meter alternates with the idle.bmp after one loop and it goes back and forth. Maybe cool, but not the intention.
Then plugging USB back in , idle stops showing...but the SD doesn't mount to the PC.
So close.
I do admit, I have a hard time following the flow of this file.

Also, I have control, but no longer get STDOUT messages over TTL with it set up correctly, so no monitoring what's going on with NO USB.
Any chance changes to serial stuff semi-recently is causing that?

Once this part is figured out, maybe a few easier things to add could be:
- being able to call for battery meter to show when On-Demand battery level invoked.
- adding support for out.bmp and in.bmp to play with ignition and retraction.

### DIFF
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -11,7 +11,7 @@ IMAGE_FILESET(clsh);
 IMAGE_FILESET(blst);
 IMAGE_FILESET(lock);
 IMAGE_FILESET(force);
-
+IMAGE_FILESET(idle);
 
 template<int Width, class col_t>
 class Display : public MonoFrame<Width, col_t> {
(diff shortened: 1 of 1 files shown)


profezzorn:
This has been implemented.

NoSloppy:
> This has been implemented.

Brilliant, thanks!

PR #293, opened 2021-10-21
Fix wav player warning

NoSloppy:
This seems to do the trick.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -715,7 +715,7 @@ public:
 // Spoken Battery Level in volts
     case EVENTID(BUTTON_POWER, EVENT_THIRD_SAVED_CLICK_SHORT, MODE_OFF):
       if (!mode_volume_) {
-        sound_library_.SayBatteryLevel();
+        sound_library_.SayTheBatteryLevelIs();
         sound_library_.SayNumber(battery_monitor.battery(), SAY_DECIMAL);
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -714,24 +715,22 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
 // Spoken Battery Level in volts
     case EVENTID(BUTTON_POWER, EVENT_THIRD_SAVED_CLICK_SHORT, MODE_OFF):
       if (!mode_volume_) {
-        sound_library_.SayBatteryLevel();
+        sound_library_.SayTheBatteryLevelIs();
```
profezzorn, line 718: This seems unrelated to this change.

props/saber_BC_buttons.h
```diff
@@ -356,6 +356,7 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     DetectTwist();
     Vec3 mss = fusor.mss();
     sound_library_.Poll(wav_player);
+    if (!wav_player) wav_player.Free();
```
profezzorn, line 359: I don't think this does anything. If wav_player is false, then there is no need to free it..

I think what you want is actually something like:

if (wav_player && !wav_player->IsPlaying()) wav_player.Free();

NoSloppy:
Resubmitting alone

PR #296, opened 2021-10-21
BC_buttons - follow sound_library functions update

NoSloppy:


profezzorn:
When you rename a function, all the calls to that function should be fixed in the same CL.

profezzorn:
PS: If you need/want to split it up in to smaller pieces, the right way to do that is to do a small set of functions at a time, not to split it by file.

PR #300, opened 2021-10-22
Add TrSequence<>

fdarosa2663:


transitions/sequence.h
```diff
@@ -0,0 +1,45 @@
...
+class TrSequence {
+public:
+  void begin() {
+    begin_ = true;
+    n_ = (n_ + 1) % sizeof...(TRANSITION);
+  }
+  void run(BladeBase* blade) {
+    if (begin_) {
+      begin_ = false;
+      while(n_ < 0) n_ += sizeof...(TRANSITION) << 8;
```
profezzorn, line 21: This while() can be removed,

transitions/sequence.h
```diff
@@ -0,0 +1,45 @@
...
+// TR1, TR2: TRANSITION
+// return value: TRANSITION
+// transition options used in sequence
+
+template<class... TRANSITION>
+class TrSequence {
+public:
+  void begin() {
+    begin_ = true;
+    n_ = (n_ + 1) % sizeof...(TRANSITION);
```
profezzorn, line 16: This increment should happen in run() if begin_ is true.

fdarosa2663: moved

PR #312, opened 2021-10-26
add 1 Button controls to Fett263 Prop

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -1249,6 +1404,37 @@ SaberFett263Buttons() : PropBase() {}
     color_mode_ = NONE;
   }
 
+  // Check if current style uses RgbArg for CC_COLOR_LIST
```
profezzorn, line 1407: and do what?
Normally when I see a function called "CheckSOMETHING" I kind of expect it to return a bool if SOMETHING is going on.  This function doesn't return a bool, so clearly it does something if the current style uses RgbArg, but what?

props/saber_fett263_buttons.h
```diff
@@ -2390,119 +2626,119 @@ SaberFett263Buttons() : PropBase() {}
...
-	} else {
-	  sound_library_.SaySelect();
-	  menu_type_ = MENU_PREON_SIZE;
-	  arg_revert_ = strtol (argspace, NULL, 0);
-	}
-	break;
-      default:
-	break;
+          break;
+        case IGNITION_OPTION_ARG:
```
profezzorn, line 2644: Is this new? It looks new in this diff?
Is it something unrelated? Or is it just moved around?

props/saber_fett263_buttons.h
```diff
@@ -3616,6 +3852,79 @@ SaberFett263Buttons() : PropBase() {}
...
+
+  void ToggleBattleModeMultiBlast() {
+    if (battle_mode_ && !swing_blast_) {
+      check_blast_ = true;
+      last_blast_millis_ = millis();
+    }
+    swing_blast_ = false;
+  }
+
+  void DoCheckMultiBlast() {
```
profezzorn, line 3898: Another CheckSOMETHING function that doesn't return a bool.
Either needs a new name or a comment.

profezzorn (commented):
Unfortunately this PR has become very difficult to read, hopefully you're testing it well because it's not really possible for me to double-check everything anymore.

props/saber_fett263_buttons.h
fdarosa2663: renamed to ToggleCCMode() and updated comment, we shouldn't use ColorWheel if styles use RgbArg so we need to check for argument and then use CC_COLOR_LIST instead

props/saber_fett263_buttons.h
fdarosa2663: nothing new just fixed indent for the sound_library_ above

props/saber_fett263_buttons.h
fdarosa2663: renamed to DoAutoMultiBlast()

fdarosa2663:
I've been testing as I go, the large diff was me trying to fix indents, no code changed.  I think when you added the sound_library_ it messed up those indents or they were a find and replace with the wrong indentation.  As I've tried to fix github seems to be adding it's own indentation but it isn't correct either.  When I view on github.com the indentation looks correct but when I edit locally it's all over the place.
I have 2 more additions lined up after this PR is merged - I'll do them separately.  As to the indentation, I'm not sure an easy way to fix without causing these large diffs and in a lot of cases they look fine in one location but then wrong in another so it's confusing to fix.

PR #313, opened 2021-10-29
add fusor.rss() and SwingAcceleration<>

fdarosa2663:


common/fuse.h
```diff
@@ -290,6 +290,13 @@ class Fusor : public Looper {
   Vec3 speed() { return speed_; }  // m/s
 #endif
 
+  float rss() {
+    Vec3 gyro_slope = gyro_extrapolator_.slope() * 1000000;
```
profezzorn, line 294: This should just call gyro_slope()

common/fuse.h
```diff
@@ -290,6 +290,13 @@ class Fusor : public Looper {
   Vec3 speed() { return speed_; }  // m/s
 #endif
 
+  float rss() {
```
profezzorn, line 293: Needs a comment to explain what it does.
(Or a MUCH better name....)

common/fuse.h
```diff
@@ -290,6 +290,13 @@ class Fusor : public Looper {
...
+    Vec3 gyro_slope = gyro_extrapolator_.slope() * 1000000;
+    // Radians per second per second
+    float rss = sqrtf(gyro_slope.z * gyro_slope.z + gyro_slope.y * gyro_slope.y) * (M_PI / 180);
```
profezzorn, line 296: variable is not needed, just do: return sqrtf(....) * (...);

functions/swing_speed.h
```diff
@@ -26,4 +26,27 @@ using SwingSpeedX = SingleValueAdapter<SwingSpeedSVF<MAX>>;
...
+// Usage: SwingAcceleration<MAX>
+// Returns 0-32768 based on swing acceleration
+// MAX defaults to 150
+// returned value: INTEGER
+template<class MAX = Int<130>>
+class SwingAccelerationSVF {
+ public:
+  void run(BladeBase* blade) { max_.run(blade); }
+ int calculate(BladeBase* blade) {
+   max_.run(blade);
```
profezzorn, line 38: No need to call max_.run() again. (It's already done in SwingAccelerationSVF::run() above)

common/fuse.h
fdarosa2663: done

common/fuse.h
fdarosa2663: renamed to swing_accel() and added comment

common/fuse.h
fdarosa2663: updated

functions/swing_speed.h
fdarosa2663: remove duplication in SwingSpeed as well

PR #318, opened 2021-11-01
Add "fix" for over-turning menu during prompts

fdarosa2663:
added reset for menu dial if "stepped" past prompt


props/saber_fett263_buttons.h
```diff
@@ -3924,7 +3918,13 @@ SaberFett263Buttons() : PropBase() {}
     battle_mode_ = true;
 #endif
   }
-	
+
```
profezzorn, line 3921: This looks like it came from the other PR?

fdarosa2663: ah, yup, I'll fix

PR #321, opened 2021-11-02
Add MenuDialIncrement()

fdarosa2663:
Allows Preset, Font, Track and Blade Length dial to increment by 5


props/saber_fett263_buttons.h
```diff
@@ -4878,27 +4902,53 @@ SaberFett263Buttons() : PropBase() {}
...
           return true;
         }
-        if (menu_ && menu_type_ == MENU_TRACK_PLAYER) {
-          track_mode_ = PLAYBACK_ROTATE;
-          sound_library_.SayRotate();
+        if (menu_) MenuDialIncrement(1);
+        return true;
+
+      case EVENTID(BUTTON_NONE, EVENT_TWIST_LEFT, MODE_ON | BUTTON_POWER):
+        if (wav_player && wav_player->isPlaying()) {
```
profezzorn, line 4914: Why don't we just put this if statement inside MenuDialIncrement() ?

fdarosa2663: Do you mean move the wav_player check to MenuDial()?  MenuDialIncrement() is just a modifier for certain menus to let the user increment by 5 "steps" then it calls MenuDial().
I suppose we could.  Do you want it in this PR or should we merge this and I'll do in a separate.

profezzorn: This CL, but only if it doesn't blow up and make this CL much bigger.

fdarosa2663: I think it's going to blow this one up because I'll need to indent the entire switch() for MenuDial() to accommodate the if-else for the wav_player_ check

profezzorn: Fine, let's do it in a separate PR then.

PR #322, opened 2021-11-02
BC_buttons - housekeeping. Prune unused, consolidate, comments edit.

NoSloppy:


props/saber_BC_buttons.h
```diff
@@ -801,7 +796,8 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     case EVENTID(BUTTON_NONE, EVENT_CLASH, MODE_ON | BUTTON_AUX):
   #endif
       if (!SaberBase::Lockup()) {
-        if (pointing_down_) {
+        //pointing down
+        if (fusor.angle1() < - M_PI / 4) {
```
profezzorn, line 800: Are you sure you're not giving up a feature here?
(Being able to decide if it's a drag when you press the button makes it possible to do "drag" effects across the ceiling for instance....)

NoSloppy: That could be cool, don't think I've ever seen someone use it like that before.
I feel it would take some massaging to differentiate it from a vertically positioned blade defensive lockup?
So let's say yes, unused for now, but thanks for the tip (no pun intended), I'll keep it in mind for next revision maybe. (already a WIP)
I guess I just assumed the fusor angle was a more accurate method that someone above my pay grade decided to start using, so I figured make things uniform throughout the prop. It did work fine that way of course, but was only used for the down position in that one spot.  ¯\_(ツ)_/¯

PR #324, opened 2021-11-02
move wav_player_ check to MenuDial()

fdarosa2663:
The switch() in MenuDial() was indented externally and pasted, no code changes were made.  I did fix a few internal indentation issues for sound_library_ as well but the only change was the addition of the

if (wav_player && wav_player->isPlaying()) {
      current_menu_angle_ = fusor.angle2();
    } else {
...
}


fdarosa2663:
will resubmit after the saber_off_time_millis_ addition is merged as the two will conflict right now.

PR #325, opened 2021-11-05
reset saber_off_time_millis_ on button press to fix MOTION_TIMEOUT

fdarosa2663:
this may be the cause of motion issues reported by sa22c


fdarosa2663:
I added some twist event fixes, it looks like these may be tied to the reported issues as the DetectMenuTwist() and last_twist_millis_ may have been interfering with normal EVENT_TWIST in certain scenarios.

PR #330, opened 2021-11-12
Track Player update

fdarosa2663:
fix for Loop, changing Player to use track_num 1 instead of default as starting track to prevent repeat, check for tracks and release track player if no Playback.


props/saber_fett263_buttons.h
```diff
@@ -1694,23 +1698,25 @@ SaberFett263Buttons() : PropBase() {}
...
     }
-    StartOrStopTrack(); 
   }
 
   void TrackPlayer() {
     if (track_mode_ != PLAYBACK_OFF) {
       if (!track_player_) {
-#ifdef ENABLE_AUDIO
-        if (track_num_ >= 0 && track_mode_ == PLAYBACK_LOOP) {
+        if (track_num_ <= 0 && track_mode_ == PLAYBACK_LOOP) {
```
profezzorn, line 1719: It seems weird to test both track_num_ and track_mode_ here.
It seems to me that just checking track_num_ == 0 would be enough?

props/saber_fett263_buttons.h
```diff
@@ -1725,19 +1731,26 @@ SaberFett263Buttons() : PropBase() {}
...
+  void PlayTrack() {
+    char playtrack[128];
+    RunCommandAndGetSingleLine("list_current_tracks", nullptr, track_num_, playtrack, sizeof(playtrack));
+    MountSDCard();
+    EnableAmplifier();
+    track_player_ = GetFreeWavPlayer();
+    if (track_player_) {
+      track_player_->Play(playtrack);
+    } else {
+      STDOUT.println("No available WAV players.");
```
profezzorn, line 1753: If this turns into a common problem, it could be helpful to work around it by checking for available wav players in loop() until one is available and THEN start the playback.

props/saber_fett263_buttons.h
fdarosa2663: I have to check for PLAYBACK_LOOP otherwise the default track just keeps playing itself no matter what the user does if they don't have other tracks in font because TrackPlayer() is in Loop() and will run StartorStopTrack()

props/saber_fett263_buttons.h
fdarosa2663: I haven't experienced any trouble with get wav players since TrackPlayer can only be entered while saber is off but if you want we can add.  Once track_player_ gets a wav player it keeps until we Free() correct?  I haven't had any reports of trouble but if there are we can add.

props/saber_fett263_buttons.h
profezzorn: ok, but it looks weird

PR #335, opened 2021-11-14
Fix for long click or hold in Color or Length Editing

fdarosa2663:
Need to make sure we revert the ShowColor style before letting users cancel Color or Length Editing with a long click or held button


fdarosa2663:
need to correct, fix didn't work

PR #338, opened 2021-11-15
Fix / Update for ShowColor exiting and Color Zoom

fdarosa2663:
These two are semi-related so submitted together.  This update is to handle clicks and holds while we have ShowColor styles active and also to make Color Zoom functionality for all modes use the same controls to reduce confusion.


props/saber_fett263_buttons.h
```diff
@@ -2024,24 +2033,88 @@ SaberFett263Buttons() : PropBase() {}
   }
 #endif
 
-  void DoColorZoom() {
-    if (color_mode_ == COLOR_LIST) {
+  bool CancelShowColor() {
+    switch (menu_type_) {
```
profezzorn, line 2037: I seem to remember you having a similar list of cases elsewhere in the code.
Would it be possible to break it out into a helper function. Something like "bool isInMenu()" maybe?

props/saber_fett263_buttons.h
```diff
@@ -2024,24 +2033,88 @@ SaberFett263Buttons() : PropBase() {}
   }
 #endif
 
-  void DoColorZoom() {
-    if (color_mode_ == COLOR_LIST) {
+  bool CancelShowColor() {
```
profezzorn, line 2036: Needs a comment to explain what this does and what it returns.

props/saber_fett263_buttons.h
```diff
@@ -2024,24 +2033,88 @@ SaberFett263Buttons() : PropBase() {}
...
-  bool EndCCColorList() {
-    if (color_mode_ != CC_COLOR_LIST) return false;
-    hybrid_font.PlayCommon(&SFX_ccend);
-    NewColor(1, BASE_COLOR_ARG);
-    current_preset_.Save();
-    show_color_all_.Stop();
-    UpdateStyle();
-    wav_player.Free();
-    return true;
+  bool EndColorZoom() {
```
profezzorn, line 2095: Needs comment.

props/saber_fett263_buttons.h
```diff
@@ -4563,9 +4613,40 @@ SaberFett263Buttons() : PropBase() {}
...
+      case EVENTID(BUTTON_POWER, EVENT_RELEASED, MODE_ON):
+#ifndef DISABLE_COLOR_CHANGE
+        if (EndColorZoom()) return true;
+        if (SaberBase::GetColorChangeMode() != SaberBase::COLOR_CHANGE_MODE_NONE) {
+          ToggleColorChangeMode();
+#ifdef FETT263_EDIT_MODE_MENU
+          if (menu_type_ == MENU_COLOR) {
+            menu_type_ = MENU_TOP;
+            MenuSave();
+	    return true;
```
profezzorn, line 4637: indentation looks wrong (on github at least)

props/saber_fett263_buttons.h
fdarosa2663: I think you're thinking of the MenuChoice(), MenuIncrement() and MenuUndo() functions although those are specific to actions taken in the menu not sure if a helper function would make it easier or harder to use.

props/saber_fett263_buttons.h
fdarosa2663: added

props/saber_fett263_buttons.h
fdarosa2663: added

props/saber_fett263_buttons.h
fdarosa2663: fixed (I think) it looks ok on my end

PR #343, opened 2021-11-18
Add CopyPreset

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -2073,6 +2076,18 @@ SaberFett263Buttons() : PropBase() {}
   }
 #endif
 
+  void DoCopyPreset() {
```
profezzorn, line 2079: "CopyPreset" and "DoCopyPreset" sounds like they do the same thing.
Add comments to explain what they do, and make sure that it clarifies the difference.

fdarosa2663: consolidated to one function, only difference is whether it's used in menu or not, when used outside menu we need an "announce" when in the menu this is not used.

profezzorn: That works, but even better would have been to do it like:

CopyPreset() { ... }
CopyPresetAndAnnounce() { CopyPreset();  /* do the announcement here */ }

PR #344, opened 2021-11-20
blaster.h fix for dual_prop use.

NoSloppy:
Smoothswings play when blaster mode is invoked by blade removal without this condition.

### DIFF
--- a/props/blaster.h
+++ b/props/blaster.h
@@ -275,7 +275,7 @@ public:
 
   // Make swings do nothing
   void DoMotion(const Vec3& motion, bool clear) override {
-    PropBase::DoMotion(motion, clear);
+    PropBase::DoMotion(Vec3(0), clear);
   }
 
   bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
--- a/props/blaster.h
+++ b/props/blaster.h
@@ -273,6 +273,7 @@ public:
     }
   }
 
+  // Make swings do nothing
   void DoMotion(const Vec3& motion, bool clear) override {
     PropBase::DoMotion(Vec3(), clear);
   }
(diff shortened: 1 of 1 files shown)


props/blaster.h
```diff
@@ -273,8 +273,10 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
     }
   }
 
+#ifndef PROPS_DUAL_PROP_H
```
profezzorn, line 276: I wonder if this might be a better fix:

void DoMotion(const Vec3& motion, bool clear) override {
  PropBase::DoMotion(Vec3(), clear);
}

(No #ifdefs required.)

NoSloppy: Tested, that works.

NoSloppy: > 

Wait - hold that thought. That breaks Clash to Unjam.

NoSloppy: Doing `PropBase::DoMotion(motion, clear);` instead seems to be right. Commit made.

props/blaster.h
```diff
@@ -274,7 +274,9 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
   }
 
   // Make swings do nothing
-  void DoMotion(const Vec3& motion, bool clear) override {}
+  void DoMotion(const Vec3& motion, bool clear) override {
+    PropBase::DoMotion(motion, clear);
```
profezzorn, line 278: This does nothing, it is exactly equivalent to not having a DoMotion() in this blaster prop at all.

Did you try replacing motion with "Vec3(0)" ? 
This would mean that DoMotion() would still be called, but with a zero vector, like i fyou were holding it perfectly still.

NoSloppy: ah. I do admit I'm unclear with how having `void DoMotion`  in 3 different files works, as far as which takes precedence etc....
but "Vec3(0)" does work, where "Vec()" did not.
Updated again.

PR #346, opened 2021-11-22
Choreography fixes per Beta thread

fdarosa2663:


fdarosa2663:
It looks like it worked.  This should be ok to merge now.  Will have to address the ColorZoom / Revert issue for 1 button next week.

PR #348, opened 2021-11-22
remove gesture "spam" delays

fdarosa2663:
trying separate PR


profezzorn:
DOn't think this worked the way you wanted to either.

fdarosa2663:
It did, the large sections are because I adjusted the indentation when I took out the delay, but they are all only the changes or removing the various "delays" for gestures

fdarosa2663:
for some reason just changing indentation, github marks as a change to the whole section but I only removed the indent on my end

PR #350, opened 2021-11-24
ssd1306 - fix no message shown if idle.bmp exists.

NoSloppy:


NoSloppy:
Although seemingly working on actual OLED, the serial monitor is spamming "read image fail" when executing this bit....

NoSloppy:
aaah. Further testing shows a leftover.
If font.bmp is interrupted by doing change preset to a non-font.bmp preset, the message shows, then when duration expires, the rest of the previous, interrupted font.bmp finishes, THEN goes to idle.
Also, images are sometimes scrambled for a moment before displaying correctly. 
So I was close, but no cigar on this "fix"

profezzorn:
I've made a different fix for this.
The PR was helpful though!

PR #352, opened 2021-11-29
fontName " Font directory not found" message is helpful

NoSloppy:
...especially with additional fonts in the path, this clarifies which one(s).
Also useful for when no speaker is hooked up.
I found myself in both situations tonight and this was great small addition.

### DIFF
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -419,29 +419,30 @@ class Effect {
             ScanAll(dir, iter.name());
           }
         }
-
-      }
-
+      } else {
 #ifdef ENABLE_AUDIO
 #if VERSION_MAJOR < 4
-      else if (strlen(dir) > 8) { // TODO: Check individual path segments
-        STDOUT << "\nFont directory not found\n";
-        STDOUT << " and name exceeds 8 characters: " << dir << "\n";
-        talkie.Say(talkie_font_directory_15, 15);
-        talkie.Say(talkie_too_long_15, 15);
-      } else if (strlen(dir)) {
-        STDOUT << "\nFont directory not found: " << dir << "\n";
-        talkie.Say(talkie_font_directory_15, 15);
-        talkie.Say(talkie_not_found_15, 15);
-      } 
+        if (strlen(dir) > 8) { // TODO: Check individual path segments
+          STDOUT.println(" done");
+          STDOUT << "Font directory not found,\n";
+          STDOUT << "and name exceeds 8 characters: " << dir << "\n";
+          talkie.Say(talkie_font_directory_15, 15);
+          talkie.Say(talkie_too_long_15, 15);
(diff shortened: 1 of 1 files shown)


sound/effect.h
```diff
@@ -427,6 +427,7 @@ class Effect {
         talkie.Say(talkie_font_directory_15, 15);
         talkie.Say(talkie_too_long_15, 15);
       } else if (strlen(dir)) {
+	STDOUT.println(" Font directory not found.");
```
profezzorn, line 430: Use the new syntax, and maybe add the actual directory, like:

STDOUT << "Font directory not found: " << dir << "\n";

NoSloppy: Cool, done. Wasn't sure how else newline would be done if not using println. I did what works but is there a better way?
Also, how about this addition to exclude the "too long" message unless it's a Teensy?

sound/effect.h
```diff
@@ -423,13 +423,24 @@ class Effect {
...
 
 #ifdef ENABLE_AUDIO
+#if VERSION_MAJOR < 4
       else if (strlen(dir) > 8) { // TODO: Check individual path segments
+        STDOUT << "\nFont directory not found\n";
+        STDOUT << " and name exceeds 8 characters: " << dir << "\n";
         talkie.Say(talkie_font_directory_15, 15);
         talkie.Say(talkie_too_long_15, 15);
       } else if (strlen(dir)) {
+        STDOUT << "\nFont directory not found: " << dir << "\n";
```
profezzorn, line 433: Why have a newline at the beginning of the string?

sound/effect.h
```diff
@@ -423,13 +423,24 @@ class Effect {
       }
 
 #ifdef ENABLE_AUDIO
+#if VERSION_MAJOR < 4
       else if (strlen(dir) > 8) { // TODO: Check individual path segments
+        STDOUT << "\nFont directory not found\n";
+        STDOUT << " and name exceeds 8 characters: " << dir << "\n";
```
profezzorn, line 429: Just do it like:

STDOUT << "line one\n"
                      "line two: " << dir << "\n"

NoSloppy: That's what I was saying about the new line confusion. 
Without a new line at the front end, it tacks on to the end previous printout like
```
Scanning sound font: TeensySFFont directory not found: TeensySF
```

profezzorn: That just means that the "done" should be moved up.

NoSloppy: Couldn't get just moving "done" to work because then all the scanned fonts had no line return, but duplicating does.
Interested what's up with the "new syntax" for the message lines as opposed to println. Would that not have been easier or is it now improper?

profezzorn:
I cleaned this up a bit and checked it in myself.

PR #354, opened 2021-11-29
Fix DialIncrement() for MENU_PRESET

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -4328,6 +4331,7 @@ SaberFett263Buttons() : PropBase() {}
         }
         if (!menu_) {
           StartMenu(MENU_PRESET);
+          num_presets_ = GetNumberOfPresets() - 1;
```
profezzorn, line 4334: put the -1 in the function

props/saber_fett263_buttons.h
```diff
@@ -3026,18 +3032,15 @@ SaberFett263Buttons() : PropBase() {}
...
-            SelectPreset(current_preset_.preset_num + direction);
-          }
-        } else {
-          if (SaberBase::IsOn()) {
             previous_preset_fast();
-          } else {
-            SelectPreset(current_preset_.preset_num + direction);
           }
+        } else {
+          int preset = (current_preset_.preset_num + direction) % num_presets_;
```
profezzorn, line 3042: Why not just call GetNumberOfPresets() here?

fdarosa2663: Since it's a static number I figured it was better to just get the value once when entering the mode instead of getting the same value on each turn.

props/saber_fett263_buttons.h
fdarosa2663: moved

props/saber_fett263_buttons.h
profezzorn: There is also a drawback to getting it once: It makes the code slightly more complicated. More complicated code is only worth it if you get something for it.  My guess is that the user will never notice the difference if you just put the call here.

fdarosa2663: OK, consolidated to the MenuDial() case

props/saber_fett263_buttons.h
```diff
@@ -3026,18 +3032,15 @@ SaberFett263Buttons() : PropBase() {}
...
-            SelectPreset(current_preset_.preset_num + direction);
-          }
-        } else {
-          if (SaberBase::IsOn()) {
             previous_preset_fast();
-          } else {
-            SelectPreset(current_preset_.preset_num + direction);
           }
+        } else {
+          int preset = (current_preset_.preset_num + direction) % GetNumberOfPresets();
```
profezzorn, line 3042: I recommend replacing this with:

int num_presets = GetNumberOfPresets();
int preset = (current_preset_.preset_num + num_presets * 10 + direction) % num_presets;

Otherwise it might end up doing weird stuff with negative numbers.

fdarosa2663: this seems to work differently, I can't actually dial to the final 2 presets using this

fdarosa2663: the previous version worked without issue for normal and incremented turns in testing in both directions

fdarosa2663: Nvm, that seems to skip the 1st two presets if I start by moving backwards first, need to figure out why

fdarosa2663: OK, fixed and working now

PR #355, opened 2021-11-30
Improve DetectMenuTurn()

fdarosa2663:
After watching some user's videos I removed the delay between detection and changed current_menu_angle_ to reset to fusor.angle2() instead of twist_menu_ on EVENT to improve responsiveness.  Dial works better now and doesn't get lost between Right/Left when turns are done past "steps" or in quick succession back and forth.


props/saber_fett263_buttons.h
```diff
@@ -1827,22 +1827,16 @@ SaberFett263Buttons() : PropBase() {}
...
-          if (current_menu_angle_ < M_PI) current_menu_angle_ += M_PI * 2;
-          Event(BUTTON_NONE, EVENT_TWIST_LEFT);
-          last_rotate_millis_ = millis();
-        }
+      float a = fusor.angle2() - current_menu_angle_;
+      if (a > M_PI) a-=M_PI*2;
+      if (a < -M_PI) a+=M_PI*2;
+      if (a > twist_menu_ * 2/3) {
+        Event(BUTTON_NONE, EVENT_TWIST_RIGHT);
+        current_menu_angle_ = fusor.angle2();
```
profezzorn, line 1835: Doing it this way means that rotating one way and then back might not go back to the same place, but if it works better for you I guess that's ok.

PR #359, opened 2021-12-05
blaster files updates

NoSloppy:


props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,797 @@
+/*
```
profezzorn, line 1: seems like this should be a separate PR

NoSloppy:
OK. I'll close and resubmit 2 separate PRs.

PR #363, opened 2021-12-08
Remove ccbegin sound when entering COLOR_CHANGE_ZOOMED

fdarosa2663:
Currently both ccbegin.wav and SayZoomingIn() sounds are playing at the same time when entering COLOR_CHANGE_ZOOMED, need to not run DoChange when entering COLOR_CHANGE_ZOOMED to prevent.  The sound for this change is now handled in prop.


common/saber_base.h
```diff
@@ -303,7 +303,7 @@ public:                                                         \
     if (mode == COLOR_CHANGE_MODE_NONE) {
       DoChange(EXIT_COLOR_CHANGE);
     } else {
-      DoChange(ENTER_COLOR_CHANGE);
+      if (mode != COLOR_CHANGE_MODE_ZOOMED) DoChange(ENTER_COLOR_CHANGE);
```
profezzorn, line 306: This doesn't seem right.
I think we need to look at what color_change_mode_ was before and call DoChange(ENTER_COLOR_CHANGE) if color_change_mode_ was COLOR_CHANGE_MODE_NONE before.

fdarosa2663: Is this what you were thinking?
Currently in my prop, since we have two types of ColorZoom (COLOR_CHANGE_MODE_ZOOMED & Color Editing), I handle the sound for changing modes in the prop itself, not sure if other props will need transitions to other modes might be simpler to leave that on the prop side.

common/saber_base.h
```diff
@@ -299,11 +299,12 @@ public:                                                         \
 
   static ColorChangeMode GetColorChangeMode() { return color_change_mode_; }
   static void SetColorChangeMode(ColorChangeMode  mode) {
+    ColorChangeMode prev_mode = color_change_mode_;
     color_change_mode_ = mode;
     if (mode == COLOR_CHANGE_MODE_NONE) {
       DoChange(EXIT_COLOR_CHANGE);
     } else {
-      DoChange(ENTER_COLOR_CHANGE);
+      if (prev_mode == COLOR_CHANGE_MODE_NONE) DoChange(ENTER_COLOR_CHANGE);
```
profezzorn, line 307: This seems like it would work better, but we can clean it up a bit, instead of:

if (mode ...) {
   Do...
} else {
  if (prev_mode....) Do...
}

do:
if (mode ...) {
  Do...
} else if (prev_mode ...) {
  Do...
}

fdarosa2663: updated

PR #365, opened 2021-12-09
Update blaster.h

NoSloppy:
- Don't Jam if empty
- Autofire now counts bullets (uses WavLen of auto.wav per shot ! )
- If already full, Reload plays full.wav


profezzorn:
This PR is also cursed.
Please re-upload.

PR #375, opened 2021-12-29
Option for OLED time out

NoSloppy:
Turning screen off after font_config.ProffieOSFontImageDuration sort of defeats the purpose if there's an idle image, as it wouldn't be shown for very long.
This edit would allow idle.bmp to display (if it exists) until IDLE_OFF_TIME expires.
If no idle image exists, it gives the user a discrete, configurable OLED_OFF_TIME option, with the original font_config.ProffieOSFontImageDuration as default if not specified.

### DIFF
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -340,7 +340,7 @@ public:
         return font_config.ProffieOSFontImageDuration;
 
       case SCREEN_PLI:
-        if (!IMG_idle && !SaberBase::IsOn() && t_ > OLED_OFF_TIME) {
+        if (!SaberBase::IsOn() && t_ > OLED_OFF_TIME) {
           screen_ = SCREEN_OFF;
           return FillFrameBuffer2(advance);
         }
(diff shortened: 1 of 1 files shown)


profezzorn:
I'm confused, this PR seem to change the PLI behavior, not the idle.bmp behavior.

NoSloppy:
Correct. 

The current bit of code was added with your major revamp of the file (adding monoframe, etc...). I had mentioned it seems odd to turn off the screen so soon, and you admitted you didn't recall how it used to be. Well, it used to not turn off the screen until IDLE_TIME_OFF.
So currently, both PLI or idle.bmp disappear after font_config.ProffieOSFontImageDuration (default at 5 seconds).

This change would assume that if we have idle.bmp, it's there to be viewed longer than that.
It also retains the current behavior if PLI is shown, with the option to extend that as well.
No good?

profezzorn:
Idle.bmp should not disappear after 5 seconds. It's supposed to just stay on. If that's not happening, that's a bug.
(Unless you have a non-looped idle animation with a duration of 5 seconds.)

NoSloppy:
Apologies. I guess I saw incorrectly. idle image does indeed stay on.
Removed that condition, leaving the option to adjust PLI time out.
Changed the define to something more accurate like PLI_OFF_TIME?

PR #376, opened 2021-12-30
add default font .ini files to /doc

NoSloppy:


profezzorn:
Don't we have a wiki page for this already?
Any particular reason it should also be in /doc/ ?

NoSloppy:
The only particular reason I figure is users definitely have the OS download, while they might not be privy to the Wiki unless prompted. Just a convenience I suppose. If it's a no-go, that's totally fine in my book.
Would it make sense putting it in the default SD content instead then? The current versions of config.ini are literally just a humstart entry.

profezzorn:
In general I find that having information in multiple places means that at least of them is going to be out of date.
That doesn't mean we can't have information in more than one place, it just means that I want reasons before doing it.
Perhaps a better way to do this would be to make a script that pulls the WIKI pages and convert them to a PDF or something when the zip files are created.

NoSloppy:
> make a script that pulls the WIKI pages and convert them to a PDF or something when the zip files are created.

Ok. That is beyond my capability. I already updated the Wiki pages, so there we are for now I suppose.

PR #380, opened 2022-01-05
CONFIG_FILE guidance idea

NoSloppy:


ProffieOS.ino
```diff
@@ -16,10 +16,16 @@
 
  You should have received a copy of the GNU General Public License
  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+-------------------------------------------------------------------
+You can have multiple configuration files, and specify which one
```
profezzorn, line 22: I think I like the text update, but I think the copyright blurb should be in a separate comment.

NoSloppy: ok how about now?

ProffieOS.ino
```diff
@@ -18,8 +18,14 @@
...
 */
 
-// You can have multiple configuration files, and specify which one
-// to use here.
+/*-------------------------------------------------------------------
+-  You can have multiple configuration files, and specify which one -
+-  to use here by removing the comment slashes at the beginning.    -
+-  **NOTE** Only ONE line should be left uncommented at a time!     -
+-  Add the slashes to any that you are not uploading.               -
+*/-------------------------------------------------------------------
```
profezzorn, line 26: Does this compile?

NoSloppy: Embarrassingly, no.
It does now though.

ProffieOS.ino
```diff
@@ -18,8 +18,16 @@
...
 */
 
-// You can have multiple configuration files, and specify which one
-// to use here.
+/*
+---------------------------------------------------------------------
+-  You can have multiple configuration files, and specify which one -
+-  to use here by removing the comment slashes at the beginning.    -
+-  **NOTE** Only ONE line should be left uncommented at a time!     -
+-  Add the slashes to any that you are not uploading.               -
```
profezzorn, line 26: uploading -> using

ProffieOS.ino
```diff
@@ -18,8 +18,16 @@
  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-// You can have multiple configuration files, and specify which one
-// to use here.
+/*
+---------------------------------------------------------------------
+-  You can have multiple configuration files, and specify which one -
+-  to use here by removing the comment slashes at the beginning.    -
```
profezzorn, line 24: comment slashes -> two slashes

ProffieOS.ino
```diff
@@ -18,8 +18,16 @@
  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-// You can have multiple configuration files, and specify which one
-// to use here.
+/*
+---------------------------------------------------------------------
```
profezzorn, line 22: How about this style?

```
/*---------------------------*\
| text in goes in here        |
\*---------------------------*/

```

I also like this style:

```
/*\
| |  Some text goes here
\*/

```

NoSloppy: ooo fancy, nice! 
How about the Version included as shown so users can confirm the OS?
I always throw that in my personal version.
Although, the master branch isn't really......

ProffieOS.ino
```diff
@@ -18,14 +18,12 @@
...
 
-/*
----------------------------------------------------------------------
--  You can have multiple configuration files, and specify which one -
--  to use here by removing the comment slashes at the beginning.    -
--  **NOTE** Only ONE line should be left uncommented at a time!     -
--  Add the slashes to any that you are not uploading.               -
----------------------------------------------------------------------
-*/
+/*---------------------- ProffieOS:6.3 ----------------------------*\
```
profezzorn, line 21: The version string on line 209 is updated automatically by the export script.
I don't want to have a version that I have to update manually.
IF we put the version here, it would look like $Id: ..... $ and when you check out the code from github you would only get a long hex string, not anything helpful. Also, the version string changes length, which would ruin all this fancy formatting....

NoSloppy: That's what i was referring to, yes. Ok cool, reverted.

PR #384, opened 2022-01-10
Kill hum if all else fails

NoSloppy:
Without this fall through, hum continues after blade off if no retraction sounds exist.

### DIFF
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -470,9 +470,9 @@ public:
             hum_player_->FadeAndStop();
             hum_player_.Free();
           } else {
-          state_ = STATE_HUM_FADE_OUT;
-          hum_fade_out_ = 0.2;
-        }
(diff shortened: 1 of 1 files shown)


sound/hybrid_font.h
```diff
@@ -469,7 +469,10 @@ class HybridFont : public SaberBase, public Looper {
             hum_player_->set_fade_time(0.2);
             hum_player_->FadeAndStop();
             hum_player_.Free();
-          }
+          } else {
+          state_ = STATE_HUM_FADE_OUT;
```
profezzorn, line 473: Fix looks reasonable, but indentation looks bad.
Could be github messing up, please check.

NoSloppy: Indeed. It looks ok online, but it's weird when pulled down.
All the adds you did look off too. Weird. Fixed.
Screen Shot 2022-01-10 at 8 57 44 PM

PR #385, opened 2022-01-10
fix for saber_off_time_millis_

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -5621,7 +5623,7 @@ SaberFett263Buttons() : PropBase() {}
   uint32_t clash_impact_millis_; // Clash timer
   uint32_t last_push_millis_; // Last Push (to prevent gesture spamming)
   uint32_t last_blast_millis_; // Last Blast (for Battle Mode Multi-Blast detection)
-  uint32_t saber_off_time_millis_; // Off timer
+  uint32_t saber_off_time_millis_ = 2100; // Off timer
```
profezzorn, line 5626: This might not be a good idea, because it means that for the first 2.1 seconds, millis() - saber_off_time_millis_ will return ~4 billion.

fdarosa2663: noted, I'll look at another way to handle.

fdarosa2663: removed the value, will look at improvements in the future

PR #388, opened 2022-01-17
Pov tools update 2 pov.h

NoSloppy:


ProffieOS.ino
```diff
@@ -417,6 +417,10 @@ struct is_same_type { static const bool value = false; };
 template<class T>
 struct is_same_type<T, T> { static const bool value = true; };
 
+#define CONFIG_POV
```
profezzorn, line 420: This isn't backwards compatible.
Also, the only reason for having a section for it would be if people wanted to put their POV file in the actual config file, which seems like a bad idea.

I think it's better to just use a define, like:

#ifdef POV_INCLUDE_FILE
#include POV_INCLUDE_FILE
#else
#include "star_wars_logo_pov_data.h"
#endif

NoSloppy: Oh! This is lovely. cool.

NoSloppy:
Resubmitting

PR #389, opened 2022-01-17
Update blaster.h

NoSloppy:
Don't Jam if empty.
Autofire now counts bullets (uses WavLen of auto.wav per shot ! )
If already full, Reload plays full.wav.

### DIFF
--- a/props/blaster.h
+++ b/props/blaster.h
@@ -117,40 +117,6 @@ public:
     if (max_shots_ != -1 && shots_fired_ >= max_shots_) {
       empty_ = true;
       SaberBase::DoEffect(EFFECT_EMPTY, 0);
-      return;
-    }
-  }
-
-  uint32_t auto_time_;
-  virtual void Fire() {
-#ifdef ENABLE_MOTION
-#ifdef BLASTER_JAM_PERCENTAGE
-    // If we're already jammed then we don't need to recheck. If we're not jammed then check if we just jammed.
-    is_jammed_ = is_jammed_ ? true : CheckJam(BLASTER_JAM_PERCENTAGE);
-    if (is_jammed_ && !empty_) {
-      // Don't jam if empty
-      SaberBase::DoEffect(EFFECT_JAM, 0);
-      return;
-    }
-#endif
-#endif
-    CheckEmpty();
-    if (empty_) return;
-    if (blaster_mode == MODE_AUTO) {
-      SelectAutoFirePair(); // Set up the autofire pairing if the font suits it.
-      SaberBase::SetLockup(LOCKUP_AUTOFIRE);
-      SaberBase::DoBeginLockup();
-      auto_firing_ = true;
-      auto_time_ = millis();
-    } else {
-      if (blaster_mode == MODE_STUN) {
-        SaberBase::DoEffect(EFFECT_STUN, 0);
-        STDOUT << "******** STUN - Remaining shots = " << GetBulletCount() << "\n";
-      } else {
-        SaberBase::DoEffect(EFFECT_FIRE, 0);
-        STDOUT << "******** FIRE - Remaining shots = " << GetBulletCount() << "\n";
-      }
-      shots_fired_++;
     }
   }
 
@@ -241,7 +207,6 @@ public:
     }
   }
 
-  uint32_t len;
   void beginArm() {
     SaberBase::SetLockup(SaberBase::LOCKUP_ARMED);
     SaberBase::DoBeginLockup();
@@ -270,14 +235,54 @@ public:
 
   RefPtr<BufferedWavPlayer> auto_player_;
 
+  uint32_t auto_time_;
+  virtual void Fire() {
+    CheckEmpty();
+    if (empty_) return;
+#ifdef ENABLE_MOTION
+#ifdef BLASTER_JAM_PERCENTAGE
+    // If we're already jammed then we don't need to recheck. If we're not jammed then check if we just jammed.
+    is_jammed_ = is_jammed_ ? true : CheckJam(BLASTER_JAM_PERCENTAGE);
+    if (is_jammed_) {
+      // Don't jam if empty
+      SaberBase::DoEffect(EFFECT_JAM, 0);
+      return;
+    }
+#endif
+#endif
+    if (blaster_mode == MODE_AUTO) {
+      SelectAutoFirePair(); // Set up the autofire pairing if the font suits it.
+      SaberBase::SetLockup(LOCKUP_AUTOFIRE);
+      SaberBase::DoBeginLockup();
+      auto_firing_ = true;
+      auto_time_ = millis();
+
+      // GET auto.wav length here
+        if (!auto_player_) {
+        auto_player_ = GetWavPlayerPlaying(&SFX_auto);
+        }
+
+    } else {
+      if (blaster_mode == MODE_STUN) {
+        SaberBase::DoEffect(EFFECT_STUN, 0);
+        STDOUT << "******** STUN - Remaining shots = " << GetBulletCount() << "\n";
+      } else {
+        SaberBase::DoEffect(EFFECT_FIRE, 0);
+        STDOUT << "******** FIRE - Remaining shots = " << GetBulletCount() << "\n";
+      }
+      shots_fired_++;
+    }
+  }
+  
+  RefPtr<BufferedWavPlayer> auto_player_;
+
   void Loop() override {
     PropBase::Loop();
     PollNextAction();
-    if (auto_firing_ && GetWavPlayerPlaying(&SFX_auto)) {
-      if (!auto_player_) {
-        auto_player_ = GetWavPlayerPlaying(&SFX_auto);
-      }
+    if (auto_firing_) {
+      // do stuff
       if (millis() - auto_time_ > 1000 * auto_player_->length()) {
+        // NULL POINTER because it doesn't know auto_player_ ?
         shots_fired_++;
         auto_time_ = millis();
         STDOUT << "******** AUTOFIRING - Remaining shots = " << GetBulletCount() << "\n";
@@ -286,6 +291,7 @@ public:
           SaberBase::DoEndLockup();
           SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
           auto_firing_ = false;
+          auto_player_.Free();
           return;
         }
       }
--- a/props/blaster.h
+++ b/props/blaster.h
@@ -278,7 +278,6 @@ public:
     PollNextAction();
     if (auto_firing_) {
       if (&auto_player_) {
-        STDOUT << millis() - auto_time_;
         if (millis() - auto_time_ > 1000 * auto_player_->length()) {
           shots_fired_++;
           auto_time_ = millis();
--- a/props/blaster.h
+++ b/props/blaster.h
@@ -233,8 +233,6 @@ public:
     }
   }
 
-  RefPtr<BufferedWavPlayer> auto_player_;
-
   uint32_t auto_time_;
   virtual void Fire() {
     CheckEmpty();
--- a/props/blaster.h
+++ b/props/blaster.h
@@ -54,7 +54,7 @@ EFFECT(mdauto);
 // For mode sounds, specific "mdstun", "mdkill", and "mdauto" may be used.
 // If just a single "mode" sound for all switches exists, that will be used.
 // If no mode sounds exist in the font, a talkie version will speak the mode on switching.
-class Blaster : public PROP_INHERIT_PREFIX PropBase, StateMachine {
+class Blaster : public PROP_INHERIT_PREFIX PropBase {
 public:
   Blaster() : PropBase() {}
   const char* name() override { return "Blaster"; }
@@ -145,10 +145,10 @@ public:
     } else {
       if (blaster_mode == MODE_STUN) {
         SaberBase::DoEffect(EFFECT_STUN, 0);
-        STDOUT << "STUN - Remaining shots = " << GetBulletCount() << "\n";
+        STDOUT << "******** STUN - Remaining shots = " << GetBulletCount() << "\n";
       } else {
         SaberBase::DoEffect(EFFECT_FIRE, 0);
-        STDOUT << "FIRE - Remaining shots = " << GetBulletCount() << "\n";
+        STDOUT << "******** FIRE - Remaining shots = " << GetBulletCount() << "\n";
       }
       shots_fired_++;
     }
@@ -184,6 +184,7 @@ public:
   virtual void ClipIn() {
     SaberBase::DoEffect(EFFECT_CLIP_IN, 0);
   }
+
   // Pull in parent's SetPreset, but turn the blaster on.
   void SetPreset(int preset_num, bool announce) override {
     PropBase::SetPreset(preset_num, announce);
@@ -239,7 +240,8 @@ public:
       next_action_ = NEXT_ACTION_NOTHING;
     }
   }
-  
+
+  uint32_t len;
   void beginArm() {
     SaberBase::SetLockup(SaberBase::LOCKUP_ARMED);
     SaberBase::DoBeginLockup();
@@ -266,39 +268,28 @@ public:
     }
   }
 
-  RefPtr<BufferedWavPlayer> tmp;
-  uint32_t len;
+  RefPtr<BufferedWavPlayer> auto_player_;
 
   void Loop() override {
     PropBase::Loop();
     PollNextAction();
-	  
-    STATE_MACHINE_BEGIN();
-    while (true) {
-      if (auto_firing_) {
-        tmp = GetWavPlayerPlaying(&SFX_auto);
-        // Set the length for WavLen<>
-        if (tmp) {
-          tmp->UpdateSaberBaseSoundInfo();
-        } else {
-          SaberBase::ClearSoundInfo();
-        }
-        if (millis() - auto_time_ > 1000 * tmp->length()) {
-          shots_fired_++;
-          auto_time_ = millis();
-          STDOUT << "AUTOFIRING - Remaining shots = " << GetBulletCount() << "\n";
-          CheckEmpty();
-          if (empty_) {
-            SaberBase::DoEndLockup();
-            SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
-            auto_firing_ = false;
-            return;
-          }
+    if (auto_firing_ && GetWavPlayerPlaying(&SFX_auto)) {
+      if (!auto_player_) {
+        auto_player_ = GetWavPlayerPlaying(&SFX_auto);
+      }
+      if (millis() - auto_time_ > 1000 * auto_player_->length()) {
+        shots_fired_++;
+        auto_time_ = millis();
+        STDOUT << "******** AUTOFIRING - Remaining shots = " << GetBulletCount() << "\n";
+        CheckEmpty();
+        if (empty_) {
+          SaberBase::DoEndLockup();
+          SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+          auto_firing_ = false;
+          return;
         }
       }
-    YIELD();
     }
-    STATE_MACHINE_END();
   }
 
   // Make clash do nothing except unjam if jammed.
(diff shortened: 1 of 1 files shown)


props/blaster.h
```diff
@@ -260,9 +266,39 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
   }
 
+  RefPtr<BufferedWavPlayer> tmp;
+  uint32_t len;
+
   void Loop() override {
     PropBase::Loop();
     PollNextAction();
+	  
+    STATE_MACHINE_BEGIN();
```
profezzorn, line 276: You don't need a state machine here.

Your state machine runs one loop every time Loop() is called, which is what the code would do even if it wasn't in a state machine.

props/blaster.h
```diff
@@ -260,9 +266,39 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+    while (true) {
+      if (auto_firing_) {
+        tmp = GetWavPlayerPlaying(&SFX_auto);
+        // Set the length for WavLen<>
+        if (tmp) {
+          tmp->UpdateSaberBaseSoundInfo();
```
profezzorn, line 282: This function call only makes sense in SB_* callbacks.
Calling it from Loop() is weird and is likely to cause problems.

props/blaster.h
```diff
@@ -260,9 +266,39 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+  uint32_t len;
+
   void Loop() override {
     PropBase::Loop();
     PollNextAction();
+	  
+    STATE_MACHINE_BEGIN();
+    while (true) {
+      if (auto_firing_) {
+        tmp = GetWavPlayerPlaying(&SFX_auto);
```
profezzorn, line 279: calling this every loop is a bit expensive.
it would be better if we got the length of the wav file right after we start playing the sound.

props/blaster.h
```diff
@@ -260,9 +266,39 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+        // Set the length for WavLen<>
+        if (tmp) {
+          tmp->UpdateSaberBaseSoundInfo();
+        } else {
+          SaberBase::ClearSoundInfo();
+        }
+        if (millis() - auto_time_ > 1000 * tmp->length()) {
```
profezzorn, line 286: A simpler way to do this might be to use tmp->pos().
If you keep track of the previous position, then you can just do:

float pos = tmp->pos();
if (pos < last_pos_) {
    // sound has wrapped around
}
last_pos_ = pos;

props/blaster.h
NoSloppy: Doesn't that happen elsewhere though? All that's here is DoBeginLockup() and SFX_auto isn't playing yet.

props/blaster.h
NoSloppy: trying this, but can't get   tmp = GetWavPlayerPlaying(&SFX_auto); unless it's inside the loop.
Can't pass 'pos' , running into a wall tonight.

props/blaster.h
profezzorn: Well, we can keep the length, or the player, basically something like:

if (auto_firing_) {
  if (!auto_player_) {
     auto_player_ = GetWavPlayerPlaying(&SFX_auto);
  }
  // rest of logic here
} else {
   auto_player_.reset();
}

props/blaster.h
profezzorn: not sure what this comment means

props/blaster.h
NoSloppy: ah. I had `if (auto_firing_ && GetWavPlayerPlaying(&SFX_auto)) {
`
I will try this, thanks.

NoSloppy: Actually, that avoids NULL POINTER while bgnauto plays.
Well, latest update works, but does it make sense?

props/blaster.h
```diff
@@ -93,9 +93,10 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
     }
   }
 
-  bool auto_firing_ = false;
   int shots_fired_ = 0;
+  bool auto_firing_ = false;
```
profezzorn, line 97: What's the point of moving this one line?

props/blaster.h
```diff
@@ -112,38 +113,43 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
     return random <= percent ? true : false;
   }
 
+  virtual void CheckEmpty() {
+    if (max_shots_ != -1 && shots_fired_ >= max_shots_) {
+      empty_ = true;
+      SaberBase::DoEffect(EFFECT_EMPTY, 0);
+      return;
```
profezzorn, line 120: what's the point of having a "return" here?

props/blaster.h
```diff
@@ -112,38 +113,43 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+
+  uint32_t auto_time_;
   virtual void Fire() {
 #ifdef ENABLE_MOTION
 #ifdef BLASTER_JAM_PERCENTAGE
     // If we're already jammed then we don't need to recheck. If we're not jammed then check if we just jammed.
     is_jammed_ = is_jammed_ ? true : CheckJam(BLASTER_JAM_PERCENTAGE);
-
-    if (is_jammed_) {
+    if (is_jammed_ && !empty_) {
```
profezzorn, line 130: I don't think we whould set is_jammed_ either if it's empty.
In fact, maybe the the jam check should be after the empty check below...

props/blaster.h
```diff
@@ -233,7 +240,8 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
       next_action_ = NEXT_ACTION_NOTHING;
     }
   }
-  
+
+  uint32_t len;
```
profezzorn, line 244: This name is too short for a global variable.
What is it for anyways?
Also, global variables should end in _.

props/blaster.h
```diff
@@ -260,9 +268,28 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
     }
   }
 
+  RefPtr<BufferedWavPlayer> auto_player_;
+
   void Loop() override {
     PropBase::Loop();
     PollNextAction();
+    if (auto_firing_ && GetWavPlayerPlaying(&SFX_auto)) {
```
profezzorn, line 276: You're still running GetWavPlayerPlaying() every loop.
Also, the logic here seems a bit wonky. It ought to be something like:

if (auto_firing_) {
  if (!auto_plaeyr_) {
    auto_player_ = GetWavPlayerPlaying(&SFX_auto);
  }
  if (auto_player_) {
      // do stuff
  }
} else {
  auto_player_.Reset();
}

profezzorn: Note that the reset is important, otherwise it will keep the same player between auto-firing sessions.

props/blaster.h
NoSloppy: I just grouped the bools together I guess.

props/blaster.h
NoSloppy: So we don't continue on and fire anyway if we're in MODE_AUTO?

props/blaster.h
NoSloppy: So can I move that block into Fire() in the MODE_AUTO part? I don't know why I'm in Loop. I guess leftover from incorrect State Mchine logic.
Trying that, but Reset() makes it unhappy:
`class RefPtr<BufferedWavPlayer>' has no member named 'Reset'

props/blaster.h
NoSloppy: An errant remnant. Removed.

props/blaster.h
NoSloppy: yes.

props/blaster.h
profezzorn: I think I meant Free(), not Reset().

profezzorn: Some of this needs to be in loop() to happen over and over again.
However, setting and unsetting the auto_player_ might be better to put elsewhere.

NoSloppy: ok thanks. Seems the problem keeps happening if i move the wav length detection out of loop.
It doesn't pickup the global auto_player_ inside loop. 
I get NULL POINTER when checking elapsed time against length.
Submitted changes but not working yet.

NoSloppy: Leaning toward making a new `Function()`, doing
`auto_player_ = GetWavPlayerPlaying(&SFX_auto)` in there, 
then in the loop, get the length with
`1000 * Function()->length()`
but what type is auto_player?
That would need a `return *type*` yes?

NoSloppy: @profezzorn The best I can determine with limited understanding is that in order to remove `auto_player_ = GetWavPlayerPlaying()` out of loop, auto_player_ needs to be a global variable or pointer. I'm not sure where that gets done, and if it's making a new function to do so, if it gets an * or $ or... beyond my level on comprehension at the moment.
I DO know that the code works if it's all in loop. 
Just trying to understand the issue with that ..."too expensive" I guess I get that it's too many calls for something that can be had once elsewhere.
But I'll need specific help on this as I've exhausted my ability at the moment.

NoSloppy: Confirming I can't get auto_player_ to be recognized inside Loop() if it's declared outside.
I'm sure it's something simple I just don't have a grasp on yet.

NoSloppy: Ok Progress...sort of.
Using &auto_player_, the RefPtr works inside Loop().
However, the auto.wav is played stretched out, but not like 1/2 samplerate, just time stretched to what sounds like 2x duration to me.
Any idea what I did?
Committed.

props/blaster.h
```diff
@@ -260,9 +233,65 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+      }
+      shots_fired_++;
+    }
+  }
+
   void Loop() override {
     PropBase::Loop();
     PollNextAction();
+    if (auto_firing_) {
+      if (&auto_player_) {
```
profezzorn, line 280: This takes the address of "auto_player_", which is always true.
Just do if(auto_player_)

props/blaster.h
```diff
@@ -260,9 +233,65 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+    if (blaster_mode == MODE_AUTO) {
+      // Set up the autofire pairing if the font suits it.
+      SelectAutoFirePair(); 
+      SaberBase::SetLockup(LOCKUP_AUTOFIRE);
+      SaberBase::DoBeginLockup();
+      auto_firing_ = true;
+      auto_time_ = millis();
+
+      // GET auto.wav here
+      if (!auto_player_) {
```
profezzorn, line 261: I think if auto_player_ is set here, we need to override it with the new one.

props/blaster.h
```diff
@@ -260,9 +233,65 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+        auto_player_ = GetWavPlayerPlaying(&SFX_auto);
```
profezzorn, line 262: If there happens to be more than one SFX_auto running, this might end up finding the wrong one. Maybe we should find the one that was started most recently?

props/blaster.h
```diff
@@ -260,9 +233,65 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+      if (&auto_player_) {
+        if (millis() - auto_time_ > 1000 * auto_player_->length()) {
+          shots_fired_++;
+          auto_time_ = millis();
+          STDOUT << "******** AUTOFIRING - Remaining shots = " << GetBulletCount() << "\n";
+          CheckEmpty();
+          if (empty_) {
+            SaberBase::DoEndLockup();
+            SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+            auto_firing_ = false;
```
profezzorn, line 289: We should call free on the auto_player_ here.

props/blaster.h
NoSloppy: But that's not in the scope. That's what i kept running up against. Not sure how to do that, spent hours :(

props/blaster.h
NoSloppy: Shoot. I had it there but thought you preferred it not in Loop(), so I put it on the button release. 
OK.

props/blaster.h
profezzorn: Why do you think it's not in scope?
Do you get a compiler error?
What is that error?

props/blaster.h
profezzorn: It should be everywhere where we finish auto fire mode.

NoSloppy: yup. done.

props/blaster.h
NoSloppy: so then just remove the if clause it seems.....

props/blaster.h
NoSloppy: > Why do you think it's not in scope? Do you get a compiler error? What is that error?

Well, of course it's working fine now! 
Man, over and over I couldn't get it to work before. No error, it just never behaved as if the if clause was true.
I put an `else` on that `if (auto_player_)` to print a message " NOPE! no auto_player_"
And that is all I would get, the counting bullets and checking empty part wasn't happening.
Must have learned or done something right overnight I guess. Like moving the declaration ouf of the Fire() function out to global scope. Anyway, MOVING ON. 
Thank you.

props/blaster.h
NoSloppy: ok

NoSloppy: Having a tough time with filtering only players playing auto.wav. Do we get &SFX_auto during bgnlock playing, or do we need to wait to check until after bgnlock is done?
I keep coming up with an empty hook when fishing for a matching player.

NoSloppy: commit named filter help needed

NoSloppy: Is this really needed? When would there be more than one SFX_auto running that would confise GetWavPlayerPlaying(&SFX_auto) ?

props/blaster.h
```diff
@@ -93,9 +93,10 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
     }
   }
 
-  bool auto_firing_ = false;
   int shots_fired_ = 0;
+  bool auto_firing_ = false;
   bool is_jammed_ = false;
+  bool empty_ = false;
```
profezzorn, line 99: I don't think we need this variable.

I think we want something like this instead:

bool IsEmpty() const {
  return max_shosts_ != -1 && shots_fired_ >= max_shots_;
}

Then instead of checking empty_, we'll just call IsEmpty()

props/blaster.h
```diff
@@ -112,39 +113,10 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
-      auto_firing_ = true;
-    } else {
-      if (blaster_mode == MODE_STUN) {
-        SaberBase::DoEffect(EFFECT_STUN, 0);
-      } else {
-        SaberBase::DoEffect(EFFECT_FIRE, 0);
-      }
-
-      shots_fired_++;
+  virtual void CheckEmpty() {
```
profezzorn, line 116: might be helpful if CheckEmpty() returns true/false so you can just do:

if (CheckEmpty()) return;

props/blaster.h
```diff
@@ -164,8 +136,10 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
   }
 
   virtual void Reload() {
-    shots_fired_ = 0;
     SaberBase::DoEffect(EFFECT_RELOAD, 0);
+    shots_fired_ = 0;
+    empty_ = false;
+    is_jammed_ = false;
```
profezzorn, line 142: This is new, and I'm not sure if it's the way people want it or not.
Should reloading really unjam the gun too?

props/blaster.h
```diff
@@ -260,9 +233,92 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+  uint32_t auto_time_;
+  virtual void Fire() {
+    CheckEmpty();
+    if (empty_) return;
+#ifdef ENABLE_MOTION
+#ifdef BLASTER_JAM_PERCENTAGE
+    // If we're already jammed then we don't need to recheck. If we're not jammed then check if we just jammed.
+    is_jammed_ = is_jammed_ ? true : CheckJam(BLASTER_JAM_PERCENTAGE);
+    if (is_jammed_) {
+      SaberBase::DoEffect(EFFECT_JAM, 0);
```
profezzorn, line 247: Seems like there should be a different effect for when a gun jams and when you try to fire a gun which is already jammed? But maybe that's just wishful thinking?

props/blaster.h
```diff
@@ -260,9 +233,92 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+#endif
+#endif
+    if (blaster_mode == MODE_AUTO) {
+      // Set up the autofire pairing if the font suits it.
+      SelectAutoFirePair();
+      auto_firing_ = true;
+      auto_time_ = millis();
+      SaberBase::SetLockup(LOCKUP_AUTOFIRE);
+      SaberBase::DoBeginLockup();
+      STDOUT << "***************************** OK Going to GetNewestAutoWav() with &SFX_auto in tow\n";
```
profezzorn, line 259: These look like debug printouts.
I don't mind having a few printouts in here, but these are large, "loud" and low on information.

props/blaster.h
```diff
@@ -260,9 +233,92 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+#ifdef BLASTER_JAM_PERCENTAGE
+    // If we're already jammed then we don't need to recheck. If we're not jammed then check if we just jammed.
+    is_jammed_ = is_jammed_ ? true : CheckJam(BLASTER_JAM_PERCENTAGE);
+    if (is_jammed_) {
+      SaberBase::DoEffect(EFFECT_JAM, 0);
+      return;
+    }
+#endif
+#endif
+    if (blaster_mode == MODE_AUTO) {
```
profezzorn, line 252: seems like this should be a switch, since there are more than two choices.

props/blaster.h
```diff
@@ -260,9 +233,92 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+        STDOUT << "******** STUN - Remaining shots = " << GetBulletCount() << "\n";
+      } else {
+        SaberBase::DoEffect(EFFECT_FIRE, 0);
+        STDOUT << "******** FIRE - Remaining shots = " << GetBulletCount() << "\n";
+      }
+      shots_fired_++;
+    }
+  }
+
+  RefPtr<BufferedWavPlayer> GetNewestAutoWav(Effect* a) {
```
profezzorn, line 275: This is super confusing, because it looks like a "find a wav player and return it", but in fact it sets auto_player_ AND returns it, but then the return value is ignored where you call it.
I would recommend not setting auto_player_ inside this function, just return it and let the caller do the assignment.

props/blaster.h
```diff
@@ -260,9 +233,92 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+    STDOUT << "***************************** for-loop done, chosen player = " << player_to_choose << " \n";
+    auto_player_ = RefPtr<BufferedWavPlayer>(wav_players + player_to_choose);
+    return auto_player_;
+  }
+
+
   void Loop() override {
     PropBase::Loop();
     PollNextAction();
+    if (auto_firing_) {
```
profezzorn, line 305: if (A) {
  if (B) {
  }
}

should almost always be written as:
  if (A && B) {
  }

props/blaster.h
```diff
@@ -260,9 +233,92 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
+      if (auto_player_) {
+        if (millis() - auto_time_ > 1000 * auto_player_->length()) {
```
profezzorn, line 307: Why don't we check if it's looped around instead?
all we need is something like:

float pos = auto_player_->pos();
if (pos < last_pos_) {
   shots_fired_++;
   // etc.
}
last_pos_ = pos;

profezzorn:
Is this still a thing, or should I close the PR?

NoSloppy:
I have a working personal version, but I think you disliked some stuff about it. This PR is a piecemeal, trying to make it better and still work, but I think I hit a wall. Might need some help, might be better to submit my "I'm happy with it" version in a separate PR then compare and combine what can be used and what should change. It's going to take a bit of effort. Time is slim these weeks, so asap, I'd love to work on this again.

TL;DR yes still a thing.

props/blaster.h
NoSloppy: I think this part can somehow re-use the code now used for KILL_OLD_PLAYERS.
I'll try to wrap my head around this maybe this week.

NoSloppy:
So I think it might be best to leave the default blaster prop as is, and just make customizations to a new file altogether and call it blaster_BC_buttons, just like how we do for saber prop files.
That way, all changes can be made and reviewed in one place, and options and features included aren't "forced" on everybody

PR #390, opened 2022-01-17
Add blaster_BC_buttons.h

NoSloppy:


common/saber_base.h
```diff
@@ -52,6 +52,7 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(UNJAM)			\
     DEFINE_EFFECT(PLI_ON)			\
     DEFINE_EFFECT(PLI_OFF)                      \
+    DEFINE_EFFECT(DESTRUCT)			\
```
profezzorn, line 55: Shouldn't this be OFF_BLAST instead?

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,807 @@
...
+EFFECT(vmbegin);    // for Begin Volume Menu
+EFFECT(vmend);      // for End Volume Menu
+EFFECT(volup);      // for increse volume
+EFFECT(voldown);    // for decrease volume
+EFFECT(volmin);     // for minimum volume reached
+EFFECT(volmax);     // for maximum volume reached
+EFFECT(quote);      // for playing quotes
+#endif
+
+class BlasterBC : public virtual PropBase, StateMachine {
```
profezzorn, line 160: 1. don't need state machine (see other review)
2. Can we inherit from blaster instead of PropBase to reduce duplication?
3. don't use "virtual" directly, use the define.
4.

profezzorn (commented):
More comments later...

common/saber_base.h
NoSloppy: Well the idea is to be an EFFECT that can be used as a blade animation.
Plus play an SFX_destruct sound. Sounds ends, boom.

props/blaster_BC_buttons.h
NoSloppy: 1. State Machine removed.
2. I'm not sure why or how to do that.
3. I tried to follow this clue several ways with no luck.

profezzorn:
I'll work more on this review after the regular blaster one is done

PR #394, opened 2022-01-18
pov data save directory direct to /styles

NoSloppy:


NoSloppy:
Done in script. Maybe a good idea for the default Makefile anyway, but closing for now.

PR #397, opened 2022-01-22
Pov tools script and read me

NoSloppy:


pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+preset=0
+# get quantity of .png files and build array to print if multiples
+files=$(ls 2>/dev/null -Ubad1 -- *.png | wc -l)
+pngs=$(find . -type f -name '*.png')
+
+if [ "$files" -ge 2 ]; then
+	echo "OOPS! Too many .png image files found!"
+	echo ""
+	for strip in ${pngs[@]}; do
+		no_path="${strip##*/}"
```
profezzorn, line 21: there no need for the "no_path" variable, you could just do:

echo "${strip##*/}"

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+cd "$(dirname "$0")"
+clear
+echo ""; echo "Welcome! Let's make some POV images!"; echo ""
+# replace whitespace in source filename with underscores
+find . -type f -name "* *" | while read file; do mv "$file" ${file// /_}; done
+have_logo=0
+height_only=0
+preset=0
+# get quantity of .png files and build array to print if multiples
+files=$(ls 2>/dev/null -Ubad1 -- *.png | wc -l)
```
profezzorn, line 14: -U (sorting) is not needed, it's also not POSIX, so it might not work with all "ls" programs.
-b (quoting) is also not needed, nor POSIX
-a is for listing .-files, which is not needed

2>/dev/null is also not needed afaik.

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+			exec "./create_POV_data_files";;
+		q) exit;;
+		*) exec "./create_POV_data_files";;
+	esac
+fi
+
+if [ "$files" -lt 1 ]; then
+	echo "OOPS! No .png image files found!"
+	echo ""
+	echo "Let's download an example Star Wars Logo for you to try :)"
```
profezzorn, line 47: Seems silly to to this just because there was no files.
Maybe ask first?

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+	if [ "$delay" = "q" ]; then exit; fi
+fi
+
+echo ""
+case $have_logo in
+	1) echo "Found the following source image file: $logo"
+		echo ""
+		sourcebase="${logo%.*}"
+		image="$logo";;
+	0)  source=$(find . -type f -name '*.png')
```
profezzorn, line 64: isn't "source" the same as "pngs" ?

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+fi
+
+if [ "$files" -lt 1 ]; then
+	echo "OOPS! No .png image files found!"
+	echo ""
+	echo "Let's download an example Star Wars Logo for you to try :)"
+	echo "DOWNLOADING: 1024px-Star_Wars_Logo.svg.png .........."
+	echo ""
+	wget https://upload.wikimedia.org/wikipedia/commons/thumb/6/6c/Star_Wars_Logo.svg/1024px-Star_Wars_Logo.svg.png&>/dev/null &
+	logo=$"1024px-Star_Wars_Logo.svg.png"
```
profezzorn, line 51: Why not just set "pngs" here?

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+    echo "2 - Preset 2 (HEIGHT=97,  LENGTH=63, OFFSET=50)"
+    echo "3 - Preset 3 (HEIGHT=72,  LENGTH=50, OFFSET=15)"
+    echo "4 - Preset 4 (HEIGHT=72,  LENGTH=50, OFFSET=39)"
+    echo "5 - Height entry only"
+    echo "6 - Custom options"
+    echo "q - to quit"
+	read option
+	case $option in
+		"")	echo "! Some option is needed. Please try again"; echo "";;
+		1)	make IMAGE=$image OPTIONS="--height=144 --length=39 --offset=18"; HEIGHT=144 preset=1 valid_option=1;;
```
profezzorn, line 101: I think it's weird to have several "make" invocations in this loop.
Why not have variables like "height", "width" and "length", have the loop adjust them, and once you exit the loop we run make IMAGE="$image" OPTIONS="--height=$height --length=$length offset=$offset"

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+do
+	echo ""
+	echo "How many pixels are in your blade? (Press ENTER for default 144) "
+	read  height
+	if [[ $height == "" ]]; then height=144 length=39 offset=18 validinput=1; echo "Default 144"
+	elif [[ $height == "q" ]]; then exit
+	elif [[ $height -gt 264 ]] && [[ $height -le 576 ]]; then
+		read -p "Are you sure there's that many pixels?? (y/n)" too_many
+		if [[ $too_many == "y" ]]; then validinput=1; else echo ""; fi
+	elif [[ $height -gt 576 ]]; then
```
profezzorn, line 132: If you check -gt 576 first, then you don't have to check it again like you do on line 129

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+        echo ""
+        echo "What is your preferred unit of measuring distance? (i - inches, c - cm) "
+        read pref
+        case $pref in
+            "") echo "! Some option is needed. Please try again"; echo "";;
+            i)  calc_len=$(printf %.0f "$((10**3 * $height * 100 / 144 * 100 / 254))e-3"); offset_default=18; validinput=1; units="inches"; echo "";;
+            c)  calc_len=$(printf %.0f "$((10**3 * $height * 100 / 144))e-3"); offset_default=46; validinput=1; units="centimeters"; echo "";;
+            *[q*]*) exit;;
+            *)  echo ""
+				echo "! Not a valid option, Please choose inches or cm,"
```
profezzorn, line 152: indented weirdly

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+	esac
+fi
+
+if [ "$files" -lt 1 ]; then
+	echo "OOPS! No .png image files found!"
+	echo ""
+	echo "Let's download an example Star Wars Logo for you to try :)"
+	echo "DOWNLOADING: 1024px-Star_Wars_Logo.svg.png .........."
+	echo ""
+	wget https://upload.wikimedia.org/wikipedia/commons/thumb/6/6c/Star_Wars_Logo.svg/1024px-Star_Wars_Logo.svg.png&>/dev/null &
```
profezzorn, line 50: Why do you have an & at the end of this line?
It will make the wget command run in the background, so the file might not exist by the time we get to the next command.

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+echo ""
+echo "preview image:"
+echo "	ProffieOS/pov_tools/"$sourcebase"_"$HEIGHT"_preview.png"
+echo "Full color image data file:"
+echo "	ProffieOS/styles/"$sourcebase"_"$HEIGHT"_FC_POV_data.h"
+echo "Single color image data file:"
+echo "	ProffieOS/styles/"$sourcebase"_"$HEIGHT"_SC_POV_data.h"
+echo "256 color image data file:"
+echo "	ProffieOS/styles/"$sourcebase"_"$HEIGHT"_8b_POV_data.h"
+rm pnmtorle; rm pnmquantizedtorle; rm pgmtorle; rm pnmwindshieldwiper
```
profezzorn, line 217: What's the point of deleting these?

NoSloppy:
oof. ok thanks for the review. All makes perfect sense.
Most of the haphazard use of multiple methods was me looking for an answer and definitely copy/pasted some stackoverflow.com replies, tested and if it worked, let it be. I see that I should have other considerations on what some flags might cause to other files. Manipulating files in that what I consider to be a working directory, I saw no need to concern with any subdirectories existing etc... that only the files that belong in there would/should/could be in there. I was focused on the user experience, and testing from that standpoint.
Also, as it grew, I learned things as I went along but didn't retro recode existing, working things to the better way maybe.
 I'll go one by one, it's going to take a bit. Thanks again.

profezzorn:
Oh, I get cut-n-pasting solutions. I've been programming for 30+ years, and I still do that. :)
But when someone shows me a better way to do something, or offers a better understanding of how I should have done, I'm always willing to do the work to make the code better. It benefits me, it benefits the code, and ultimately, it's really not that much work. Once you've written something one, re-writing it takes much less time, because figuring out what you want is usually what takes the longest.

pov_tools/create_POV_data_files
NoSloppy: > ... when someone shows me a better way to do something, or offers a better understanding of how I should have done, I'm always willing to do the work to make the code better. It benefits me, it benefits the code, and ultimately, it's really not that much work.

I'm soaking it all in and really appreciate the professional guidance. I'll be back with fxes.

pov_tools/create_POV_data_files
NoSloppy: Just cleaning up after the process is all. I had asked about these hex versions being generated and you didn't know why, which to me means they're superfluous once the job is done.  They could just as well stay if it's a problem removing them.

pov_tools/create_POV_data_files
NoSloppy: > Why do you have an & at the end of this line? It will make the wget command run in the background, so the file might not exist by the time we get to the next command.

In the background intentional just to keep the console cleaner. The "fake downloading" line is there instead, and the pause until a key is pressed was enough time for it to download in testing. I suppose someone might have McDonalds WiFi and need more time.... I'll make it active.

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
+#!/bin/bash
+
+# create_POV_data_files
+
+cd "$(dirname "$0")"
+clear
+echo ""; echo "Welcome! Let's make some POV images!"; echo ""
+# replace whitespace in source filename with underscores
+find . -type f -name "* *" | while read file; do mv "$file" ${file// /_}; done
```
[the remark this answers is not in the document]
NoSloppy: Sloppy. Goes against my username. Fixed.

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,218 @@
...
+have_logo=0
+height_only=0
+preset=0
+# get quantity of .png files and build array to print if multiples
+files=$(ls 2>/dev/null -Ubad1 -- *.png | wc -l)
+pngs=$(find . -type f -name '*.png')
```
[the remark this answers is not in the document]
NoSloppy: Sloppy. Fixed.

pov_tools/create_POV_data_files
NoSloppy: Makes sense. Cool, thanks.

pov_tools/create_POV_data_files
NoSloppy: gone.

pov_tools/create_POV_data_files
NoSloppy: Courtesy extended.

pov_tools/create_POV_data_files
NoSloppy: Yup. Now just $pngs throughout.

pov_tools/create_POV_data_files
NoSloppy: Agreed. And no need to differentiate that its the logo image. Set it as the image and done.

pov_tools/create_POV_data_files
NoSloppy: Just setting parameters now. One make command for all at the end.

pov_tools/create_POV_data_files
NoSloppy: Still need to check both ranges though as much as I can figure... see alteration.

pov_tools/create_POV_data_files
NoSloppy: fixed/gone anyway

pov_tools/create_POV_data_files
NoSloppy: foreground

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,251 @@
...
+
+pngname="${pngs##*/}"
+pngbase="${pngname%.*}"
+echo ""
+echo "Using source image file: $pngname"
+echo ""
+
+# check for 'preview' in filename
+if [[ $pngname == *"preview"* ]]; then
+	echo "OOPS! Source file name can not contain 'preview'."
```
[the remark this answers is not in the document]
profezzorn: I get that it can't be *exactly* "preview.png", but it it could be "obiwan_preview.png".

pov_tools/create_POV_data_files
profezzorn: I think I misunderstood then.
These files are the actual programs. If you delete them they are going to need to be re-compiled each time which slows things down.

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,251 @@
...
+rename=0
+remove=0
+
+echo ""
+echo "Welcome! Let's make some POV images!"
+echo ""
+echo "Found these image files:"
+# get file names and quantity of existing .png files
+# check for whitespace in name
+for pngs in `find . -maxdepth 1 -type f -name "*.png"`; do
```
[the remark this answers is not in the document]
NoSloppy: The initial find might be multiples. Later it could become singular png, sure. 
If we're sticking with the way it has worked up until now (processing one file only).
Questions as to "why" no multiples or spaces or 'preview' contained in the name are all answered with: It's just conventions to make the automation work the way doing a manual make did. If we're changing how many files can be processed and are allowed etc.. that of course changes the course of this script.

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,251 @@
...
+	for strip in ${pngs[@]}; do echo "${strip##*/}"; done
```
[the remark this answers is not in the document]
NoSloppy: The array contains the files and the list is printed under "Found these files: "
At the beginning, there may be multiple .pngs existing.

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,251 @@
...
+	echo "Type y for yes,"
+	echo "type q to quit,"
+	read -p $'or press ENTER to start over..\n' dl
+	case $dl in
+		y) download=1;;
+		q) exit;;
+		*) exec "./create_POV_data_files";;
+	esac
+fi
+if [[ $download == 1 ]]; then
```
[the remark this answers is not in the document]
NoSloppy: 1. fixed
2. because oops/leftovers. They should just be the more friendly [[ ]]. fixed

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,251 @@
...
+fi
+
+# Check if too many files, get it down to one
+if [ "$pngcount" -ge 2 ]; then
+	echo ""
+	echo "OOPS! Too many .png image files found!"
+	echo ""
+	echo "There needs to be only ONE source .png image in the pov_tools folder."
+	echo ""
+	echo "Type r to remove ALL FILES with 'preview' in the name. "
```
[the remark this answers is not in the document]
NoSloppy: Great. Done. Note to self to add to the ReadMe "File names that contain 'preview' will  be ignored as valid source image files."

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,251 @@
...
+	echo "Files containing *preview* deleted if they existed"
+	# give user time to read message
+	read -n 1 -r -s -p $'Press any key to continue.\n'
+	exec "./create_POV_data_files"
+fi
+
+# if only one file, but has whitespace in file name, offer underscores
+if [[ $wscount != 0 ]]; then
+	echo ""
+	echo "OOPS! Spaces found in .png image file name. That's not ideal. "
```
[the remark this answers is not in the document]
NoSloppy: Could this stay? Aside from being more work to accommodate escaped characters etc... I'd like to encourage the world to not do it in the first place! 
Seriously though, if it's not an issue, I'd prefer this remains.

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,251 @@
...
+	for strip in ${pngs[@]}; do echo "${strip##*/}"; done
+	pngcount=$((pngcount+1))
+	if [[ $pngs = *" "* ]]; then wscount=$((wscount+1)); fi
+done
+echo ""
+echo ".png files found: " $pngcount
+echo ".png files with spaces in file name: " $wscount
+
+# If NO images in the folder, offer logo
+if [ $pngcount -lt 1 ]; then
```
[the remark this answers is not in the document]
NoSloppy: As I said above, this script was just to automate the process a bit for what was already in place using cmd line and making manually.
If we want to change it and process multiples, that seems like a reasonable idea to me. Let me fix all this other stuff so it's right, then if that's what you think we should do, I could work that out.

pov_tools/create_POV_data_files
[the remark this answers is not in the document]
NoSloppy: ok....well we could make this all MUCH simpler and just process ALL .png files it finds in the folder.
Then the instructions are just "Any .png files in the 'pov_tools' are going to get processed to the parameters set below"

I'm game if that's what you want to do!

NoSloppy:
I've been pulling my hair out for hours over this, and maybe it turns out 'make' won't take spaces no matter how much quoting?

NoSloppy:
ok. simplified everything, worked around spaces. How does it look now?

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,251 @@
...
+	echo "Type q to quit."
+	read -p $'Press ENTER to start over.\n' fix
+	case $fix in
+		r) remove=1;;
+		q) exit;;
+		*) exec "./create_POV_data_files";;
+	esac
+fi
+if [[ $remove == 1 ]]; then
+	rm *preview*.png;
```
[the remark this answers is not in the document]
NoSloppy: Irrelevant now. Files containing 'preview' not read nor deleted.

pov_tools/create_POV_data_files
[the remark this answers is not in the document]
NoSloppy: No need to keep this now. Worked around 'make' not taking them anyway. Resulting files are named properly, user's files are left as they are.

pov_tools/create_POV_data_files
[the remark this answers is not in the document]
NoSloppy: Converting all the files now.

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,201 @@
...
+	wget https://upload.wikimedia.org/wikipedia/commons/thumb/6/6c/Star_Wars_Logo.svg/1024px-Star_Wars_Logo.svg.png
+	pngs=$"1024px-Star_Wars_Logo.svg.png"
+fi
+
+valid_option=0
+while [[ $valid_option != 1 ]]; do
+	echo "Please select the settings you want to create the POV image(s) with:"
+	echo "	height - number of pixels in your blade (one strip emitter to tip)"
+	echo "	length - blade length in the measurement unit of choice"
+	echo "	offset - the distance (in same unit as L) from the bottom of the blade"
```
kaQnub, line 50: "(in same unit as L)" should be: "(in same unit as length)". This is probably still from when there was L for length in the comments)

pov_tools/create_POV_data_files
```diff
@@ -0,0 +1,201 @@
...
+	echo "q - to quit"
+	read option
+	case $option in
+		"") echo "! Some option is needed. Please try again"; echo "";;
+		1) units="inches" height=144 length=39 offset=18 preset=1 valid_option=1;;
+		2) units="centimeters" height=97 length=63 offset=50 preset=1 valid_option=1;;
+		3) units="centimeters" height=72 length=50 offset=15 preset=1 valid_option=1;;
+		4) units="centimeters" height=72 length=50 offset=39 preset=1 valid_option=1;;
+		5) height_only=1 valid_option=1;;
+		6) valid_option=1;;
```
profezzorn, line 68: "custom options" (6) doesn't seem to do anything, how is that supposed to work?

kaQnub: With option 6 the whole package of questions is getting asked:
- how many pixels in blade
- preferred unit
- length of blade (in selected unit) (there is a default suggestion based on pixels and selected unit)
- offset (in selected unit) (there is a default option based on selected unit that matches the 18" from windshieldwiper default)
- -> make

kaQnub: the valid_option=1 helps to break the while-do loop

profezzorn: I get that option 6 exits the menu, but how does it make it "custom" ?

NoSloppy: Custom options are then input by the questionnaire that follows..

pov_tools/create_POV_data_files
NoSloppy: fixed

PR #398, opened 2022-01-24
Prevent SPARK_MS = 0 to match WAVE_MS

fdarosa2663:


transitions/wave.h
```diff
@@ -73,6 +73,7 @@ class TrSparkX : public TransitionBaseX<SPARK_MS> {
 public:
 
   void run(BladeBase *blade) {
+    static_assert(!is_same_type<SPARK_MS, Int<0>>::value);
```
profezzorn, line 76: There is no division by zero to worry about in this code, so this is not needed.

PR #401, opened 2022-01-27
Option to limit Low Batt notification to only play once

NoSloppy:
Tried to do this in prop file and not edit a system file, but doesn't seem easily done, if even possible.

### DIFF
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -737,13 +737,19 @@ public:
   }
 
   void SB_LowBatt() {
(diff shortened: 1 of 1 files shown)


profezzorn:
Isn't it just a matter of overriding CheckLowBattery() ?

NoSloppy:
But this lets the saber still turn off if low.

profezzorn:
You don't have to override CheckLowBattery() with one that does nothing though.

profezzorn:
Still think this can be done by overriding CheckLowBattery().

PR #402, opened 2022-02-02
add StylePOV<> ContinuousPOV<> usage. 

NoSloppy:


styles/pov.h
```diff
@@ -4,15 +4,34 @@
...
 // the result without a slow exposure camera. It's possible that
 // dotstar APA102 pixels would be fast enough to do the effect
 // full justice, but I haven't tried that.
 
+// Usage: &style_pov
+// or: StylePtr<StylePOV<>>()
+// or: StylePtr<ContinuousPOV<>>()
+// return value: suitable for preset array
+// StylePOV<> draws one instance of the image in the top portion of the swing "window".
+//ContinuousPOV<> draws the image multiple times continuously through 360 degrees of the swing.
```
profezzorn, line 19: space after //

NoSloppy: Done.

styles/pov.h
```diff
@@ -4,15 +4,34 @@
...
-// return value: suitable for preset array
 
 // This style draws images in the air if you swing the saber
 // carefully back and forth. Unfortunately you can't really see
 // the result without a slow exposure camera. It's possible that
 // dotstar APA102 pixels would be fast enough to do the effect
 // full justice, but I haven't tried that.
 
+// Usage: &style_pov
+// or: StylePtr<StylePOV<>>()
```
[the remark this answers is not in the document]
NoSloppy: Which are we talking about? Something else to go in the style template?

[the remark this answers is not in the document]
profezzorn: StylePOV takes four (optional) arguments:
template<int MIN_DEGREES=-45, int MAX_DEGREES=45, bool REPEAT=false, bool MIRROR=true>

ContinousPOV takes 1:
template<int DEGREES=90>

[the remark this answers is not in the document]
profezzorn: GRR, github ate everything between less-than and greater-than :(
See line 102 and 130 to see the arguments.

styles/pov.h
```diff
@@ -4,14 +4,41 @@
...
+full justice, but I haven't tried that.
+
+Usage: &style_pov
+or: StylePtr<StylePOV<>>()
+or: StylePtr<ContinuousPOV<>>()
+return value: suitable for preset array
+
+StylePOV<> draws one instance of the image in the top portion of the swing "window".
+StylePOV takes four (optional) arguments:
+template<int MIN_DEGREES=-45, int MAX_DEGREES=45, bool REPEAT=false, bool MIRROR=true>
```
profezzorn, line 21: This is a weird way to write it.
I would write it something like: (hopefully github doesn't eat the less-than and greater-then)

or: StylePOV<MIN_DEGREES, MAX_DEGREES, REPEAT, MIRROR>>()

MIN_DEGREES: integer (default -45)
MAX_DEGREES: integer (default 45)
REPEAT: bool (default false)
MIRROR: bool (default true)
return value: COLOR

Since MIN_DEGREES, MAX_DEGREES, REPEAT and MIRROR all have defaults, you can write it without them, like: StylePOV<>

NoSloppy: updated

styles/pov.h
```diff
@@ -4,14 +4,48 @@
...
-// full justice, but I haven't tried that.
+/* 
+This style draws images in the air if you swing the saber
+carefully back and forth. Unfortunately you can't really see
+the result without a slow exposure camera. It's possible that
+dotstar APA102 pixels would be fast enough to do the effect
+full justice, but I haven't tried that.
+
+Usage: &style_pov
+or: StylePtr<StylePOV<>>()
```
profezzorn, line 15: Remove StylePtr

styles/pov.h
```diff
@@ -4,14 +4,48 @@
...
+or: StylePtr<StylePOV<MIN_DEGREES, MAX_DEGREES, REPEAT, MIRROR>>()
+MIN_DEGREES: integer (default -45)
+MAX_DEGREES: integer (default 45)
+REPEAT: bool (default false)
+MIRROR: bool (default true)
+return value: COLOR
+Since MIN_DEGREES, MAX_DEGREES, REPEAT and MIRROR all have defaults, those arguments can be omitted.
+StylePOV<> draws one instance of the image in the top portion of the swing "window".
```
profezzorn, line 23: We also need to tell people what the arguments actually do.

styles/pov.h
```diff
@@ -4,14 +4,48 @@
...
-// dotstar APA102 pixels would be fast enough to do the effect
-// full justice, but I haven't tried that.
+/* 
+This style draws images in the air if you swing the saber
+carefully back and forth. Unfortunately you can't really see
+the result without a slow exposure camera. It's possible that
+dotstar APA102 pixels would be fast enough to do the effect
+full justice, but I haven't tried that.
+
+Usage: &style_pov
```
profezzorn, line 14: I think it might be better to move this one below and say something like:

In older files, you may see &style_pov used, which is essentially the same as StylePtr<StylePOV<>>()

styles/pov.h
```diff
@@ -4,14 +4,48 @@
...
+or: StylePtr<StylePOV<MIN_DEGREES, MAX_DEGREES, REPEAT, MIRROR>>()
+MIN_DEGREES: integer (default -45)
+MAX_DEGREES: integer (default 45)
+REPEAT: bool (default false)
+MIRROR: bool (default true)
+return value: COLOR
+Since MIN_DEGREES, MAX_DEGREES, REPEAT and MIRROR all have defaults, those arguments can be omitted.
+StylePOV<> draws one instance of the image in the top portion of the swing "window".
+
+Usage:StylePtr<ContinuousPOV<>>()
```
profezzorn, line 25: remove StylePtr

styles/pov.h
```diff
@@ -4,14 +4,48 @@
...
+or: StylePtr<ContinuousPOV<DEGREES>>()
+DEGREES: integer (default 90)
+return value: COLOR
+Omitting the DEGREES argument uses the default value of 90.
```
profezzorn, line 29: This line seems redundant.

styles/pov.h
```diff
@@ -4,14 +4,48 @@
...
+ContinuousPOV<> draws the image multiple times continuously through 360 degrees of the swing.
```
profezzorn, line 30: People might be confused about the difference between ContinuousPOV and StylePOV<> with REPEAT=true, so we should probably explain the difference, something like:

StylePOV tries to always display the same line when the saber is at a particular angle.
ContinuousPOV simply moves from line to line, and the speed is controlled by the swing speed.
If you move the saber left and then right, StylePOV will go *back*, ContinuousPOV will continue forward.

styles/pov.h
```diff
@@ -4,14 +4,53 @@
...
+or: StylePOV<MIN_DEGREES, MAX_DEGREES, REPEAT, MIRROR>
+MIN_DEGREES: integer (default -45)
+MAX_DEGREES: integer (default 45)
+REPEAT: bool (default false)
+MIRROR: bool (default true)
+return value: COLOR
+Since MIN_DEGREES, MAX_DEGREES, REPEAT and MIRROR all have defaults, those arguments can be omitted.
+StylePOV<> draws one instance of the image in the top portion of the swing "window".
+MIN and MAX degrees set the start and end points for the edges of the image.
+REPEAT draws multple instances of the image through the 360 arc, butting each next image adjacent to the previous.
```
profezzorn, line 24: Some of these lines are too long.
Please stick with something reasonable, like maybe 100 characters.

styles/pov.h
NoSloppy: I think this is done?

PR #403, opened 2022-02-06
Kill off oldest effect if run out of wavplayers.

NoSloppy:


sound/hybrid_font.h
```diff
@@ -234,11 +234,38 @@ class HybridFont : public SaberBase, public Looper {
     EnableAmplifier();
     if (!f->files_found()) return RefPtr<BufferedWavPlayer>(nullptr);
     RefPtr<BufferedWavPlayer> player = GetFreeWavPlayer();
-    if (player) {
-      player->set_volume_now(font_config.volEff / 16.0f);
-      player->PlayOnce(f);
-      current_effect_length_ = player->length();
+    if (!player) {
+      STDOUT.println("Out of WAV players! Getting more...");
+      float oldest_ = 0;
```
profezzorn, line 239: kill_ and oldest_ are local variables, they should not end with _.
They should also have better names, like maybe "highest_progress" and "player_to_restart".

sound/hybrid_font.h
```diff
@@ -234,11 +234,38 @@ class HybridFont : public SaberBase, public Looper {
...
+      // check wavplayers, rule out hum and smoothswings then kill oldest one 
+      for (size_t i = 0; i < NELEM(wav_players); i++) {
+        if (wav_players[i].isPlaying() && wav_players[i].current_file_id().GetEffect() == f) {
+          float pos = wav_players[i].pos();
+          if (pos > oldest_) {
+            oldest_ = pos;
+            kill_ = i;
+          }    
+        } else {
+          // unsuccessful acquisition. Let's try again.
```
profezzorn, line 250: This comment doesn't make any sense.
I don't see any "trying again" here.

sound/hybrid_font.h
```diff
@@ -234,11 +234,38 @@ class HybridFont : public SaberBase, public Looper {
...
+          if (i == 6) return player;
```
profezzorn, line 251: this is the wrong end condition
also, we should never hardcode a number like this, as it would stop working if we ever change the number of wav players.

sound/hybrid_font.h
```diff
@@ -234,11 +234,38 @@ class HybridFont : public SaberBase, public Looper {
...
+        }
+      }
+      if (wav_players[kill_].pos() > .002) {
```
profezzorn, line 254: What is this if statement for?

sound/hybrid_font.h
```diff
@@ -234,11 +234,38 @@ class HybridFont : public SaberBase, public Looper {
...
+        wav_players[kill_].set_fade_time(0.001);
+        wav_players[kill_].FadeAndStop();
+        STDOUT << "Killing off " << wav_players[kill_].filename() << "\n";
+        while(wav_players[kill_].isPlaying()) {
```
profezzorn, line 258: need a space after "while".

sound/hybrid_font.h
```diff
@@ -234,11 +234,38 @@ class HybridFont : public SaberBase, public Looper {
...
+#ifdef VERSION_MAJOR >= 4
+          armv7m_core_yield();
+#endif
+        }
+      }
+      player = GetFreeWavPlayer();
```
profezzorn, line 264: We don't need to call GetFreeWavPlayer() since we already know which wav player we killed.
player = RefPtr<BufferedWavPlayer>(wav_players + kill_);

sound/hybrid_font.h
```diff
@@ -234,11 +234,38 @@ class HybridFont : public SaberBase, public Looper {
...
     if (!f->files_found()) return RefPtr<BufferedWavPlayer>(nullptr);
     RefPtr<BufferedWavPlayer> player = GetFreeWavPlayer();
-    if (player) {
-      player->set_volume_now(font_config.volEff / 16.0f);
-      player->PlayOnce(f);
-      current_effect_length_ = player->length();
+    if (!player) {
+      STDOUT.println("Out of WAV players! Getting more...");
+      float oldest_ = 0;
+      uint32_t kill_ = 0;
```
profezzorn, line 240: make this an "int", and then change 0 to -1 so that we can recognize when nothing was found below. I also think that we should also count how many players are playing  the same sound and only do the killing if the number of players is > 1, otherwise it's going to be too obvious.

sound/hybrid_font.h
```diff
@@ -234,11 +234,38 @@ class HybridFont : public SaberBase, public Looper {
...
+      // check wavplayers, rule out hum and smoothswings then kill oldest one 
+      for (size_t i = 0; i < NELEM(wav_players); i++) {
+        if (wav_players[i].isPlaying() && wav_players[i].current_file_id().GetEffect() == f) {
```
profezzorn, line 243: We also need to make sure that the wav player doesn't have any references. Players with references are supposed to be left alone.

sound/hybrid_font.h
```diff
@@ -234,11 +234,38 @@ class HybridFont : public SaberBase, public Looper {
     EnableAmplifier();
     if (!f->files_found()) return RefPtr<BufferedWavPlayer>(nullptr);
     RefPtr<BufferedWavPlayer> player = GetFreeWavPlayer();
```
profezzorn, line 236: We should probably break out the stuff below into a new function, something like:

RefPtr<BufferedWavPlayer> GetOrFreeWavPlayer(Effect* e);

Then we can replace this call with a call to the new function.

sound/hybrid_font.h
NoSloppy: Well this was the fix for freezing and squee of death.
For some reason, it just doesn't get it sometimes and we get to here, so "try again" meaning give up this time and go back to listening for PlayPolyphonic().
Which, in testing, is an  imperceivable skip.

sound/hybrid_font.h
NoSloppy: Good point on the future.
If not returning, what's more appropriate? It did not like break;

sound/hybrid_font.h
NoSloppy: I was thinking a guard.
Would it be fine if we had less wav left than the fade time that follows?

sound/hybrid_font.h
NoSloppy: sure.

sound/hybrid_font.h
NoSloppy: Ah. ok.

sound/hybrid_font.h
profezzorn: this code just needs to go away
after the loop we should check if kill_ (or whatever rename that variable to) is -1.

sound/hybrid_font.h
profezzorn: a guard against what?
This if statements seems to check that we have played at last 2 milliseconds of audio on this wav file...

sound/hybrid_font.h
NoSloppy: > make this an "int", and then change 0 to -1 so that we can recognize when nothing was found below.

I was thinking with the un-hardcoding of the i == 6 above, to just do ` if (i == NELEM(wav_players))` 
In testing with printouts, I saw the for loop cycle 0-6 over and over when not getting a sound, therefore never updating kill (now player_to_restart). So checking if we've changed that int wouldn't work. Well, not the way it's written currently anyway.

> I also think that we should also count how many players are playing the same sound and only do the killing if the number of players is > 1, otherwise it's going to be too obvious.

It doesn't seem possible we'd ever kill off a single instance of a sound though, no?
With 7 wavplayers,
a max of 4 tied up with non-effects 
we're spamming new effect events, using the last 3 or 4 players,
the oldest player_to_restart should always be 3 or 4 players older than our current sound.

sound/hybrid_font.h
NoSloppy: ok.

sound/hybrid_font.h
NoSloppy: Attempt made, although I'm not 100% that the filtering out only the wavplayers playing the effect is foolproof. 
I did have a 1 in 1000 occasion where humm was stopped to free up a player, and that's unacceptable of course.

sound/hybrid_font.h
NoSloppy: fixed.

sound/hybrid_font.h
NoSloppy: oh yeah oops. Adjusted to be if length - current position < .002, then bail.

sound/hybrid_font.h
NoSloppy: changed.

NoSloppy: Ok I've found the fail point in this for-loop. I think we just need to check if the sound  is *an* effect, instead of if it is the *currect* effect.  
`wav_players[i].current_file_id().GetEffect() == e` will fail if we're loaded up with blst wavs, then a clsh comes through.
This is where it freezes because this scenario is not covered.  
How could we broaden the filter so any file NOT hum smoothswing or track is ok?

NoSloppy: I got it! We check with GetEffect() but don't compare to *current* effect.
Working well.
Removed time left for fade out check, unless that's needed. I think it goes away.
Removed fail case for the for-loop as it will always get a player playing an effect now.
I still don't get what we would need to return in order to fail gracefully,  ie: not get a player to free up, but still satisfy the function somehow. But not failing seems like a good idea.

sound/hybrid_font.h
NoSloppy: Now removed.

sound/hybrid_font.h
NoSloppy: > 

What would we do if it were still -1 because for SOME reason we got no players to kill?
Looking for what we would end up returning to the function.

sound/hybrid_font.h
NoSloppy: removed.

sound/hybrid_font.h
profezzorn: If we don't check that it's the same effect, then things gets complicated, because we should never kill the hum, humm, in, out, track, swingl, swingh, and probably a lot of other things. Basically it gets hard to know what we can/should kill.

Also, the assertion that we should never see a case where we can't kill an old wav player is just not correct.
However, this particular if statement should still go, it was never working the way you meant to anyways.

sound/hybrid_font.h
```diff
@@ -230,16 +231,45 @@ class HybridFont : public SaberBase, public Looper {
...
+  RefPtr<BufferedWavPlayer> GetOrFreeWavPlayer(Effect* e) {
+    free_player_ = GetFreeWavPlayer();
+    if (!free_player_) {
+      STDOUT.println("Out of WAV players! Getting more...");
+      float highest_progress = 0;
+      int player_to_restart = -1;
+      // check wavplayers, rule out hum and smoothswings then kill oldest one 
+      for (size_t i = 0; i < NELEM(wav_players); i++) {
+        if (wav_players[i].isPlaying() && 
+          wav_players[i].current_file_id().GetEffect() &&
```
profezzorn, line 243: It's not safe to just kill *any* player, since we don't know what they are used for in this code.
Basically, if you're going to consider all sounds killable, you also need to review all code that plays sound and make sure that any sound that we don't want to to be killed has a reference so that it doesn't get killed.

Also, if we end up in a situation where all the sounds that are playing are using different effects, then I don't think we should kill any of them. Only effects that play at least two sounds should be considered killable IMHO.

sound/hybrid_font.h
```diff
@@ -203,6 +203,7 @@ class HybridFont : public SaberBase, public Looper {
   RefPtr<BufferedWavPlayer> next_hum_player_;
   RefPtr<BufferedWavPlayer> swing_player_;
   RefPtr<BufferedWavPlayer> lock_player_;
+  RefPtr<BufferedWavPlayer> free_player_;
```
profezzorn, line 206: I think this should be a local variable, not a class level variable.

sound/hybrid_font.h
NoSloppy: Yes, this particular if statement was wrong.
Currently though, it's working. I tested it hard, with a printout for every step and it was all happy throughout, working as expected.
I did only test spamming blaster blocks mixed with clashes. Both were recognized as Effects, while hum smoothswings and track playing were ignored.
Not sure how out.wav or in.wav would get in the way. Maybe though.
I've hit a wall on what I can discover throughout the OS as far as what else to use to filter results to be NOT be a sound on the bad list you wrote above.

sound/hybrid_font.h
NoSloppy: So here's example output from a test. It prints out "file not referenced" only if it satisfies the if statement checking for that. 
hum, smoothswings and track are referenced and therefore not subject to being killed.
```
21:50:02.717 -> Out of WAV players! Getting more...
21:50:02.717 -> *************** Loop i = 0 wav_player has aa_Juansith/Red_Five/humm/humm02.wav
21:50:02.717 -> *************** Loop i = 1 wav_player has aa_Juansith/Red_Five/tracks/Xwing.wav
21:50:02.717 -> *************** Loop i = 2 wav_player has aa_Juansith/Red_Five/swingh/swingh01.wav
21:50:02.717 -> *************** Loop i = 3 wav_player has aa_Juansith/Red_Five/swingl/swingl01.wav
21:50:02.717 -> *************** Loop i = 4 wav_player has aa_Juansith/Red_Five/blst/blst08.wav
21:50:02.717 -> *************** file is not referenced
21:50:02.717 -> *************** oldest so far, setting it as the player_to_restart =4
21:50:02.717 -> *************** Loop i = 5 wav_player has aa_Juansith/Red_Five/blst/blst01.wav
21:50:02.717 -> *************** file is not referenced
21:50:02.717 -> *************** oldest so far, setting it as the player_to_restart =5
21:50:02.717 -> *************** Loop i = 6 wav_player has aa_Juansith/Red_Five/blst/blst01.wav
21:50:02.717 -> *************** file is not referenced
21:50:02.717 -> Stopping wav_player playing aa_Juansith/Red_Five/blst/blst01.wav
```
This is working, no?

I could really use some understanding of a negative result and how to handle it.
Like what do we do if the checks fail. How do we handle reaching the end of this process without an answer?
What would we return?
Like if we're going to NOT kill the player because there's only 1 instance of that kind of sound, what's the exit strategy then since we need to go back to PlayPolyphonic and play something but we're empty handed and don't have a free player?
Lastly, would using `std::string` be a way to get which type of sound it is(clsh, blst, etc..) so as to build a check for multiple of the same kind?

sound/hybrid_font.h
profezzorn: GetOrFreeWavPlayer needs to return it.
So it needs to be a local variable in multiple functions.

sound/hybrid_font.h
profezzorn: > This is working, no?

Perhaps, but it's not enough for it to work, it must also make sense.
This means that you have to be able to make a reasonable argument that explains why it works, not just try it.
The way the code is written now, this means every other piece of code that plays sounds in ProffieOS and make sure that it isn't impacted by what you're doing. Earlier, when you were only preempting players that were playing the *same* effect, that was much much much easier. Of course, that doesn't always work, but it's *easier*.

> Like what do we do if the checks fail. How do we handle reaching the end of this process without an answer?
> What would we return?

Just like GetFreeWavPlayer, we would return a null refptr if we fail to find an available player.

> Like if we're going to NOT kill the player because there's only 1 instance of that kind of sound, what's the exit strategy > then since we need to go back to PlayPolyphonic and play something but we're empty handed and don't have a free >player?

In this case PlayPolyphonic() will not play anything, just like it does today if GetFreeWavPlayer fails to return a player.

> Lastly, would using std::string be a way to get which type of sound it is(clsh, blst, etc..) so as to build a check for multiple of the same kind?
.
Most of the time, you would count multiples by using std::map<>, iterating over each player and then doing something like map[player[i].getEffect()]++. Unfortunately, std::map allocates memory, and it's not a good idea to use it in proffieos.
Instead we probably have to use a double loop, something like:

   for ( i  = 0 .. players ) {
       Effect e = player[i].geteffect();
       count = 0;
       for ( j = 0 .. players ) {
          if (player[j].geteffect() == e) count++;
       }
       if (count > 1) {
          // do stuff
       }
   }

This takes a little more cpu, but since it won't run that often it's not that big of a deal.

NoSloppy: > Effect e = player[i].geteffect(); 

I think I interpreted your shorthand for the most part, but this one's giving me
`error: 'class BufferedWavPlayer' has no member named 'GetEffect'`
I tried adding `current_file_id` like elsewhere, I also tried asterisk `Effect*`, `const Effect*`
to no avail.

NoSloppy: I post too quickly and make it hard to follow. apologies.

It seems we don't need the loop in the loop? We are already checking if the sound is the effect we want in the if statement of the first for-loop. We can just increment count there once that check passes, no. 
The problem is if there's a fail, it freezes with `NULL POINTER!`
Tried:
return RefPtr<BufferedWavPlayer>(nullptr);
return RefPtr<BufferedWavPlayer>();
return RefPtr<BufferedWavPlayer>;

sound/hybrid_font.h
NoSloppy: ok

sound/hybrid_font.h
NoSloppy: ok figured it out

sound/hybrid_font.h
NoSloppy: SO I went another step and added saving the 2nd oldest playing sound as well. The idea is if a new effect comes in, and the loop doesn't get a player to restart, then we'd use the backup.
However, this has reached a point of diminishing returns.
All this checking and processing causes a lag when it runs no different than just waiting like the original code did.
Unless you can optimize some things to have this run quicker, I'm not sure it's worth it.

It used to be snappy a few iterations ago, before checking quantity playing. Worked well actually. 
I don't know if you've tried it in real life, but I'd say there's just no way you'd hear that ANY are being killed, much less which one. It's a frantic cacophony of effect sounds. The goal was no dead spots when spamming the button or clashing like crazy, which it satisfied.
If the concern of hearing a sound get cut off could be reconsidered, this has a nice result. I could revert back to what works safely, makes sense, and maybe be done?

profezzorn: Running all of this code takes a couple of microseconds, so I don't think that's the problem.

profezzorn:
There may be a better/easier way to do all of this.
We could add a variable in the Effect class that keeps track of whether an effect is killable or not. This variable could be overriden in the font config file, but the default would be that we would make effects like clash, blast, force etc. killable, while hum, in, out would not be.
By doing this we won't have to worry too much about killing the wrong thing, so then we just need to make the code sound good.

NoSloppy:
> 

This is where I'm confused.
In testing, the loop is already consistently ignoring the ones that should not be killed (which seems to be the check if it's referenced)

sound/hybrid_font.h
```diff
@@ -230,18 +230,72 @@ class HybridFont : public SaberBase, public Looper {
...
+          STDOUT << "Stopping wav_player " << saved_player_to_restart_ << " playing " << wav_players[saved_player_to_restart_].filename() << "\n";
+          while (wav_players[saved_player_to_restart_].isPlaying()) {
+#ifdef VERSION_MAJOR >= 4
+            armv7m_core_yield();
+#endif
+          }
+          free_player = RefPtr<BufferedWavPlayer>(wav_players + saved_player_to_restart_);
+        }
+      }
+      return RefPtr<BufferedWavPlayer>(); // failure to fulfill is same lag as just waiting like we did before all this. Not great.
```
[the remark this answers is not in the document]
NoSloppy: The comment will go away. Just a personal note really. "same lag as we had before", meaning a dead button while we wait for a free player, which is what all of this was supposed to prevent.

sound/hybrid_font.h
```diff
@@ -230,18 +230,72 @@ class HybridFont : public SaberBase, public Looper {
     if (loop) hum_player_->PlayLoop(loop);
   }
 
+  int saved_player_to_restart_ = -1 ;
+  RefPtr<BufferedWavPlayer> GetOrFreeWavPlayer(Effect* e) {
```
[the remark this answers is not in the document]
NoSloppy: Isn't that the current effect trying to be played?
We use that to check if wav_players[i].current_file_id().GetEffect() == e 
I may very well not understand something correctly.

profezzorn:
That's true for every use case you tested, but is it true for every sound played in proffieos?

sound/hybrid_font.h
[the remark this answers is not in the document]
profezzorn: Wait, I think I might be confused.
I thought you had stopped doing that and would consider sounds from other effects for killing as well?

sound/hybrid_font.h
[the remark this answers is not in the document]
profezzorn: There is no way to prevent it in every single case, but the problem here is that the comments makes no sense.
Lag means that a sound has been delayed somehow, not missed, which is what happens in this case.

sound/hybrid_font.h
```diff
@@ -230,18 +230,72 @@ class HybridFont : public SaberBase, public Looper {
...
     EnableAmplifier();
     if (!f->files_found()) return RefPtr<BufferedWavPlayer>(nullptr);
-    RefPtr<BufferedWavPlayer> player = GetFreeWavPlayer();
-    if (player) {
-      player->set_volume_now(font_config.volEff / 16.0f);
-      player->PlayOnce(f);
-      current_effect_length_ = player->length();
-    }
-    return player;
+    RefPtr<BufferedWavPlayer> effect_player = GetOrFreeWavPlayer(f);
```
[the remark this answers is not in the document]
NoSloppy: Ok, I don't see how that works then.
When it's just `GetOrFreeWavPlayer();` without the (f), there's "no matching function for call".
As just `GetOrFreeWavPlayer(f);`, having changed the RefPtr to be local, "effect_player hasn't been declared in this scope".
If free_player_ goes back to being class level, and effect_player is changed back to free_player_, 
then `GetOrFreeWavPlayer(f);` by itself works.
That's what I know.

sound/hybrid_font.h
```diff
@@ -230,18 +230,72 @@ class HybridFont : public SaberBase, public Looper {
...
+        wav_players[player_to_restart].FadeAndStop();
+        STDOUT << "Stopping wav_player " << player_to_restart << " playing " << wav_players[player_to_restart].filename() << "\n";
+        while (wav_players[player_to_restart].isPlaying()) {
+#ifdef VERSION_MAJOR >= 4
+          armv7m_core_yield();
+#endif
+        }
+        free_player = RefPtr<BufferedWavPlayer>(wav_players + player_to_restart);
+      } else {
+        if (saved_player_to_restart_ > -1) {
```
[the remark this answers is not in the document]
NoSloppy: Not identical. One instance when using player_to_restart, the other when using saved version saved_player_to_restart.

sound/hybrid_font.h
```diff
@@ -230,18 +230,72 @@ class HybridFont : public SaberBase, public Looper {
...
+            highest_progress = pos;
+            saved_player_to_restart_ = player_to_restart;
+            player_to_restart = i;
+          } else if (pos > saved_progress) {
+            saved_progress = pos;
+            saved_player_to_restart_ = i;
+          }
+        }
+      }
+      if (count > 1) {
```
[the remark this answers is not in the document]
NoSloppy: I had that actually, but deduced there can't be case that we've incremented count and not also incremented player_to_restart.

sound/hybrid_font.h
```diff
@@ -230,18 +230,72 @@ class HybridFont : public SaberBase, public Looper {
...
+      float saved_progress = 0.00;
+      float highest_progress = 0.00;
+      int player_to_restart = -1;
+      int count = 0;
+      // check wavplayers, rule out hum and smoothswings then kill oldest one 
+      for (size_t i = 0; i < NELEM(wav_players); i++) {
+        if (wav_players[i].isPlaying() &&
+          wav_players[i].current_file_id().GetEffect() == e &&
+          wav_players[i].refs() == 0) {
+          count++;
```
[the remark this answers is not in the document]
NoSloppy: I see. Isn't the == e is already ensuring we're only counting one kind of sound, the current requested effect, not all effects?

NoSloppy:
> 

Ok , yes, this has just been tested with sounds I think are likely to cause a lack of wav_players: blst and clsh. (even did quote). I can't think of any other sounds that this would be an issue though. Sound Library sounds queue up and don't overlap, and I don't see too many battery level wavs being a thing.
Thinking the opposite way, what sounds do we want to NOT kill other than hum, smoothswings and track, all which currently seem to be "safe" from being killed?

sound/hybrid_font.h
```diff
@@ -230,18 +230,72 @@ class HybridFont : public SaberBase, public Looper {
     if (loop) hum_player_->PlayLoop(loop);
   }
 
+  int saved_player_to_restart_ = -1 ;
+  RefPtr<BufferedWavPlayer> GetOrFreeWavPlayer(Effect* e) {
+    RefPtr<BufferedWavPlayer> free_player = GetFreeWavPlayer();
+    if (!free_player) {
+      STDOUT.println("Out of WAV players! Getting more...");
+      float saved_progress = 0.00;
```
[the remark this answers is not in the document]
NoSloppy: See my comment above about " added saving the 2nd oldest playing sound as well. "

sound/hybrid_font.h
NoSloppy: done

sound/hybrid_font.h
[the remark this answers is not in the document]
NoSloppy: Yes, == e was removed, but then your with your suggestion that it's much much much easier to only check the *same* effect, I reverted.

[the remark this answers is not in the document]
profezzorn: Ah, well, if you put that back, then things gets a lot simpler.

sound/hybrid_font.h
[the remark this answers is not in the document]
profezzorn: ok, but why is here at all?

sound/hybrid_font.h
NoSloppy: I think this is what I was thinking.
So wav_players are all full. We come through with an effect that's new.
Because we fail the if statement for if == e, we do not choose a player to kill.
This is where a saved_player_to_restart, which we acquired the last time around as the 2nd oldest sound, can be killed and make room for our new current effect to play.

sound/hybrid_font.h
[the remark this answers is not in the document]
NoSloppy: explained above https://github.com/profezzorn/ProffieOS/pull/403#discussion_r804361882

sound/hybrid_font.h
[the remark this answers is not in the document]
NoSloppy: removed

sound/hybrid_font.h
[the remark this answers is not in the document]
NoSloppy: Simple would be good. I feel this is not as simple as it could be. It's just what I can muster up :)

sound/hybrid_font.h
```diff
@@ -235,29 +235,37 @@ class HybridFont : public SaberBase, public Looper {
...
       int player_to_restart = -1;
-      // check wavplayers, rule out hum and smoothswings then kill oldest one 
+      int count = 0;
+      // check wavplayers, rule out non-effects, then kill oldest one
+      // IF it's not the only instance playing. 
       for (size_t i = 0; i < NELEM(wav_players); i++) {
-        if (wav_players[i].isPlaying() && 
-          wav_players[i].current_file_id().GetEffect() &&
+        if (wav_players[i].isPlaying() &&
+          wav_players[i].current_file_id().GetEffect() == e &&
```
NoSloppy, line 245: restored specific effect here.

sound/hybrid_font.h
profezzorn: Ok, now I understand why it's here.
But it doesn't really work, because the "saved" variable is local and will not survive until the next call.
How about this, we add a bit to the Effect class, but instead of adding an ini file setting for it, we just set it from this function.

Basically, at the beginning of this function we would do: e->set_killable(true);
Then when we iterate over the players, we check if it's killable:

    Effect* effect = wav_players[i].GetEffect();
    if (!effect || !effect->is_killable()) continue;

NoSloppy: > Ok, now I understand why it's here. But it doesn't really work, because the "saved" variable is local and will not survive until the next call. 

int saved_player_to_restart_ = -1 ; is initialized outside the function and therefore not local, and should, and does work that way in tests. 


How about this, we add a bit to the Effect class, but instead of adding an ini file setting for it, we just set it from this function.
> 
> Basically, at the beginning of this function we would do: e->set_killable(true); Then when we iterate over the players, we check if it's killable:
> 
> ```
> Effect* effect = wav_players[i].GetEffect();
> if (!effect || !effect->is_killable()) continue;
> ```

Ok, however I think there's 2 different objectives at play here. The first being NOT killing an incorrect file, as you've addressed right here. I'm still unsure why more needs to be done here, as referenced sounds are indeed being ignored as we iterate through the players. Perhaps referenced is not enough "proof" it's not killable and I don't understand, and/or maybe I need to try to test with in.wavs?

But the second (and my primary) objective is trying to prevent a missed effect, a dead button... a dud.
That's what I was working on above. Killing the next viable player based on previously recorded oldest duration when a new effect is current (and therefore not triggering a kill based on itself). That code is also working as-is in tests according to a slew of printouts. 

Last I tested with the printouts though, there was new, significant delay everytime we hit "Out of WAV Players!" and GetOrFreeWavPlayer() ran. As you pointed out, we're talking milliseconds to run that bit, so something else must have crept up. Won't have hands-on test time for about a week though as of today.

sound/hybrid_font.h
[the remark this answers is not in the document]
profezzorn: I was talking about the changes below this line.
I don't think we need any of those changes.

sound/hybrid_font.h
```diff
@@ -230,18 +230,72 @@ class HybridFont : public SaberBase, public Looper {
     if (loop) hum_player_->PlayLoop(loop);
   }
 
+  int saved_player_to_restart_ = -1 ;
```
profezzorn, line 233: I don't think this saved player to restart is the right way to do it.
In fact, it could potentially cause bugs since this player could be playing something completely different by the time we get around to try  to use it.

Either we should do the simplest possible thing, which is just to kill effects of the same type as the one we're now trying to play....

OR

We should do the full-on-complicated thing which, which means looking for any effect which is being played by multiple players right now, rank them by time left to play and kill the one with the smallest number, but that also means making sure that we don't hit vital effects, like effects with ref counts, looping effects, or an effect with a next pointer.

profezzorn:
I created my own version of this. It's checked in and compiles, but not well tested.
If you have a chance to test it, you can enable it with KILL_OLD_PLAYERS.
If it works well, I can enable it by default.

NoSloppy:
OK, tested with adding some printouts.
Didn't compile for me at first with `#ifdef VERSION_MAJOR >= 4`
Made it just `#if` and it's happy.

Was still getting duds when running out of players. 
With each dud, I see the wav_players iteration loop running, but never catch a "killable" one to stop.
If this line is bypassed, I see the rest of the process run.
 `wav_players[unit].current_file_id().GetEffect()->GetKillable())`
 but of course that's ignoring the important part of killable or not.
 

How can the wav_player that's chosen to kill off be printed to verify which it was?
```
  if (p) {
    p->set_fade_time(0.001);
    p->FadeAndStop(); 
    ```
Not sure the syntax to print what p was here.

profezzorn:
Did you have a chance to test the latest version to see if it works properly?

NoSloppy:
> Did you have a chance to test the latest version to see if it works properly?

I can tonight

NoSloppy:
> Did you have a chance to test the latest version to see if it works properly?

I did and it's perfectly selecting the player playing the effect wav with shortest remaining time EVERY time.
No duds. Sweet!

profezzorn:
Nice, now we just need to figure out if it should be on by default.

NoSloppy:
Unless it's using too much memory, I don't know of a drawback to not use all the time. It makes things feel much snappier.

PR #405, opened 2022-02-09
Bc_buttons.  alt blade ctrl, spam blast

NoSloppy:


props/saber_BC_buttons.h
```diff
@@ -680,8 +737,19 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
         //Don't change preset if in colorchange mode
         if (SaberBase::GetColorChangeMode() != SaberBase::COLOR_CHANGE_MODE_NONE) return false;
           previous_preset_fast();
+      } else {
+        // Alt blade control
+        hybrid_font.DoEffect(EFFECT_USER4, 0);
+      if (!alt_blade_on_) {
+          hybrid_font.PlayCommon(&SFX_out);
```
profezzorn, line 744: WavLen and friends won't work properly if you play the sound here.
You really should override SB_Effect() and play the sound from there instead.

props/saber_BC_buttons.h
```diff
@@ -571,6 +611,10 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
         last_twist_ = millis();
         saber_off_time_ = millis();
         battle_mode_ = false;
+        if (alt_blade_on_) {
+          hybrid_font.DoEffect(EFFECT_USER4, 0);
+          alt_blade_on_ = false;
```
profezzorn, line 616: I think this is confusing, since SB_Effect() will also modify alt_blade_on_.
I think you need to decide if SB_Effect is doing this, or if it should be done where DoEffect is called.

props/saber_BC_buttons.h
```diff
@@ -913,7 +980,17 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
         SaberBase::DoEffect(EFFECT_POWERSAVE, 0);
         return true;
       } else {
+  #ifndef BC_SWAP
+        sound_library_.SayZoomingIn();
+        if (!spam_blast_) {
```
profezzorn, line 985: this whole if statement can be written as:

spam_blast_ = !spam_blast_;

profezzorn:
Is this still in progress?

NoSloppy:
Life happened and other ProffieOS distractions. I'll pick up on it.

profezzorn:
No hurry, I'm just checking to make sure it's not just sitting around for not reason.

props/saber_BC_buttons.h
NoSloppy: ok so this just flips the bool

props/saber_BC_buttons.h
NoSloppy: I was under the impression that DoEffect triggers the blade animation, while SB_Effect plays the sound, no?

NoSloppy:
While this was fun, I dont think it's needed going forward. 
1 - The advent of some OS7 features will handle an alternate blade better than this.
2 -  Spamming blaster blocks is not really mainstream, and dependent on killing oldest wav player anyway, so that should probably get focus first.

PR #407, opened 2022-02-18
Fixes for lockup behavior outside Battle Mode

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -5295,7 +5300,7 @@ SaberFett263Buttons() : PropBase() {}
           return true;
         }
 #endif
-        if (!battle_mode_ || swinging_) {
+        if ((!battle_mode_) || (battle_mode_ && swinging_)) {
```
profezzorn, line 5303: This does exactly the same as what you had before, but less clearly.

props/saber_fett263_buttons.h
```diff
@@ -1664,15 +1664,20 @@ SaberFett263Buttons() : PropBase() {}
...
-      } else {
-        SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
-        SaberBase::DoBeginLockup();
-        auto_lockup_on_ = true;
-        clash_type_ = CLASH_NONE;
-      }
+        if (clash_type_ == CLASH_BATTLE_MODE) {
+          if (SaberBase::GetClashStrength() < saved_gesture_control.clashdetect) {
+            SaberBase::DoClash();
+            clash_type_ = CLASH_NONE;
```
profezzorn, line 1670: This line is the same in both branches, so it should probably go after the if-else.

props/saber_fett263_buttons.h
fdarosa2663: It fixed a "hole", If it's not battle mode this should only do a clash ever, but if it is battle mode and the blade doesn't stop swinging then it should do a clash, the previous would let it through this check outside battle_mode and create a lockup if you had a clash without the swing stopping.  I added a comment to help clarify.

props/saber_fett263_buttons.h
fdarosa2663: consolidated

props/saber_fett263_buttons.h
```diff
@@ -1664,15 +1664,20 @@ SaberFett263Buttons() : PropBase() {}
...
+          } else {
+            SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
+            SaberBase::DoBeginLockup();
+            auto_lockup_on_ = true;
+            clash_type_ = CLASH_NONE;
+          }
+	} else {
+          SaberBase::DoClash();
```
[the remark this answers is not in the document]
fdarosa2663: fixed, thanks

props/saber_fett263_buttons.h
profezzorn: Ok, time for logic tables:

B=battle_mode, S=swinging_, T=true, F=false

battle_mode:
__|_S__!S_
 B|   T    T 
!B|   F    F

swinging_:
__|_S__!S_
 B|   T    F 
!B|   T    F

!battle_mode_
__|_S__!S_
 B|   F    F 
!B|   T    T

!battle_mode || swinging_
__|_S__!S_
 B|   T    F 
!B|   T    T

battle_mode && swinging_
__|_S__!S_
 B|   T    F 
!B|   F    F

(!battle_mode) || (battle_mode_ && swinging_)
__|_S__!S_
 B|   T    F 
!B|   T    T

Assuming you get what the logic tables mean, you'll see that 
     (!battle_mode || swinging_)
and
     ((!battle_mode_) || (battle_mode_ && swinging_))
have exactly the same results in all cases, so it's not possible for this change to change any behavior, and it looked better before.

fdarosa2663: Makes sense (or should have I guess), I reverted.

PR #412, opened 2022-04-28
ProffieOSHumDelay

NoSloppy:


sound/hybrid_font.h
```diff
@@ -395,43 +399,49 @@ class HybridFont : public SaberBase, public Looper {
...
-	STDOUT << "HUM fade-in time: " << hum_fade_in_ << "\n";
-      }
-      else if (font_config.humStart && tmp) {
-        int delay_ms = 1000 * tmp->length() - font_config.humStart;
-        if (delay_ms > 0 && delay_ms < 30000) {
-          hum_start_ += delay_ms;
+        hum_fade_in_ = tmp->length();
+        STDOUT << "HUM fade-in time: " << hum_fade_in_ << "\n";
+      } else {
+        if (font_config.ProffieOSHumDelay && tmp && font_config.ProffieOSHumDelay > 0 && font_config.ProffieOSHumDelay < 30000) {
```
profezzorn, line 433: This part is not needed:
font_config.ProffieOSHumDelay < 30000

NoSloppy: ok removed

sound/hybrid_font.h
```diff
@@ -395,43 +399,49 @@ class HybridFont : public SaberBase, public Looper {
...
-	hum_fade_in_ = tmp->length();
-	STDOUT << "HUM fade-in time: " << hum_fade_in_ << "\n";
-      }
-      else if (font_config.humStart && tmp) {
-        int delay_ms = 1000 * tmp->length() - font_config.humStart;
-        if (delay_ms > 0 && delay_ms < 30000) {
-          hum_start_ += delay_ms;
+        hum_fade_in_ = tmp->length();
+        STDOUT << "HUM fade-in time: " << hum_fade_in_ << "\n";
+      } else {
```
profezzorn, line 432: Instead of:

} else {
   if () {
   }
}

it should be:

} else if (....) {
}

Same for the next else-if below.

sound/hybrid_font.h
```diff
@@ -395,43 +399,49 @@ class HybridFont : public SaberBase, public Looper {
...
+        if (font_config.ProffieOSHumDelay && tmp && font_config.ProffieOSHumDelay > 0) {
```
profezzorn, line 433: just: if (font_config.ProffieOSHumDelay > 0)
We don't use "tmp" here so, no need to check that it is non-null.

PR #413, opened 2022-04-28
Add Subtract<> function

fdarosa2663:


functions/subtract.h
```diff
@@ -0,0 +1,40 @@
...
+    svfb_.run(blade);
+  }
+  int calculate(BladeBase* blade) {
+    return (svfa_.calculate(blade) - svfb_.calculate(blade));
+  }
+ private:
+  PONUA SVFA svfa_;
+  PONUA SVFB svfb_;
+};
+
```
profezzorn, line 39: This won't work because it's missing the selection code.

fdarosa2663: Is this what is missing, wasn't sure how the SVF was working
https://github.com/profezzorn/ProffieOS/blob/master/functions/sum.h#L40

profezzorn: Basically yes, but you're probably better off using the code from the Mult<> function. Sum<> has some extra code to allow arbitrary number of arguments, which we probably don't want/need for Subtract<>.

functions/subtract.h
```diff
@@ -0,0 +1,40 @@
+#ifndef FUNCTIONS_SUBTRACT_H
+#define FUNCTIONS_SUBTRACT_H
+
+// Usage: SUBTRACT<A, B>
```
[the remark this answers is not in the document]
fdarosa2663: fixed

functions/subtract.h
```diff
@@ -0,0 +1,40 @@
...
+// Subtracts B from A (A - B)
+// A, B: INTEGER
```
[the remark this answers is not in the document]
fdarosa2663: fixed

functions/subtract.h
```diff
@@ -0,0 +1,40 @@
...
+// return value: INTEGER
```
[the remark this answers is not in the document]
fdarosa2663: fixed

functions/subtract.h
```diff
@@ -0,0 +1,46 @@
...
+
+template<class SVFA, class SVFB>
+class SubtractSVF {
+ public:
+  void run(BladeBase* blade) {
+    svfa_.run(blade);
+    svfb_.run(blade);
+  }
+  int calculate(BladeBase* blade) {
+    return (svfa_.calculate(blade) - svfb_.calculate(blade));
```
profezzorn, line 33: remove extra parenthesis

fdarosa2663: removed

PR #415, opened 2022-04-28
Add Divide<>

fdarosa2663:


functions/divide.h
```diff
@@ -0,0 +1,56 @@
...
+public:
+  void run(BladeBase* blade) {
+    f_.run(blade);
+    v_.run(blade);
+  }
+  int getInteger(int led) {
+    if (v_.getInteger(led) == 0) {
+      return 0;
+    } else {
+      return (f_.getInteger(led) / v_.getInteger(led));
```
profezzorn, line 23: remove extra parenthesis
(same for the SVF function below.)

functions/divide.h
```diff
@@ -0,0 +1,56 @@
...
+
+template<class F, class V>
+class DivideBase {
+public:
+  void run(BladeBase* blade) {
+    f_.run(blade);
+    v_.run(blade);
+  }
+  int getInteger(int led) {
+    if (v_.getInteger(led) == 0) {
```
profezzorn, line 20: We shouldn't call v_.getInteger(led) twice.
We should call it once and store the result in a local variable.
(Same for the SVF function below.)

functions/divide.h
fdarosa2663: updated

functions/divide.h
```diff
@@ -0,0 +1,58 @@
...
+  void run(BladeBase* blade) {
+    f_.run(blade);
+    v_.run(blade);
+  }
+  int getInteger(int led) {
+    int v = v_.getInteger(led);
+    if (v == 0) {
+      return 0;
+    } else {
+      return f_.getInteger(led) / v;
```
profezzorn, line 24: I wonder if we want this to be f * 32768 / v  ?
If it's used mostly for fixed-point math where the values are between 0 and 32768, that would make more sense.
If it's used for other things, then it would not.

fdarosa2663: The original intent is not fixed-point, the uses I have been testing are to divide one function by another to get the actual value.

profezzorn:
I think this is ready to commit, but it needs to be merged/rebased with regards to the other PR I just committed.

fdarosa2663:
Github giving a hard time will just resubmit from clean branch

PR #416, opened 2022-05-02
add Divide<>

fdarosa2663:


functions/divide.h
```diff
@@ -0,0 +1,58 @@
+#ifndef FUNCTIONS_DIVIDE_H
+#define FUNCTIONS_DIVIDE_H
+
+// Usage: Divide<F, V>
+// Divide F by V
+// If V = 0, returns 0
+// F, V: FUNCTION, 
+// return value: FUNCTION
+
```
profezzorn, line 9: We should probably at least add a comment about Divide<> not being fixed-point since Mult<> *is* fixed-point, and it would normally be reasonable to assume that Divide<> and Mult<> would be similar in behavior.

fdarosa2663: let me know if that's better or if you have better wording

profezzorn: I don't think people will understand that.
How about something like:

Please note that Divide<> isn't an exact inverse of Mult<> because mult uses fixed-point mathematics
(it divides the result by 32768) while Divide<> doesn't, it just returns F / V.

fdarosa2663: Updated

PR #417, opened 2022-05-11
Add AUDIO_CLASH_SUPPRESSION_LEVEL

fdarosa2663:


props/prop_base.h
```diff
@@ -17,6 +17,11 @@
 #undef SAVE_VOLUME
 #endif
 
+#ifndef AUDIO_CLASH_SUPPRESSION_LEVEL
+// Account for Audio Volume in Clash Detection (range 1 ~ 50)
+#define AUDIO_CLASH_SUPPRESSION_LEVEL 1
```
profezzorn, line 22: If the default value is this low, is there really any point?

props/prop_base.h
```diff
@@ -721,9 +726,9 @@ class PropBase : CommandParser, Looper, protected SaberBase {
     } else {
       v = diff.len();
     }
-    // If we're spinning the saber, require a stronger acceleration
-    // to activate the clash.
-    if (v > CLASH_THRESHOLD_G + fusor.gyro().len() / 200.0) {
+    // If we're spinning the saber or if loud sounds are playing, 
+    // require a stronger acceleration to activate the clash.
+    if (v > (CLASH_THRESHOLD_G + fusor.gyro().len() / 200.0) + (dynamic_mixer.audio_volume() * (AUDIO_CLASH_SUPPRESSION_LEVEL * 0.000001))) {
```
profezzorn, line 731: This needs some line-breaking to make the line shorter.
(Probably after the plusses)

props/prop_base.h
fdarosa2663: I was originally thinking you'd want it to be almost nothing if the define was missing to keep existing behavior from previous versions

props/prop_base.h
fdarosa2663: added line break

PR #419, opened 2022-05-12
Add TransitionPulseL<>

fdarosa2663:


styles/transition_pulse.h
```diff
@@ -0,0 +1,53 @@
...
+    }
+    if (run_) {
+      transition_.run(blade);
+      if (transition_.done()) run_ = false;
+    }
+    last_detected_blade_effect = nullptr;
+    if (!run_) {
+      // All effects now wake the blade up again if needed, so it's safe to
+      // always return TRANSPARENT_UNTIL_IGNITION
+      return LayerRunResult::TRANSPARENT_UNTIL_IGNITION;
```
profezzorn, line 31: This return (and the if statement) needs to go away, because it's not valid for pulses.

PR #420, opened 2022-05-16
DISABLE defines

fdarosa2663:
Allow users and/or installers to disable the following features with defines:
CHANGE FONT (outside Edit Mode)
CHANGE STYLE (outside Edit Mode)
COPY PRESET (outside Edit Mode)
Multi-Blast Mode


fdarosa2663:
Just pinging this, it's some defines for my prop to remove some features that users/installers asked about and updated comments

PR #421, opened 2022-05-21
OLED images update

NoSloppy:
Add OLED effects and option to sync to wav.

### DIFF
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -23,7 +23,6 @@ IMAGE_FILESET(clipin);
 IMAGE_FILESET(clipout);
 IMAGE_FILESET(destruct); // BC TODO
 
-
 enum Screen {
   SCREEN_UNSET,
   SCREEN_STARTUP,
@@ -287,11 +286,11 @@ public:
   int FillFrameBuffer(bool advance) override {
     if (advance) {
       if (next_screen_ != SCREEN_UNSET) {
-  screen_ = next_screen_;
-  next_screen_ = SCREEN_UNSET;
-  t_ = 0;
+	screen_ = next_screen_;
+	next_screen_ = SCREEN_UNSET;
+	t_ = 0;
       } else {
-  t_ += last_delay_;
+	t_ += last_delay_;
       }
     }
     return last_delay_ = FillFrameBuffer2(advance);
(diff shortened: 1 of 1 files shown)


display/ssd1306.h
```diff
@@ -16,6 +16,12 @@ IMAGE_FILESET(blst);
 IMAGE_FILESET(lock);
 IMAGE_FILESET(force);
 IMAGE_FILESET(idle);
+IMAGE_FILESET(reload);
```
profezzorn, line 19: I'm trying to decide the best way to have these not use any space for props that have displays, but are not guns.

Right now, the best I can come up with is to use defines, which gets a little ugly.
Basically we would want a set of defines which specifies which EFFECT_** entries to use.
For each EFFECT_** entry, we would also add a define HAVE_EFFECT_**, which can be use to #ifdef filesets and the case statements.

To make it easier to manage the defines, I was thinking that I could add sections to the prop files, similar to how the config files have sections. Then I could include the prop file early in the compilation progress, and let the prop file help define which EFFECT_** we need.

Not exactly pretty, but it could work.

display/ssd1306.h
```diff
@@ -468,22 +474,80 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
+    }
+  }
+ 
+  void SB_Effect2(EffectType effect, float location) override {
+    switch (effect) {
+      case EFFECT_DESTRUCT:
+        ShowFile(&IMG_destruct, SaberBase::sound_length * 1000);
+        return;
+        // STDOUT.println(SaberBase::sound_length * 1000);
+#ifdef OLED_SYNCED_EFFECTS
```
profezzorn, line 524: Instead of having a define for this, a config variable might be better.
Not sure if that config variable would need to be per-effect or not though....

Or maybe we use the sound length only if the relevant duration variable is unset?

profezzorn (commented):
You're doing two things in one PR again.
Adding new effects and introducing synchronization would ideally be two separate PRs.

display/ssd1306.h
NoSloppy: In the previous PR, That's what I thought too. A setting of zero in the config.ini could indicate to use WavLen.
I had typed this:


> instead of another define, I tried something like this where the user could set the config.ini value to zero and it would use wavlen.
Then I thought maybe just commenting it out in config.ini would be better, but I don't know how to check if it was read from the file as a VariableOp or not.

  case EFFECT_CLASH:
    ShowFile(&IMG_clsh, (font_config.ProffieOSClashImageDuration = 0) ? SaberBase::sound_length * 1000 : **iterateVariables VariableOp-values somehow goes here** );
    return;
   
  float dur;
  case EFFECT_BLAST:
    if (!font_config.ProffieOSBlastImageDuration) {         // can't check if this entry exists this way
      dur = (SaberBase::sound_length * 1000);
    } else { 
      dur = font_config.ProffieOSBlastImageDuration;
    }
    ShowFile(&IMG_blst, dur);

NoSloppy: or a hybrid of those 2. 
Can we default the time to zero
`CONFIG_VARIABLE2(ProffieOSBlastImageDuration, 0.0f);`
 and then have the config.ini part describe like
 ```
# These are the durations the images will play for.
# Enter a custom value to manually set a duration,
# or leave commented out to automatically sync to the effect sound.
# ProffieOSBlastImageDuration=1000.0
```
float dur;
case EFFECT_BLAST:
if (ProffieOSBlastImageDuration = 0) {
    dur = (SaberBase::sound_length * 1000);
} else {
dur = font_config.ProffieOSBlastImageDuration;
}
ShowFile(&IMG_blst, dur);
```

NoSloppy:
Closing and resubmitting separately.

PR #423, opened 2022-05-24
Prevent PLAYBACK_LOOP in TrackPlayer if no track is present

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -1797,6 +1797,7 @@ SaberFett263Buttons() : PropBase() {}
       if (!track_player_) {
         if (track_num_ <= 0 && track_mode_ == PLAYBACK_LOOP) {
           StartOrStopTrack();
+          if (!track_player_->isPlaying()) track_mode_ = PLAYBACK_OFF;
```
profezzorn, line 1800: Should really check that track_player_ isn't null as well.

fdarosa2663: added

PR #424, opened 2022-05-25
Enable Force/Quote Player while OFF

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -4602,6 +4615,17 @@ SaberFett263Buttons() : PropBase() {}
         return true;
 #endif
 
+      case EVENTID(BUTTON_POWER, EVENT_THIRD_SAVED_CLICK_SHORT, MODE_OFF):
+        if (SFX_quote) {
```
profezzorn, line 4619: Create a helper function for this so we don't have to repeat it again below.

fdarosa2663: Added

PR #425, opened 2022-06-14
BC buttons: Add sequential quote play toggle

NoSloppy:


NoSloppy:
Actually, now I see there's something wrong here. Toggling back to NOT use SelectNext(), the quotes are not randomized again, but instead play the last selected every time.....hmmm.
What needs to make them random again?

props/saber_BC_buttons.h
```diff
@@ -912,6 +923,14 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
       if (fusor.angle1() >  M_PI / 3) {
         SaberBase::DoEffect(EFFECT_POWERSAVE, 0);
         return true;
+      } else if (fusor.angle1() < - M_PI / 4) {
+      // pointing down
+        sequential_quote_ = !sequential_quote_;
+        beeper.Beep(0.1, 1300);
```
profezzorn, line 929: Might want to make these beeps different when sequential is turned on vs. off.
Like maybe hi/lo/hi/lo is on and lo/hi/lo is off?
Or maybe use sound library?

NoSloppy: Good Idea, thanks.
Also updated instructions for this feature.

props/saber_BC_buttons.h
```diff
@@ -912,6 +928,18 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
+      } else if (fusor.angle1() < - M_PI / 4) {
+      // pointing down
+        sequential_quote_ = !sequential_quote_;
+        if (sequential_quote_) {
+          beeper.Beep(0.1, 600);
+          beeper.Beep(0.1, 950);
+          beeper.Beep(0.1, 1300);
+          beeper.Beep(0.1, 600);
+        } else {
+          sound_library_.SayRandom();
```
profezzorn, line 940: How about "random enabled" / "random disabled" ?

NoSloppy: Love that, thanks.

profezzorn: You can move the SayRandom() to before the if statement since it is the same for both branches.

NoSloppy: Yes, thank you.

PR #428, opened 2022-06-18
Add #define BOOT_VOLUME option

NoSloppy:


sound/dynamic_mixer.h
```diff
@@ -185,7 +185,10 @@ template<int N> class AudioDynamicMixer : public ProffieOSAudioStream, Looper {
   int32_t peak_sum_ = 0;
   int32_t peak_ = 0;
   int32_t num_samples_ = 0;
-  int32_t volume_ = VOLUME;
+#ifndef BOOT_VOLUME
```
profezzorn, line 188: The #ifdef should go in ProffieOS.ino so that BOOT_VOLUME can be used anywhere.

sound/dynamic_mixer.h
```diff
@@ -185,7 +185,10 @@ template<int N> class AudioDynamicMixer : public ProffieOSAudioStream, Looper {
...
+  #define BOOT_VOLUME VOLUME
```
profezzorn, line 189: No indentation for #defines

sound/dynamic_mixer.h
NoSloppy: ah. Ok done.

sound/dynamic_mixer.h
NoSloppy: done

PR #429, opened 2022-06-21
Additional OLED images and sync to wavs

NoSloppy:
Github indentation might be weird.

### DIFF
--- a/styles/tests.cpp
+++ b/styles/tests.cpp
@@ -117,6 +117,7 @@ Monitoring monitor;
 #include "../common/color.h"
 #include "../blades/blade_base.h"
 #include "cylon.h"
+#include "../common/arg_parser.h"
 #include "style_ptr.h"
 #include "colors.h"
 #include "inout_helper.h"
@@ -165,7 +166,6 @@ Monitoring monitor;
 #include "fire.h"
 #include "sparkle.h"
 #include "../common/command_parser.h"
-#include "../common/arg_parser.h"
 CommandParser* parsers = NULL;
 ArgParserInterface* CurrentArgParser;
 


profezzorn:
So, In ProffieOS 6.x, I added some basic stuff for using custom display controllers. I think we need to write a custom controller for blasters so that it's only used for blasters. Might take a little more coding to make it actually work though...

NoSloppy:
Cool. So maybe just the in, out, preon, and pstoff ones for saber for now?

NoSloppy:
In my tests, defining OLED_USE_BLASTER_IMAGES used a total of 32 additional bytes of memory.

sound/hybrid_font.h
```diff
@@ -22,6 +22,21 @@ class FontConfigFile : public ConfigFile {
     CONFIG_VARIABLE2(ProffieOSBlastImageDuration, 1000.0f);
     CONFIG_VARIABLE2(ProffieOSClashImageDuration, 500.0f);
     CONFIG_VARIABLE2(ProffieOSForceImageDuration, 1000.0f);
+    CONFIG_VARIABLE2(ProffieOSPreonImageDuration, 5000.0f);
```
profezzorn, line 25: Seems weird to have a duration for preon, shouldn't it run until it turns on?

display/ssd1306.h
```diff
@@ -450,14 +468,14 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
   void SB_Effect(EffectType effect, float location) override {
     switch (effect) {
       case EFFECT_BOOT:
-	if (IMG_boot) {
-	  ShowFile(&IMG_boot, font_config.ProffieOSFontImageDuration);
+      	if (IMG_boot) {
+	  ShowFile(&IMG_boot, font_config.ProffieOSBootImageDuration);
```
profezzorn, line 472: For backwards compatibility, I think this needs to default to ProffieOSFontImageDuration if ProffieOSBootImageDuration isn't set.

display/ssd1306.h
```diff
@@ -468,26 +486,63 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
+        ShowFile(&IMG_blst, font_config.ProffieOSBlastImageDuration);
+        return;
+      case EFFECT_CLASH:
+        ShowFile(&IMG_clsh, font_config.ProffieOSClashImageDuration);
+        return;
+      case EFFECT_FORCE:
+        ShowFile(&IMG_force, font_config.ProffieOSForceImageDuration);
+        return;
+      case EFFECT_PREON:
+        ShowFile(&IMG_preon, ProffieOSPreonImageDuration;);                                                                                                // BC another one that doesn't like float....?
```
profezzorn, line 513: This seems untested, because this won't compile.
(You have a stray semicolon in there..)

display/ssd1306.h
```diff
@@ -468,26 +486,63 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
-	break;
+      	ShowDefault(true);
+      	break;
+
+      default: break;
+    }
+  }
+
 
+ void SB_Effect2(EffectType effect, float location) override {
```
profezzorn, line 501: I'm not sure why this was moved to SB_Effect2, but it does mean that we can know the sound lengths that went with these effects, and we could potentially use that to set the durations below...

NoSloppy: All of this was an incremental step. I have further changes pending, but you asked for just the effects like this as a baby step towards the end goal; to just have all of them follow the wav duration (if #define OLED_SYNCED_EFFECTS or not).
I will revisit all this ASAP.

profezzorn: That's all fine, but it doesn't make sense to add a new variable for how long to show the preon image.
If you don't want to add the code to make it depend on the wav, then just use the fontImageDuration so you don't have to add a new variable that doesn't really make sense.

NoSloppy: >  If you don't want to add the code to make it depend on the wav, 

I do. I think that's a choice way to handle these, auto timed to the effect itself, no?
Again, trying to get some time. Will need to revisit this and submit what I have working/happy on my end.

display/ssd1306.h
NoSloppy: Not sure I understand this one.
Boot and font were the original 2 images supported.
And BootImage has a default set in hybrid_font incase not set in the font config.ini.

NoSloppy:
I'm sure there's a better way to consolidate all the float-dur stuff, I don't know the syntax.
something like:
  void Set_Duration(EffectType effect, float dur) { 
    font_config.ProffieOS *Effect ImageDuration == 0 ? SaberBase::sound_length * 1000 : font_config.ProffieOS *Effect ImageDuration;
    ShowFile(&IMG_*Effect), round (dur);
  }

NoSloppy:
I'm sure there's a better way to consolidate all the float-dur stuff, I don't know the syntax.
something like:
  void Set_Duration(EffectType effect, float dur) { 
    font_config.ProffieOS *Effect ImageDuration == 0 ? SaberBase::sound_length * 1000 : font_config.ProffieOS *Effect ImageDuration;
    ShowFile(&IMG_*Effect), round (dur);
  }

PR #430, opened 2022-06-22
Add QuickSelect

fdarosa2663:


fdarosa2663:
In order to keep PRs small I have single pieces being done one-at-a-time, but there are several in the queue for my prop if we can merge this one.

props/saber_fett263_buttons.h
```diff
@@ -4311,7 +4317,11 @@ SaberFett263Buttons() : PropBase() {}
 
   RefPtr<BufferedWavPlayer> wav_player;
 
+#ifndef FETT263_QUICK_SELECT_ON_BOOT
   float current_menu_angle_ = 0.0;
+#else
+  float current_menu_angle_ = fusor.angle2();
```
profezzorn, line 4323: I don't think this will work, as this assignment will be executed before the fusor has had a chance to actually do anything.

props/saber_fett263_buttons.h
```diff
@@ -5654,7 +5679,11 @@ SaberFett263Buttons() : PropBase() {}
   bool auto_lockup_on_ = false; // Battle Mode Lockup active
   bool auto_melt_on_ = false; // Battle Mode Melt/Drag active
   bool battle_mode_ = false; // Battle Mode active
+#ifndef FETT263_QUICK_SELECT_ON_BOOT
```
profezzorn, line 5682: It's better to use #ifdef instead of #ifndef for these. (and swap the position of the two cases.)  Inverted logic takes a few extra moments to parse for anybody who reads the code.

props/saber_fett263_buttons.h
fdarosa2663: I have updated and moved the current_menu_angle_ into Loop with a delay for boot to allow fusor startup.

props/saber_fett263_buttons.h

PR #436, opened 2022-07-08
Welcome message edit

NoSloppy:
Should we have something like this?


profezzorn:
No.
We should add a link to a wiki page instead.

PR #437, opened 2022-07-14
Fix blaster jam check so that 0 percent actually is 0 percent

shadowfoil:
rand mod check on blaster jam percentage was incorrectly comparing in range 0-99 instead of 1-100. 0 percent chance should never jam.

### DIFF
--- a/props/blaster.h
+++ b/props/blaster.h
@@ -108,7 +108,7 @@ public:
   }
 
   virtual bool CheckJam(int percent) {
-    int random = rand() % 100;
+    int random = rand() % 100 + 1;
     return random <= percent ? true : false;
   }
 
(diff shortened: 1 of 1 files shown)


props/blaster.h
```diff
@@ -108,7 +108,7 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
   }
 
   virtual bool CheckJam(int percent) {
-    int random = rand() % 100;
+    int random = rand() % 100 + 1;
```
profezzorn, line 111: Good find, but it would be better to just change the next line to:

return random < percent;

shadowfoil: ah yes, that was the original thing I was going to do but I thought 100 percent passed in wouldn't function correctly but I think it will after looking again

shadowfoil:
There we go, fix the comparator and removed the redundant ternary operator

PR #438, opened 2022-07-18
Add TwistAcceleration<>, update TwistAngle comments

fdarosa2663:


functions/twist_angle.h
```diff
@@ -24,4 +24,23 @@ class TwistAngleSVF {
...
+// Returns 0-32768 based on acceleration of twist in one direction
+// MAX : Maximum acceleration needed to return 32768
+// returned value: FUNCTION, same for all leds
+
+template<class MAX>
+class TwistAccelerationSVF {
+ public:
+  void run(BladeBase* blade) { max_.run(blade); }
+  int calculate(BladeBase* blade) {
+   float v = fusor.twist_accel() * 100 / max_.calculate(blade);
```
profezzorn, line 37: Why times 100?
I think twist_accel() is in degrees per second per second. (or is it in radians per second per second?)
If it is in degrees per second per second, maybe we should multiply by 32768 / 360, as that would make 32768 mean one rotation per second per second.

fdarosa2663: gyro_slope().x was returning 0 ~ 32000 values when I added a print, but to allow users to lower the threshold I have it multiplied by 100 so we can divide by MAX for customization.  I can try 32768 / 360 and update.

fdarosa2663: updated both files and tested and it's working as envisioned.

common/fuse.h
```diff
@@ -296,6 +296,11 @@ class Fusor : public Looper {
     return sqrtf(gyro_slope_.z * gyro_slope_.z + gyro_slope_.y * gyro_slope_.y) * (M_PI / 180);
   }
 
+  // Acceleration into twist (one direction) in degrees per second per second
+  float twist_accel() {
+    return fabs(gyro_slope().x);
```
profezzorn, line 301: no fabs here
If you need to take the absolute value, we can maybe do that in the style, but not here.

common/fuse.h
```diff
@@ -296,6 +296,11 @@ class Fusor : public Looper {
     return sqrtf(gyro_slope_.z * gyro_slope_.z + gyro_slope_.y * gyro_slope_.y) * (M_PI / 180);
   }
 
+  // Acceleration into twist (one direction) in radians per second per second
+  float twist_accel() {
+    return sqrtf(gyro_slope().x * gyro_slope().x);
```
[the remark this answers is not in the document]
profezzorn: Slope is naturally a function that can be both positive and negative.
Using fabs() isn't a solution, it just confuses things.
Even if we do need fabs() in the style, it should go there, not here.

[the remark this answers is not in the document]
fdarosa2663: ok, moved fabs() into the function.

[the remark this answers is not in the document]
profezzorn: I'm still not convinced that we need/want a fabs()
Do you not care if the action is speeding up or slowing down?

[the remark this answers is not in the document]
fdarosa2663: Since all other functions, SwingSpeed, BladeAngle, SwingAcceleration, etc return 0 ~ 32768 I kept this in the same vein.  In testing, particularly if used in Shapes, negative numbers created very odd movement or results.
Nearly all of my intended uses are looking for the acceleration of the twist action or the acceleration up to it, not really slowing down.  
I suppose we can remove the fabs() but we may want to add an Absolute<> function separately so it can be used in Shapes or other instances where a negative number causes issues or where the slowing down is unwanted.

functions/twist_angle.h
```diff
@@ -24,4 +24,23 @@ class TwistAngleSVF {
...
+// Returns 0-32768 based on acceleration of twist in one direction
+// MAX : Maximum acceleration needed to return 32768
+// returned value: FUNCTION, same for all leds
+
+template<class MAX>
+class TwistAccelerationSVF {
+ public:
+  void run(BladeBase* blade) { max_.run(blade); }
+  int calculate(BladeBase* blade) {
+   float v = (fabs(fusor.twist_accel()) * 32768 / 360) / max_.calculate(blade);
```
profezzorn, line 37: I still think it's super weird to have a fabs() in here.
Might make more sense to multiply it by -1 if fusor.gyro().z < 0 ?
(This would make positive values mean that it is speeding up, regardless of whether it is twisting to the left or to the right.)

fdarosa2663: Updated

functions/twist_angle.h
```diff
@@ -24,4 +24,25 @@ class TwistAngleSVF {
...
+// MAX : Maximum acceleration needed to return 32768
+// returned value: FUNCTION, same for all leds
+
+template<class MAX>
+class TwistAccelerationSVF {
+ public:
+  void run(BladeBase* blade) { max_.run(blade); }
+  int calculate(BladeBase* blade) {
+    float accel = fusor.twist_accel();
+    if (accel < 0) accel *= -1;
```
profezzorn, line 38: This is the same as fabs() and not what I meant.
I meant:
  if (fusor.gyro().z < 0) accel = -accel;

(if that doesn't work, try > instead of <)

fdarosa2663: Fixed

PR #439, opened 2022-07-20
add Wiki link for Serial Commands

NoSloppy:
better?

### DIFF
--- a/common/serial.h
+++ b/common/serial.h
@@ -133,6 +133,8 @@ public:
       while (!SA::Connected()) YIELD();
       if (!SA::AlwaysConnected()) {
         STDOUT << "Welcome to ProffieOS " << version << ". Type 'help' for more info.\n";
+        STDOUT << "Serial Commands Wiki Page:\n";
+	STDOUT << "https://github.com/profezzorn/ProffieOS/wiki/Serial-Monitor-Commands\n";
       }
 
       while (SA::Connected()) {


common/serial.h
```diff
@@ -133,6 +133,8 @@ class Parser : Looper, StateMachine {
       while (!SA::Connected()) YIELD();
       if (!SA::AlwaysConnected()) {
         STDOUT << "Welcome to ProffieOS " << version << ". Type 'help' for more info.\n";
+        STDOUT << "Serial Commands Wiki Page:\n";
```
profezzorn, line 136: I'm thinking this should be what the "help" command prints.
Then we can get rid of all the rest of the "help" code.

NoSloppy: As in...cutting out every printout in Help() functions throughout the OS, and putting them on the Wiki page instead?
Would need to have sections like ifndef DISABLE_DIAGNOSTIC_COMMANDS, then the following will also work, etc...

profezzorn: The help command isn't very accurate about specifying which commands are available with which defines anyways.

NoSloppy: This would mean no internet access=no help. But it might save some memory, which I think is more popular than using the help command .

profezzorn: People can always print out the wiki page if they need it.

NoSloppy:
Separate PR coming

PR #441, opened 2022-07-27
LockupTrL updates

fdarosa2663:


styles/lockup.h
```diff
@@ -98,8 +98,8 @@ template<class BASE,
   class LOCKUP_SHAPE = Int<32768>, class DRAG_SHAPE = SmoothStep<Int<28671>, Int<4096>> >
   using Lockup = Layers<BASE, LockupL<LOCKUP, DRAG_COLOR, LOCKUP_SHAPE, DRAG_SHAPE>>;
 
-// Usage: LockupTr<BASE, COLOR, BeginTr, EndTr, LOCKUP_TYPE>
-// Or: LockupTrL<COLOR, BeginTr, EndTr, LOCKUP_TYPE>
+// Usage: LockupTr<BASE, COLOR, BeginTr, EndTr, LOCKUP_TYPE, F>
+// Or: LockupTrL<COLOR, BeginTr, EndTr, LOCKUP_TYPE, F>
```
profezzorn, line 102: Let's call it CONDITION instead of F.

profezzorn:
Have you tested how many bytes this change costs?

profezzorn (commented):
Have you measured how many bytes this change costs for a style where F just uses the default value?

fdarosa2663:
Just measured a single style, single lockup and it was 48 bytes more with F (CONDITION).  

I'll work on changing F to CONDITION tomorrow unless you want me to break back out into it's own separate style instead to save the bytes for existing styles that won't need this feature.

profezzorn:
48 bytes is not bad, and I can special-case it later if that's not good enough.

PR #448, opened 2022-09-11
Adding OS7 EFFECTS

fdarosa2663:


common/saber_base.h
```diff
@@ -38,6 +38,11 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(POWERSAVE)                    \
     DEFINE_EFFECT(BATTERY_LEVEL)                \
     DEFINE_EFFECT(FAST_ON)                      \
+    DEFINE_EFFECT(SECONDARY_IGNITION)		\
+    DEFINE_EFFECT(SECONDARY_RETRACTION)		\
+    DEFINE_EFFECT(QUOTE)			\
+    DEFINE_EFFECT(ALT_SOUND)			\
```
profezzorn, line 44: Some of these (particularly ALT_SOUND and TRANSITION_SOUND) could use a comment explaining what they are meant for.

fdarosa2663: these are placeholders, the actual EFFECT_ALT_SOUND and EFFECT_TRANSITION_SOUND effects haven't been added yet. The ALT_SOUND will be the effect related to alt/ font set up and TRANSITION_SOUND will be related to TrDoEffect but the alt/ font implementation affects exactly how both will work and be utilized.  I added these to the enum to keep them in front of us and simplify the PRs.  I can add comments once the alt/ font and new transition are finalized as there's a few ways they can still go.

fdarosa2663: I also need the enums in place for the controls in the prop to be submitted.

profezzorn: They are not placeholders.
They need names that either self-explanatory, or comments that explain how to use them.

fdarosa2663: Not "placeholders" but they don't actually do anything yet, they'll be self-explanatory when the features they need are added.  I added comments for intended use in the interim.

common/saber_base.h
```diff
@@ -38,6 +38,13 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(POWERSAVE)                    \
     DEFINE_EFFECT(BATTERY_LEVEL)                \
     DEFINE_EFFECT(FAST_ON)                      \
+    DEFINE_EFFECT(SECONDARY_IGNITION)		\
+    DEFINE_EFFECT(SECONDARY_RETRACTION)		\
+    DEFINE_EFFECT(QUOTE)			\
+    /* Change Alt font sounds */                \
```
profezzorn, line 44: This comment doesn't explain much, two of the words are already in the ALT_SOUND names.
If you read this (and nothing else) would you understand what it meant?
I don't think it needs a long explanation, but maybe something like:
/* Triggered when we change sets of sounds within the font from one alternative to another. */

common/saber_base.h
```diff
@@ -38,6 +38,13 @@ extern SaberBase* saberbases;
...
+    DEFINE_EFFECT(ALT_SOUND)			\
+    /* Transition Effect/Pulse Sound */         \
```
profezzorn, line 46: This comment is also pretty unhelpful.
Transition from what to what?
What would trigger this effect?

common/saber_base.h
fdarosa2663: ok, updated.  although this effect will be what triggers the change for the sets of sounds when it's triggered.  the actual trigger is style / prop specific.

common/saber_base.h
fdarosa2663: updated the comment, what triggers this effect depends entirely on the style itself, this is part of the new transition effect / pulse capabilities all tied to the alt font and tr sound parts we discussed.

common/saber_base.h
```diff
@@ -38,6 +38,13 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(POWERSAVE)                    \
     DEFINE_EFFECT(BATTERY_LEVEL)                \
     DEFINE_EFFECT(FAST_ON)                      \
+    DEFINE_EFFECT(SECONDARY_IGNITION)		\
+    DEFINE_EFFECT(SECONDARY_RETRACTION)		\
+    DEFINE_EFFECT(QUOTE)			\
+    /* Triggers the change for sets of sounds within the font from one alternative to another */                \
+    DEFINE_EFFECT(ALT_SOUND)			\
+    /* Optional sound used during Transition Effect or TransitionPulse */         \
```
profezzorn, line 46: If this effect is meant to be triggered by the style, then I think it might be a good idea to call that out.

common/saber_base.h
profezzorn: I agree that they don't do anything yet, but people shouldn't have to go read a bunch of code in order to have some idea what these values are for.

They should be able to look at this list and have a reasonable idea of what the purpose of the value is.

Of course, they might still need to go look at the code to understand exactly how/when something is used, but that is how code always works.

Remember; the goal isn't just to make code that works, but code that people can read, understand and learn from.

common/saber_base.h
fdarosa2663: added

common/saber_base.h
profezzorn: "a style, mini game or via prop" is basically the same as saying "called from anywhere", which is just not needed.

I thought this one was meant to be triggered from a style, but I guess I still don't know what it is meant to do, which means it still need better comments.

fdarosa2663: I have these both planned to be triggered from styles or mini games or optional defines within my prop depending on the desired use, they're multi-use depending.  The test styles and mini games I've built in my local version both use and (depending on the end result of the alt/ font capability) I have several ideas to make features in the prop that can utilize.  There not "single use".  
I think the comments for use will make more sense in the parts that aren't in place yet.

profezzorn: If you need generic names, then why not use something generic like USER1 or something?

fdarosa2663: The effect is not generic, the sounds are.  The sounds can be completely different for each font/style combination they are called from the transition as sound effects to enhance the transitions themselves; a fireball, a bolt of lightning, a mechanized transformation.  It's been a while but we discussed this already on theCrucible several months ago.
Because the sounds can be anything we want, they fit into the mini games concept as well as styles.  
If you recall I was also originally going to set the sound itself up in hybrid_font but you told me to leave only in my prop.  
I updated comments on both again, hopefully it's better?  I think no matter what comment I put in, it won't really make sense to anyone else until the other pieces are in place.

fdarosa2663:
simplified the TRANSITION_SOUND comment to original intent

common/saber_base.h
profezzorn: This is a much better description!

PR #451, opened 2022-09-16
Edit Mode OS7 additions

fdarosa2663:
Adding OS7 arguments and added missing OFF_OPTION_ARG to menu, sound file already exists in Voice Pack, the submenu was missing to use.

### DIFF
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -3822,8 +3822,7 @@ SaberFett263Buttons() : PropBase() {}
       case MENU_RETRACTION_OPTION2:
         if (SaberBase::IsOn()) {
           calc_ += (direction * 1000);
-          if (calc_ > 32768) calc_ = 0;
-          if (calc_ < 0) calc_ = 32768;
+          calc_ &= 0x7fff;
           SetInOut();
           sound_library_.SayNumber(calc_, SAY_WHOLE);
         }
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -184,8 +184,12 @@ Edit Mode*
   While in Edit Mode controls are as follows:
     Rotate Forward, Increase Value, Confirm "Yes" = Turn Right (Stepped)
       Increment by 5 (Fonts, Tracks, Blade Length) = Hold PWR + Turn Right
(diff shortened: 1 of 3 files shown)


props/saber_fett263_buttons.h
```diff
@@ -3712,6 +3818,16 @@ SaberFett263Buttons() : PropBase() {}
           sound_library_.SayNumber(calc_, SAY_WHOLE);
 	}
         break;
+      case MENU_IGNITION_OPTION2:
+      case MENU_RETRACTION_OPTION2:
+        if (SaberBase::IsOn()) {
+          calc_ += (direction * 1000);
+          if (calc_ > 32768) calc_ = 0;
```
profezzorn, line 3825: just do:
calc_ &= 0x7fff;

This can replace this line and the next one.

styles/edit_mode.h
```diff
@@ -49,4 +55,8 @@ template<int DEFAULT_VALUE = 300> using IgnitionTime = Scale<IsLessThan<IntArg<I
 // RetractionTime = RETRACTION_TIME_ARG - uses "Auto" feature with WavLen when = 0
 template<int DEFAULT_VALUE = 0> using RetractionTime = Scale<IsLessThan<IntArg<RETRACTION_TIME_ARG,DEFAULT_VALUE>,Int<1>>,IntArg<RETRACTION_TIME_ARG,DEFAULT_VALUE>,WavLen<EFFECT_RETRACTION>>;
 
+// Bend Options for Ignition/Retraction effects - defaults to BEND = 32768
```
profezzorn, line 58: Do we really need these aliases?

props/saber_fett263_buttons.h
fdarosa2663: replaced.

styles/edit_mode.h
fdarosa2663: They're meant to simplify the style code from the library where these are applied, not required but makes the code a little simpler when the feature is added.  If they're an issue I can remove.

profezzorn: Aliases should be used sparingly, as they add complication in the style editor.
Also, they can hide what's actually going on sometimes, making it harder for the end user.
My first impression is that it would be better to not have these aliases.

fdarosa2663: OK, I've removed.

PR #454, opened 2022-10-12
Add Clamp<>

fdarosa2663:


functions/clamp.h
```diff
@@ -0,0 +1,37 @@
+#ifndef FUNCTIONS_CLAMP_H
+#define FUNCTIONS_CLAMP_H
+
+#include "int.h"
+#include "svf.h"
+
+// Usage: Clamp<F, MIN, MAX>
```
profezzorn, line 7: Should be called ClampF

functions/clamp.h
```diff
@@ -0,0 +1,37 @@
...
+// Clamps value between MIN and MAX
+// F, A, B: INTEGER
+// return value: INTEGER
+
+template<class F, class MIN = Int<0>, class MAX = Int<32768>>
+class ClampXSVF {
```
profezzorn, line 13: Needs both SVF and non-SVF versions.
(The non-svf version should do the clamp inside getInteger())

functions/clamp.h
fdarosa2663: renamed

functions/clamp.h
fdarosa2663: Added, but the SVF, non-SVF still trips me up a bit so open to improvement.

profezzorn:
I added this code and the parts needed to make the SVF work. (hopefully, I didn't really test it.)

PR #455, opened 2022-10-15
Update HandleFeature

fdarosa2663:


blades/blade_base.h
```diff
@@ -95,6 +100,21 @@ class OneshotEffectDetector {
       case EFFECT_STAB:
 	BladeBase::HandleFeature(HANDLED_FEATURE_STAB);
 	break;
+      case EFFECT_INTERACTIVE_PREON:
```
profezzorn, line 103: Don't we need to add this effect first?

fdarosa2663: Ah, yes.  Added.  My local already had.

fdarosa2663:
Based on button mappings in my prop, I only need INTERACTIVE_PREON and FIRE to be HandledFeature, others can be controlled in prop now

PR #457, opened 2022-10-15
Use TransitionPulseL for TransitionEffectL

fdarosa2663:
switched TransitionEffectL to a using for TransitionPulseL

### DIFF
--- a/styles/transition_effect.h
+++ b/styles/transition_effect.h
@@ -2,6 +2,8 @@
 #define STYLES_TRANSITION_EFFECT_H
 
 #include "../transitions/concat.h"
+#include "../functions/effect_increment.h"
+#include "transition_pulse.h"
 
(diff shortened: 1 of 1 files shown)


styles/transition_effect.h
```diff
@@ -14,41 +16,7 @@
...
-    }
-    if (run_) {
-      transition_.run(blade);
-      if (transition_.done()) run_ = false;
-    }
-    last_detected_blade_effect = nullptr;
-    if (!run_) {
-      // All effects now wake the blade up again if needed, so it's safe to
-      // always return TRANSPARENT_UNTIL_IGNITION
-      return LayerRunResult::TRANSPARENT_UNTIL_IGNITION;
```
profezzorn, line 32: TransitionPulseL will need some optimizations before we can do this, otherwise blades that use TransitionEffectL will never turn off the power, resulting in poor standby time.

profezzorn:
Closing this (for now) as we are not ready to do this.

PR #458, opened 2022-10-16
TransitionEffectL using TransitionPulseL

fdarosa2663:
Added TRANSPARENT_UNTIL_IGNITION to TransitionPulseL

### DIFF
--- a/styles/transition_pulse.h
+++ b/styles/transition_pulse.h
@@ -25,6 +25,11 @@ public:
       if (transition_.done()) run_ = false;
     }
     last_detected_blade_effect = nullptr;
+    if (!run_) {
+      // All effects now wake the blade up again if needed, so it's safe to
+      // always return TRANSPARENT_UNTIL_IGNITION
+      return LayerRunResult::TRANSPARENT_UNTIL_IGNITION;
+    }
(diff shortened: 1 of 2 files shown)


styles/transition_pulse.h
```diff
@@ -25,6 +25,11 @@ class TransitionPulseL {
       if (transition_.done()) run_ = false;
     }
     last_detected_blade_effect = nullptr;
+    if (!run_) {
+      // All effects now wake the blade up again if needed, so it's safe to
+      // always return TRANSPARENT_UNTIL_IGNITION
+      return LayerRunResult::TRANSPARENT_UNTIL_IGNITION;
```
profezzorn, line 31: This is incorrect.
We can only return TRANSPARENT_UNTIL_IGNITION if the pulse returns ZERO_UNTIL_IGNITION.

fdarosa2663: I forgot about ZERO_UNTIL_IGNITION.  Thinking about it more, we should leave TransitionEffectL separate (and different) from TransitionPulseL.  While there's some crossover, there are other uses where TransitionPulseL needs a little more flexibility.

PR #459, opened 2022-10-19
BC_buttons. add EFFECT_QUOTE

NoSloppy:


NoSloppy:
@profezzorn  It's been brought to my attention this should be SaberBase::DoEffect(), not hybrid_font.
However, it's working fine as-is, and indeed a blade TransitionEffectL is triggered.
So...
1. Why is it working? ( for that matter, why do any calls to hybrid_font.DoEffect() work as that function doesn't even exist in hybrid_font anyway?)
2. Should I make this and any visual triggers to be SaberBase:: instead?

PR #460, opened 2022-10-19
Add AltF<>

fdarosa2663:


functions/alt.h
```diff
@@ -0,0 +1,14 @@
+#ifndef FUNCTIONS_ALTF_H
+#define FUNCTIONS_ALTF_H
+
+// Usage: AltF
+// return value: INTEGER
+// Returns current_alternative for use in ColorSelect<>, TrSelect<> or IntSelect<>
+
+class AltF {
```
profezzorn, line 8: This should be an SVF, see Int<> for an example.

fdarosa2663: Is that correct or do I need to do the SVFWrapper as well?

functions/alt.h
```diff
@@ -0,0 +1,18 @@
...
+#include "svf.h"
+
+// Usage: AltF
+// return value: INTEGER
+// Returns current_alternative for use in ColorSelect<>, TrSelect<> or IntSelect<>
+
+class AltSVF {
+ public:
+    void run(BladeBase* blade) {}
+    int getInteger(int led) { return current_alternative; }
```
profezzorn, line 13: Need both calculate() and getInteger() here.

functions/alt.h
```diff
@@ -0,0 +1,18 @@
...
+};
+
```
profezzorn, line 15: Need the "optimized specializations" from int.h (line 22/23) here to avoid SingleValueAdapter from creating an unnecessary variable.

functions/alt.h
fdarosa2663: gotcha, added, still not completely clear on all the SVF stuff

functions/alt.h
fdarosa2663: I think I did it right, it compiles.

PR #461, opened 2022-10-20
Add Interactive Blast and Preon

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -1573,6 +1573,30 @@ SaberFett263Buttons() : PropBase() {}
     }	  
   }
 
+  bool CheckInteractivePreon() {
+    bool includes_interactive_preon;
```
profezzorn, line 1577: First of all, variables like these should be initialized, otherwise they will have a random value, so like this:

   bool include_interactive_preon = false;

However, you don't need this variable. Instead of doing

    includes_interactive_preson |= CONDITION;

you can just do:
    if (CONDITION) return true;

and then end the function with:
   return false;

props/saber_fett263_buttons.h
```diff
@@ -1573,6 +1573,30 @@ SaberFett263Buttons() : PropBase() {}
     }	  
   }
 
+  bool CheckInteractivePreon() {
```
profezzorn, line 1576: I think this function (and the one below) may be generically useful, so you should put them in prop_base.h. They will not take any space if not called.

props/saber_fett263_buttons.h
```diff
@@ -4725,8 +4753,12 @@ SaberFett263Buttons() : PropBase() {}
           MenuChoice();
           return true;
         }
-        ToggleBattleModeMultiBlast();
-        SaberBase::DoBlast();
+        if (CheckInteractiveBlast()) {
+          SaberBase::DoEffectR(EFFECT_FIRE);
```
profezzorn, line 4757: I'm confused, shouldn't this be SaberBase::DoEffectR(EFFECT_INTERACTIVE_BLAST);
?

props/saber_fett263_buttons.h
```diff
@@ -4735,14 +4767,22 @@ SaberFett263Buttons() : PropBase() {}
           StopTrackPlayer();
           return true;
         }
-        ToggleBattleModeMultiBlast();
-        SaberBase::DoBlast();
+        if (CheckInteractiveBlast()) {
```
profezzorn, line 4770: This chunk seems to be repeated in a few places. I suggest a helper function

props/saber_fett263_buttons.h
```diff
@@ -5832,6 +5897,7 @@ SaberFett263Buttons() : PropBase() {}
   bool auto_lockup_on_ = false; // Battle Mode Lockup active
   bool auto_melt_on_ = false; // Battle Mode Melt/Drag active
   bool battle_mode_ = false; // Battle Mode active
+  bool preon_effect_ = false; // used for interactive preon
```
profezzorn, line 5900: This variable (and it's comment) are confusing.
What is it for?
Why set it in SB_Effect()?
When is it reset back to false?

props/saber_fett263_buttons.h
fdarosa2663: Ah, thanks.  OK simplified both of these functions and moved to prop_base

props/saber_fett263_buttons.h
fdarosa2663: moved

props/saber_fett263_buttons.h
fdarosa2663: Originally when I first built we were using EFFECT_FIRE (borrowed from blaster) but I've added INTERACTIVE_BLAST and updated

props/saber_fett263_buttons.h
fdarosa2663: added

props/saber_fett263_buttons.h
fdarosa2663: comment updated, for Interactive Preon to work properly EFFECT_INTERACTIVE_PREON sets to true, then the conditional effect for EFFECT_FAST_ON within the style resets to false.  This is to allow the "interactive" part as much time as the user wants based on the style code.  Calling EFFECT_FAST_ON from within the Interactive Preon is how we ignite the blade based on the condition set within the effect.

fdarosa2663:
Just pinging this one, I have a lot of small PRs lined up for my prop, I'll do one-at-a-time to keep them small.

props/prop_base.h
```diff
@@ -701,6 +701,20 @@ class PropBase : CommandParser, Looper, protected SaberBase {
     FindBlade();
   }
 
+  bool CheckInteractivePreon() {
+    #define USES_INTERACTIVE_PREON(N) \
+    if (current_config->blade##N->current_style() && current_config->blade##N->current_style()->IsHandled(HANDLED_FEATURE_INTERACTIVE_PREON)) return true;
+    return false;
```
profezzorn, line 707: return false must be after ONCEPERBLADE
Did you test this at all?

props/saber_fett263_buttons.h
```diff
@@ -4533,7 +4542,11 @@ SaberFett263Buttons() : PropBase() {}
             return true;
 #endif
           } else {
-            DoIgnition();
+            if (!CheckInteractivePreon()) {
```
profezzorn, line 4545: switch if and else branches around and remove the "!"

props/saber_fett263_buttons.h
```diff
@@ -4889,7 +4899,11 @@ SaberFett263Buttons() : PropBase() {}
           return true;
 #endif
         } else {
-          DoIgnition();
+          if (!CheckInteractivePreon()) {
```
profezzorn, line 4902: repeated, create helper function?

common/saber_base.h
```diff
@@ -46,6 +46,7 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(OFF_CLASH)                    \
     DEFINE_EFFECT(NEXT_QUOTE)                   \
     DEFINE_EFFECT(INTERACTIVE_PREON)            \
+    DEFINE_EFFECT(INTERACTIVE_BLAST)            \
```
profezzorn, line 49: Since this is obviously different from EFFECT_FIRE in some way, please add a comment that explains what that difference is.

props/prop_base.h
fdarosa2663: Fixed, my local version had it reversed when it was in my prop, I had to break up pieces to keep PRs small but it's getting harder to keep track of differences in the files since these have been sitting a while.

common/saber_base.h
fdarosa2663: added

props/saber_fett263_buttons.h
fdarosa2663: helper function added

props/saber_fett263_buttons.h
fdarosa2663: moved to helper function and reversed

PR #465, opened 2022-10-21
default AUDIO_CLASH_SUPPRESSION to 1

fdarosa2663:
Already appears to be confusion on this define and we're not even in Alpha so better to default to 1 at this time and leave it to users to add.  Users were testing without define and had harder time clashing because it defaulted to 10.


profezzorn:
First of all, I'm fairly certain people will need to re-jigger their clash thresholds anyways because of changes in the clash detection code. If this code is helpful to people, then we should keep it.

If we are going to disable it by default, we should do it by leaving the define undefined and adding some ifdefs to the code to strip out the code completely. It's only a few bytes I guess, but still, why waste those bytes?

fdarosa2663:
OK, I'm fine with leaving at 10, it worked as expected for me or we can do ifdefs, whichever you feel is better.  Feel free to close this PR I'll leave open for now just so it stays in front of us.

PR #467, opened 2022-10-24
Add TransitionSequenceWithResetL

fdarosa2663:
For use in Interactive effects and mini games


fdarosa2663:
Looking back, I think this was replaced trying to catch up all versions, closing for now.  If alternative doesn't work will resubmit.

PR #469, opened 2022-10-28
ModF<>

fdarosa2663:


profezzorn:
We need a non-svf version as well.

fdarosa2663:
I "think" that's what you meant.

functions/mod.h
```diff
@@ -3,14 +3,31 @@
...
+  }
+private:
+  PONUA F f_;
+  PONUA MAX max_;
+};
+
 // Usage: ModF<F, MAX>
 // F: FUNCTION
-// MAX: FUNCTION (not zero) defaults to Int<32768>
+// MAX: FUNCTION (not zero)
```
profezzorn, line 25: "not zero": Good point, but we should probably check for that just like we do in Divide

PR #470, opened 2022-10-28
Add EFFECT_VOLUME_LEVEL

NoSloppy:


functions/volume_level.h
```diff
@@ -0,0 +1,20 @@
...
+
+// Usage: VolumeLevel
+// Returns 0-32768 based on volume level.
+// returned value: INTEGER
+
+class VolumeLevelSVF {
+ public:
+  void run(BladeBase* blade) {}
+  int calculate(BladeBase* blade) {
+    return clampi32((dynamic_mixer.get_volume() * 100 / VOLUME) * 32768 / 100, 0, 32768);
```
profezzorn, line 14: no point in multiplying by 100 and dividing by it later.
Also, I don't think we need the clamp, I think this is enough:

   return dynamic_mixder.get_volume() * 32768 / VOLUME;

PR #471, opened 2022-10-30
Fix current_alternative = num_alternatives

fdarosa2663:
If a style returned current_alternative >= num_alternatives no alt sounds found or played

### DIFF
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -527,13 +527,14 @@ public:
 	if (num_alternatives) {
 	  if (SaberBase::sound_number == -1) {
 	    // Next alternative
-	    if (++current_alternative >= num_alternatives)  current_alternative = 0;
+	    ++current_alternative;
 	  } else {
 	    // Select a specific alternative.
 	    current_alternative = std::min<int>(SaberBase::sound_number, num_alternatives);
 	    // Set the sound num to -1 so that the altchng sound is random.
 	    SaberBase::sound_number = -1;
 	  }
+          if (current_alternative >= num_alternatives) current_alternative = 0;
 	  RestartHum();
 	}
 	PlayCommon(&SFX_altchng);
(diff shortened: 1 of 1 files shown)


sound/hybrid_font.h
```diff
@@ -527,13 +527,14 @@ class HybridFont : public SaberBase, public Looper {
...
 	    // Next alternative
-	    if (++current_alternative >= num_alternatives)  current_alternative = 0;
+	    ++current_alternative;
 	  } else {
 	    // Select a specific alternative.
 	    current_alternative = std::min<int>(SaberBase::sound_number, num_alternatives);
 	    // Set the sound num to -1 so that the altchng sound is random.
 	    SaberBase::sound_number = -1;
 	  }
+          if (current_alternative >= num_alternatives) current_alternative = 0;
```
profezzorn, line 537: I thought you didn't want wrap-around for alternatives?

sound/hybrid_font.h
```diff
@@ -527,13 +527,14 @@ class HybridFont : public SaberBase, public Looper {
 	if (num_alternatives) {
 	  if (SaberBase::sound_number == -1) {
 	    // Next alternative
-	    if (++current_alternative >= num_alternatives)  current_alternative = 0;
+	    ++current_alternative;
 	  } else {
 	    // Select a specific alternative.
 	    current_alternative = std::min<int>(SaberBase::sound_number, num_alternatives);
```
profezzorn, line 533: If you don't want wrap-around, then the correct fix is to change this to num_alternatives - 1
(I think, I don't remember how I lined this variables up actually...)

sound/hybrid_font.h
fdarosa2663: need to think this through, need to figure out best approach so the style can't set an alt that doesn't exist because all sounds in the alt stop when that happens.  will try out a few things tomorrow and update

PR #473, opened 2022-10-31
ModF<>

fdarosa2663:


functions/mod.h
```diff
@@ -0,0 +1,59 @@
...
+
+template<class F, class MAX>
+class ModBase {
+public:
+  void run(BladeBase* blade) {
+    f_.run(blade);
+    max_.run(blade);
+  }
+  int getInteger(int led) {
+    if (max_.getInteger(led) == 0) {
```
profezzorn, line 16: That is indeed what I meant, but we shouldn't call max_.getInteger twice.
We should stick the return value in a variable and use that twice instead.
Same goes for max_.calculate below.

fdarosa2663: should've know that, fixed

fdarosa2663: just pinging, the MOD fix in stripes corrected usage

PR #476, opened 2022-11-03
sub_files_ error in font direct & NO_REPEAT_RANDOM

fdarosa2663:
There was an "error in font directory" caused by adding 1 to sub_files_ and added NO_REPEAT_RANDOM for sub_files_ returned.

### DIFF
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -213,7 +213,7 @@ class Effect {
 	isDigit(rest[2]) &&
 	isDigit(rest[3])) {
       int sub = strtol(rest+1, nullptr, 10);
-      sub_files_ = std::max<int>(sub_files_, sub);
+      sub_files_ = std::max<int>(sub_files_, sub + 1);
       rest += 4;
     }
 
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -359,8 +359,10 @@ class Effect {
     if (!sub_files_) return 0;
 #ifdef NO_REPEAT_RANDOM
     int sel = rand() % sub_files_;
-    for (int i = 0; i < 3 && sel == last_sub_; i++) {
-      sel = clamp(sel + 1 - (rand() & 2), 0, sub_files_ - 1);
+    if (sub_files_ > 2 && selected_ == last_) {
+      for (int i = 0; i < 3 && sel == last_sub_; i++) {
+        sel = rand() % sub_files_;
+      }
     }
     last_sub_ = sel;
     return sel;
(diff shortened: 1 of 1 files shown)


sound/effect.h
```diff
@@ -213,7 +213,7 @@ class Effect {
 	isDigit(rest[2]) &&
 	isDigit(rest[3])) {
       int sub = strtol(rest+1, nullptr, 10);
-      sub_files_ = std::max<int>(sub_files_, sub + 1);
+      sub_files_ = std::max<int>(sub_files_, sub);
```
profezzorn, line 216: This is wrong.
Sub files now start at 000
so if you have 000, 001 and 002 you have 3 files.
002 + 1 == 3

sound/effect.h
```diff
@@ -352,11 +352,21 @@ class Effect {
 
 #ifdef NO_REPEAT_RANDOM
   int last_;
+  int last_sub_;
```
profezzorn, line 355: we should make last_ and last_sub_  int16_t

sound/effect.h
```diff
@@ -352,11 +352,21 @@ class Effect {
...
 #endif
 
   int random_subid() {
     if (!sub_files_) return 0;
+#ifdef NO_REPEAT_RANDOM
+    int sel = rand() % sub_files_;
+    for (int i = 0; i < 3 && sel == last_sub_; i++) {
+      sel = clamp(sel + 1 - (rand() & 2), 0, sub_files_ - 1);
```
profezzorn, line 363: we don't need this complicated logic for sub-files sine there is no need to try to select select a "nearby" value.
Instead we should just do sel = rand() % sub_files_; again.

profezzorn: also, we may need special cases if sub_files_ is two.
And, we may want to avoid doing this if last_ is different from whatever file is getting selected this time.

sound/effect.h
fdarosa2663: ok, so I needed to change in my font then, I thought it was only the alt that changed.  will test again.

fdarosa2663: maybe sub_files_ shouldn't have 0 index?  unable to to Select(0) now, send PM in theCrucible with more detail.

sound/effect.h
fdarosa2663: updated

sound/effect.h
fdarosa2663: updated and accounting for >2 and selected_ = last_ now

profezzorn:
Implemented this myself.

PR #478, opened 2022-11-17
Add DISABLE_QUOTE_PLAYER

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -4400,6 +4408,7 @@ SaberFett263Buttons() : PropBase() {}
   }
 #endif
 
+#ifndef FETT263_DISABLE_QUOTE_PLAYER
```
profezzorn, line 4411: You can simplify this a lot by doing:

void CHeckQuote() {
#ifndef FETT263_DISABLE_QUOTE_PLAYER
    // quote player code
#else
    SaberBase::DoEffect(EFFECT_FORCE, 0);
#endif
}

Then leave all the places you call CheckQuote alone.

fdarosa2663: thanks, updated

PR #480, opened 2022-11-20
New effects

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -5835,6 +5835,16 @@ SaberFett263Buttons() : PropBase() {}
     switch (effect) {
       case EFFECT_INTERACTIVE_BLAST: hybrid_font.PlayCommon(&SFX_blast); return;
       case EFFECT_QUOTE: hybrid_font.PlayCommon(&SFX_quote); return;
+      case EFFECT_NEXT_QUOTE:
+        SFX_quote.SelectNext();
```
profezzorn, line 5839: This is a bit unusual.
In most other places, we would use the SFX_quote.SelectNext(); and then use EFFECT_QUOTE.
What makes this different?

fdarosa2663: This effect will be called from within the style code so it needs to be handled via SB_EFFECT separately from EFFECT_QUOTE

PR #482, opened 2022-11-21
More New Effects

fdarosa2663:


common/saber_base.h
```diff
@@ -57,6 +57,8 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(ALT_SOUND)			\
     /* Triggers an optional sound effect during transitions from within a style via TrDoEffect. */         \
     DEFINE_EFFECT(TRANSITION_SOUND)		\
+    /* Toggles an optonal sound effect loop ON/OFF from within a style via TrDoEffect. */          \
+    DEFINE_EFFECT(SOUND_LOOP)                   \
```
profezzorn, line 61: Wouldn't it be better if you used a value of -2 to stop the sound?
Like:
Play a specific sound: TrDoEffect<SOUND_LOOP, 1>
Play any sound: TrDoEffect<SOUND_LOOP, -1>
Stop playing: TrDoEffect<SOUND_LOOP, -2>
?

props/saber_fett263_buttons.h
```diff
@@ -1927,6 +1929,21 @@ SaberFett263Buttons() : PropBase() {}
     }
   }
 
+void SoundLoop() {
+    if (wav_player->isPlaying()) {
+      wav_player->Stop();
```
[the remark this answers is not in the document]
fdarosa2663: thanks, added

common/saber_base.h
fdarosa2663: The intended use is a "toggle" from the same control so if you do the effect it will turn itself on the first time and off the second, etc.  Separating on/off into two controls will use up a control "slot" that could be used for something else.  It would also potentially make the style more complex since you'd need separate DoEffect layers.  Specific and random are already supported.

props/saber_fett263_buttons.h
```diff
@@ -1927,6 +1929,21 @@ SaberFett263Buttons() : PropBase() {}
...
+      wav_player.Free();
+      STDOUT.println("End Wav Player");    
+    } else {
+      wav_player = GetFreeWavPlayer();
+      if (wav_player) {
+           wav_player->PlayOnce(&SFX_trloop);
```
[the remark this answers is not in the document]

common/saber_base.h
profezzorn: Toggles can get out of sync though.
Like if you wanted to have a display or blade effect to go with your sound loop.

fdarosa2663: Not if the DoEffect is triggered by the ColorSelect controlling the visual effect itself, which is the intended use.  There's probably pros and cons to both approaches but I think the cons of needing a -2 in DoEffect in the style outweigh the pros of not needing to take up another control for most of the intended uses I have.  For all of the style based sounds the user already needs to plan out the visual effects to coordinate with the sounds, so the "toggle" fits for how I've been building and testing thus far.  I think it would actually be more complicated trying to synch up DoEffect with a ColorSelect if you needed to fit -2 into it the selection method as it would then need a separate layer, a separate transition or a more complex function to be able to handle.

PR #483, opened 2022-11-28
Add LockupPulseF

fdarosa2663:


functions/effect_increment.h
```diff
@@ -21,6 +21,25 @@ class EffectPulseSVF {
 template<BladeEffectType EFFECT>
 using EffectPulseF = SingleValueAdapter<EffectPulseSVF<EFFECT>>;
 
+// Usage: LockupPulseF<LOCKUP_TYPE>
+// LOCKUP_TYPE: a SaberBase::LockupType
+// Returns 32768 once for each time the given lockup occurs.
+template<SaberBase::LockupType LOCKUP_TYPE>
+class LockupPulseSVF {
+public:
+  void run(BladeBase* blade) {}
```
profezzorn, line 30: I think we should register the LOCKUP_TYPE as handled when someone uses one of these.

fdarosa2663: Not sure how to do, can you point me to example?

PR #484, opened 2022-11-28
Add plug-in instructions for convenience

NoSloppy:
Useful?


profezzorn:
Not very accurate since you can also compile ProffieOS for teensys. (And maybe soon for esp32)

PR #485, opened 2022-12-02
Add Caiwyn's Lightsaber Controls

Caiwyn:
props/saber_caiwyn_buttons.h implements Caiwyn's lightsaber controls, intended to provide fast, low-latency response to core lightsaber effects while omitting, limiting, or otherwise compromising access to non-critical effects or features.  Instructions on how to use these controls are contained in the comments at the top of the file.


Caiwyn:
Discovered I could remove some unnecessary cruft; will create a new pull request.

PR #486, opened 2022-12-02
Add Caiwyn's Lightsaber Controls

Caiwyn:
props/saber_caiwyn_buttons.h implements Caiwyn's lightsaber controls, intended to provide fast, low-latency response to core lightsaber effects while omitting, limiting, or otherwise compromising access to non-critical effects or features. Instructions on how to use these controls are contained in the comments at the top of the file.

### DIFF
--- a/props/saber_caiwyn_buttons.h
+++ b/props/saber_caiwyn_buttons.h
@@ -32,6 +32,7 @@
 //                    Lockup: Hold Aux during impact
 //                      Drag: Hold Aux during impact with saber pointed down.
 //                      Melt: Hold Aux and stab
+//           Lightning Block: Hold Aux2 and Press Aux (3 buttons required)
 //
 //         Enter Volume Menu: Double-click and hold Aux for 1 second
 //                            | Be aware that the first click will trigger a
@@ -577,6 +578,14 @@ public:
         break;
 #endif
 
+// Lightning Block
+#if NUM_BUTTONS > 2
+      case EVENTID(BUTTON_AUX, EVENT_PRESSED, MODE_ON | BUTTON_AUX2):
+        SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
+        SaberBase::DoBeginLockup();
+        return true;
+#endif
+
 #ifdef BLADE_DETECT_PIN
       case EVENTID(BUTTON_BLADE_DETECT, EVENT_LATCH_ON, MODE_ANY_BUTTON | MODE_ON):
       case EVENTID(BUTTON_BLADE_DETECT, EVENT_LATCH_ON, MODE_ANY_BUTTON | MODE_OFF):
@@ -598,9 +607,6 @@ public:
       // Events that need to be handled regardless of what other buttons are pressed.
       case EVENTID(BUTTON_POWER, EVENT_RELEASED, MODE_ANY_BUTTON | MODE_ON):
       case EVENTID(BUTTON_AUX, EVENT_RELEASED, MODE_ANY_BUTTON | MODE_ON):
-#if NUM_BUTTONS > 2
-      case EVENTID(BUTTON_AUX2, EVENT_RELEASED, MODE_ANY_BUTTON | MODE_ON):
-#endif
         if (SaberBase::Lockup()) {
           SaberBase::DoEndLockup();
           SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
--- a/props/saber_caiwyn_buttons.h
+++ b/props/saber_caiwyn_buttons.h
@@ -1,60 +1,93 @@
 // Caiwyn's Lightsaber Controls
-// This config is designed to provide fast, low-latency response to button presses for the most basic functions (blaster blocks, lockups, and clashes) by omitting, or limiting non-critical features.
-// At least 2 buttons are required.  A third button can be added, but is not necessary.
+// This config is designed to provide fast, low-latency response to button
+// presses for the most basic functions (blaster blocks, lockups, and clashes)
+// by omitting, or limiting non-critical features.
 //
-// While saber is DEACTIVATED:
-// Start/Stop Track: Click Aux
-// Track Player Mode: Double-click Aux
-//   This cycles through three playback modes:
-//   1. Play a single track and stop
-//   2. Repeat a single track in a loop
-//   3. Repeat all tracks in a loop
-//   Tracks must be in <font>/tracks/*.wav, and will be selected in alphabetical order.
-// Next Track: Hold Aux while track is playing
-// Next Preset: Hold Aux when track is not playing
-// Check Battery: Double-click and hold Aux
-// Activate Saber: Press Power
-// Activate Saber and Start Track: Hold Aux or Aux2 and press Power
+// At least 2 buttons are required.  A third button can be added, but is not
+// necessary.
 //
-// While saber is ACTIVATED:
-// Non-impact Clash/Lockup: Click Power (This generates a clash/lockup effect with no impact to the blade; quick press for a short clash, hold for a lockup)
-// Blaster Block: Click Aux
-// Lightning Block: Hold Aux
-// Lockup: Hold Aux during impact
-// Drag: Hold Aux during impact with saber pointed down.
-// Melt: Hold Aux and stab
-// Enter Volume Menu: Double-click and hold Aux (Be aware that the first click will trigger a blaster block)
-//   Increase Volume: Rotate hilt right while in Volume Menu
-//   Decrease Volume: Rotate hilt left while in Volume Menu
-//   Save Changes and Exit Volume Menu: Click Power
-//   Discard Changes and Exit Volume Menu: Click Aux
-// Enter Color Change Mode (Color Wheel): Triple-click and hold Aux (Be aware that the first two clicks will trigger blaster blocks)
-//   Change Color: Rotate hilt while in Color Change Mode
-//   Color Zoom: Press & Hold Power and Rotate hilt while in Color Change Mode
-//   Save Changes and Exit Color Change Mode: Release Power
-//   Reset Color and Exit Color Change mode: Click Aux
-// Deactivate Saber: Hold Aux or Aux2 and press Power
+// While saber is OFF:
+//          Start/Stop Track: Click Aux
+//         Track Player Mode: Double-click Aux
+//                            * This cycles through three playback modes:
+//                            * 1. Play a single track and stop
+//                            * 2. Repeat a single track in a loop
+//                            * 3. Repeat all tracks in a loop
+//                            * Tracks must be in <font>/tracks/*.wav, and will
+//                            * be selected in alphabetical order.
+//                Next Track: Hold Aux for 1 second while track is playing
+//               Next Preset: Hold Aux for 1 second when track is not playing
+//             Check Battery: Double-click and hold Aux for 1 second
+//                   Turn On: Press Power
+//     Turn On & Start Track: Hold Aux or Aux2 and press Power
+//
+// While saber is ON:
+//   Non-impact Clash/Lockup: Click/Hold Power
+//                            * This generates a clash/lockup effect with no
+//                            * impact to the blade; quick press for a short
+//                            * clash, hold for a lockup
+//             Blaster Block: Click Aux
+//           Lightning Block: Hold Aux
+//                    Lockup: Hold Aux during impact
+//                      Drag: Hold Aux during impact with saber pointed down.
+//                      Melt: Hold Aux and stab
+//
+//         Enter Volume Menu: Double-click and hold Aux for 1 second
+//                            * Be aware that the first click will trigger a
+//                            * blaster block
+//           Increase Volume: Rotate hilt right while in Volume Menu
+//           Decrease Volume: Rotate hilt left while in Volume Menu
+//   Save & Exit Volume Menu: Click Power
+// Cancel & Exit Volume Menu: Click Aux
+//
+//   Enter Color Change Mode: Triple-click and hold Aux for 1 second
+//                            * Be aware that the first two clicks will
+//                            * trigger blaster blocks
+//              Change Color: Rotate hilt while in Color Change Mode
+//                Color Zoom: Press & Hold Power and Rotate hilt
+//                            * This will fine-tune the selected color before
+//                            * saving your change
+//  Save & Exit Color Change: Release Power while in Color Zoom
+// Reset & Exit Color Change: Click Aux
+//
+//                  Turn Off: Hold Aux or Aux2 and press Power
 //
 // You will need the following sound files in order for menus to work properly:
 // vmbegin.wav - Enter Volume Change Menu
 // vmend01.wav - Save Volume Change
 // vmend02.wav - Cancel Volume Change
-// monce.wav - Set Track Player to play a single track one time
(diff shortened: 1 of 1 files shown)


props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
...
+//   Save Changes and Exit Volume Menu: Click Power
+//   Discard Changes and Exit Volume Menu: Click Aux
+// Enter Color Change Mode (Color Wheel): Triple-click and hold Aux (Be aware that the first two clicks will trigger blaster blocks)
+//   Change Color: Rotate hilt while in Color Change Mode
+//   Color Zoom: Press & Hold Power and Rotate hilt while in Color Change Mode
+//   Save Changes and Exit Color Change Mode: Release Power
+//   Reset Color and Exit Color Change mode: Click Aux
+// Deactivate Saber: Hold Aux or Aux2 and press Power
+//
+// You will need the following sound files in order for menus to work properly:
```
profezzorn, line 38: Are all of these sounds available in the voice packs?
If so, maybe add an url to where they can be found...

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
...
+// vmend02.wav - Cancel Volume Change
+// monce.wav - Set Track Player to play a single track one time
+// mloop.wav - Set Track Player to repeat a single track
+// mrotate.wav - Set Track Player to repeat all tracks
+// ccbegin.wav - Enter Color Change Mode
+// ccend01.wav - Save Color and Exit Color Change Mode
+// ccend02.wav - Reset Color and Exit Color Change Mode
+//
+// Options you can add to your config file:
+// #define DISABLE_BEEPS        - Disables the single beep that occurs when selecting fonts, tracks, or entering or exiting the volume and color change menus.
```
profezzorn, line 50: All defines must have a prefix unique to your config file, something like:

CAIWYN_DISABLE_BEEPS

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
...
+
+#include "prop_base.h"
+#include "../sound/hybrid_font.h"
+#include "../sound/sound_library.h"
+
+#undef PROP_TYPE
+#define PROP_TYPE CaiwynButtons
+
+#ifndef NUM_BUTTONS
+#define NUM_BUTTONS 2
```
profezzorn, line 70: This is a weird thing to do in a prop file, I suggest an #error instead.

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
...
+#ifndef BUTTON_HELD_MEDIUM_TIMEOUT
+#define BUTTON_HELD_MEDIUM_TIMEOUT 1000
+#endif
+
+#ifndef BUTTON_HELD_LONG_TIMEOUT
+#define BUTTON_HELD_LONG_TIMEOUT 2000
+#endif
+
+EFFECT(vmbegin);
+EFFECT(vmend);
```
profezzorn, line 98: vmbegin and vmend are available in the sound library, why not just use SayEnterVolumeMenu() / SayVolumeMenuEnd() ?

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
...
+
+// Fett263 Track Player, modified by Caiwyn
+  enum TrackMode {
+    PLAYBACK_ONCE,
+    PLAYBACK_LOOP,
+    PLAYBACK_ROTATE,
+    PLAYBACK_SHUFFLE,
+  };
+
+  void TrackPlayer() {
```
profezzorn, line 179: TrackPlayerLoop might be a better name

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
...
+      }
+      if (a < -M_PI / 6) {
+        VolumeDown();
+        current_menu_angle_ = fusor.angle2();
+      }
+    }
+  }
+
+// Fett263 Track Player, modified by Caiwyn
+  enum TrackMode {
```
profezzorn, line 172: The instructions at the top say there are three modes, but this shows four?

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
...
+          talkie.Say(spPERCENT);
+#endif
+          SaberBase::DoEffect(EFFECT_BATTERY_LEVEL, 0);
+          return true;
+        }
+        break;
+
+// Track Playback Mode
+      case EVENTID(BUTTON_AUX, EVENT_SECOND_CLICK_SHORT, MODE_OFF):
+        switch (track_mode_) {
```
profezzorn, line 339: Probably better to break this out into a helper function.

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
...
+          }
+          ToggleColorChangeMode();
+          return true;
+        }
+        break;
+#endif
+
+// Cancel Volume / Reset Color / Blaster Block
+      case EVENTID(BUTTON_AUX, EVENT_CLICK_SHORT, MODE_ON):
+        if (mode_volume_) {
```
profezzorn, line 463: In general, all of these large blocks of code should not go in this function, but in helper functions.

profezzorn (commented):
Not sure I got everything, we'll see what I find once these comments have been addressed.

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
+// Caiwyn's Lightsaber Controls
+// This config is designed to provide fast, low-latency response to button presses for the most basic functions (blaster blocks, lockups, and clashes) by omitting, or limiting non-critical features.
+// At least 2 buttons are required.  A third button can be added, but is not necessary.
+//
+// While saber is DEACTIVATED:
```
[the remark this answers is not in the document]
Caiwyn: "Activated/Deactivated" seemed more precise to me, but I can change it to "On/Off," sure.

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
+// Caiwyn's Lightsaber Controls
+// This config is designed to provide fast, low-latency response to button presses for the most basic functions (blaster blocks, lockups, and clashes) by omitting, or limiting non-critical features.
```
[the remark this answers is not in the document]
Caiwyn: Will do.

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,604 @@
...
+//
+// While saber is DEACTIVATED:
+// Start/Stop Track: Click Aux
+// Track Player Mode: Double-click Aux
+//   This cycles through three playback modes:
+//   1. Play a single track and stop
+//   2. Repeat a single track in a loop
+//   3. Repeat all tracks in a loop
+//   Tracks must be in <font>/tracks/*.wav, and will be selected in alphabetical order.
+// Next Track: Hold Aux while track is playing
```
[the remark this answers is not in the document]
Caiwyn: 1 second.  All "Hold" controls that aren't lockup-related are 1 second.  I left that information out of the comments for the sake of brevity.

props/saber_caiwyn_buttons.h
Caiwyn: Most of them are; I used standard sounds that were used by Fett263's button config file whenever possible.  I have specified separate vmend and ccend files that are Selected depending on whether the user saves or cancels a volume or color change, but if you only have one, the code will use the same file for both.  The only sound that isn't already available is monce.wav.  The rest are all standard.  I'm putting together my own voice pack that will include the extra voice prompts I use, but it's not quite finished.

props/saber_caiwyn_buttons.h
Caiwyn: Will do.

props/saber_caiwyn_buttons.h
Caiwyn: I just used the same name that was in saber_fett263_buttons.h

props/saber_caiwyn_buttons.h
Caiwyn: Yes, that's planned future functionality.  I intend to add a shuffle mode eventually, but I haven't figured out how to do it just yet.

props/saber_caiwyn_buttons.h
Caiwyn: I can do that.

props/saber_caiwyn_buttons.h
Caiwyn: That was just a stylistic decision.  I didn't put anything into a helper function that didn't need to be repeated.

props/saber_caiwyn_buttons.h
Caiwyn: The Say_____() functions invoke the wav_player buffer, which is used to string multiple sounds together.  The only time I use that is for the battery level check.  I used Effect() for these because I thought that was the more standard way of doing it, and because I do have monce.wav that isn't included in sound_library.  I can switch over to using the sound_library functions, though... I just have to figure out how to extend them with monce.wav

props/saber_caiwyn_buttons.h
Caiwyn: I could've sworn I copied that from an existing prop file but I'm not seeing it anywhere else now.  Probably should just remove it.

props/saber_caiwyn_buttons.h
Caiwyn: Oh, now I remember.  I did it this way because it allows me to test for whether the sound files exist by doing "if (SFX_vmend)" and if the files don't exist I play a series of beeps instead.  I didn't see a way to do that in sound_library.  Is there one?

props/saber_caiwyn_buttons.h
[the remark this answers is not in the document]
Caiwyn: Done, wrapped to 79 characters.

props/saber_caiwyn_buttons.h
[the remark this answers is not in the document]
Caiwyn: Done.

props/saber_caiwyn_buttons.h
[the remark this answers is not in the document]
Caiwyn: Done.

props/saber_caiwyn_buttons.h
Caiwyn: Done.

props/saber_caiwyn_buttons.h
Caiwyn: Done, this is removed.

props/saber_caiwyn_buttons.h
Caiwyn: Removed this for now.

props/saber_caiwyn_buttons.h
Caiwyn: Done.

props/saber_caiwyn_buttons.h
Caiwyn: Done.

props/saber_caiwyn_buttons.h
Caiwyn: Leaving this as is for now. I prefer to keep the same nomenclature as in saber_fett263_buttons.h but if there’s an important reason I should change it, let me know.

props/saber_caiwyn_buttons.h
Caiwyn: Leaving this as is for now. I’m not opposed to using sound_library.Say***() to trigger sounds instead of hybrid_font.PlayPolyphonic(), but I would still need EFFECT() to create the SFX_*** variables that I test for the light the code, or some other alternative to test whether the wav files exist.

props/saber_caiwyn_buttons.h
Caiwyn: I am still working on the mnum*.wav files. The config works without them; and any missing file is replaced by a beep code or a talkie.Say(). The standard edit mode menu wav files will work in a pinch.

profezzorn: Don't explain it to me though.. It would be much better if you explain it in the file so that everybody who reads the file will know which files are expected and where to find the files needed.

props/saber_caiwyn_buttons.h
profezzorn: I think the easy thing to do is just to tell people to install a voice pack, then you can basically assume that all of those files exists. No real need to test each file individually. Might be good to have an EFFECT for one of them, just to make sure that people didn't miss installing the voice pack though.

props/saber_caiwyn_buttons.h
profezzorn: Not particularly important, it just sounds more like a class name than a function name to me.

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,639 @@
...
+        }
+        return true;
+
+// Battery Level
+      case EVENTID(BUTTON_AUX, EVENT_SECOND_HELD_MEDIUM, MODE_OFF):
+        if (track_player_on_) {
+          ChangeTrackMode();
+          return true;
+        } else if (!wav_player_->isPlaying()) {
+          SaberBase::DoEffect(EFFECT_BATTERY_LEVEL, 0);
```
profezzorn, line 455: Maybe another helper function?
(I find that the Event2 function gets easier to read if it's mostly just cases and function calls.)

props/saber_caiwyn_buttons.h
Caiwyn: Finished the voice pack and made it publicly available.  There is now a link in the instructions.

props/saber_caiwyn_buttons.h
Caiwyn: Easy, yes, but I like having beep codes as a fallback in case sound files are missing or in the wrong format.  That's how the Color Wheel menu works already, and I modeled the behavior of my volume and track playback menus off that.  Is there some disadvantage to using EFFECT() to set up the sounds?  I notice this is how saber_sa22c_buttons.h sets up vmbegin and vmend, along with other sounds.

If there's another way to check to see if a file exists, I would be interesting in implementing it.

I did remove mbatt and mpercent from the list of EFFECT() definitions; I can just test for mnum, which is already defined in sound_library.

props/saber_caiwyn_buttons.h
```diff
@@ -0,0 +1,639 @@
...
+          if (num_tracks_ <= 0) {
+            StartOrStopTrack();
+          } else {
+            PlayTrack();
+          }
+          break;
+        case PLAYBACK_ROTATE:
+          NextTrack();
+          break;
+        default:
```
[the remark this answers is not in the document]
Caiwyn: The compiler will throw an error if all values aren't accounted for in the switch?  I didn't know that.  The only reason the default was there was to handle that exact scenario.  I've changed it as you suggested.

props/saber_caiwyn_buttons.h
Caiwyn: Done.

props/saber_caiwyn_buttons.h
profezzorn: EFFECTS take a little more memory, but almost never enough to matter.
If you want the fallbacks, I'm ok with that.

PR #488, opened 2022-12-21
Add FETT263_BC_SAY_BATTERY_VOLTS_PERCENT

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -4473,14 +4485,24 @@ SaberFett263Buttons() : PropBase() {}
 
   // Battery Level
   void DoBattery() {
-#if defined (FETT263_SAY_BATTERY_PERCENT)
-    sound_library_.SayBatteryLevel();
-    sound_library_.SayNumber(battery_monitor.battery_percent(), SAY_WHOLE);
-    sound_library_.SayPercent();
-#elif defined(FETT263_SAY_BATTERY_VOLTS)
+#if defined (FETT263_SAY_BATTERY_PERCENT) && defined (FETT263_SAY_BATTERY_VOLTS)
```
profezzorn, line 4488: This is pretty messy.
In this case, the repeated "bad" part is actually the #if line itself.

One potential way to clean it up a bit would be something like:

void SayBattery() {
#ifdef FETT263_SAY_BATTERY_VOLTS
#ifdef FETT253_SAY_BATTERY_PERCENT
       if (fusor.angle1() < -M_PI / 3)
#endif
       {
             sound_library_.SayBatteryLevel();
              sound_library_.SayNumber(battery_monitor.battery(), SAY_DECIMAL);
              sound_library_.SayVolts();
            return;
      }
#endif // FETT263_SAY_BATTERY_VOLTS
#ifdef FETT263_SAY_BATTERY_PERCENT
     sound_library_.SayBatteryLevel();
    sound_library_.SayNumber(battery_monitor.battery_percent(), SAY_WHOLE);
    sound_library_.SayPercent();
#endif
}

void DoBattery() {
  SayBattery();
  doeffect...
}


Another way to do it would be to use helper functions to reduce duplicated code:

void SayBatteryVolts() {
  sound_library_.SayBatteryLevel();
    sound_library_.SayNumber(battery_monitor.battery(), SAY_DECIMAL);
    sound_library_.SayVolts();
}
void SayBatteryPercent() {
    sound_library_.SayBatteryLevel();
    sound_library_.SayNumber(battery_monitor.battery_percent(), SAY_WHOLE);
    sound_library_.SayPercent();
}

void DoBattery() {
#if defined (FETT263_SAY_BATTERY_PERCENT) && defined (FETT263_SAY_BATTERY_VOLTS)
   if (fusor.angle1() < - M_PI / 3) {
      SayBatteryVoltage();
   } else {
       SayBatteryPercent();
   }
#elif defined(FETT263_SAY_BATTERY_PERCENT)
   SayBatteryPercent();
#elif defined(FETT263_SAY_BATTERY_VOLTAGE)
  SayBatteryVoltage();
#endif
  doeffect..
}

I think I like the last one the best, because we can put those helper functions directly in the sound library.

fdarosa2663: Good idea, updated and moved to sound_library

PR #489, opened 2022-12-21
Add FETT263_EDIT_SETTINGS_MENU

fdarosa2663:
Adding a "Settings Only" Menu option for Editing (Volume, Gestures, Blade Length, Brightness).  The bulk of this PR is just re-organizing the functions and case statements for the affected menus to separate from full Edit Mode in the new #if sections.  The only code changes are accounting for changing the MENU_TOP to MENU_SETTINGS_SUB as the starting menu.


props/saber_fett263_buttons.h
```diff
@@ -785,7 +829,7 @@ EFFECT(push); // for Force Push gesture in Battle Mode
 EFFECT(quote); // quote on force effect
 EFFECT(tr);
 EFFECT2(trloop, trloop);
-#ifdef FETT263_EDIT_MODE_MENU
+#if defined(FETT263_EDIT_MODE_MENU) || defined(FETT263_EDIT_SETTINGS_MENU)
```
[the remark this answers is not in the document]
fdarosa2663: Updated, thanks.

props/saber_fett263_buttons.h
```diff
@@ -599,6 +635,10 @@ CUSTOM SOUNDS SUPPORTED (add to font to enable):
 #define FETT263_SAY_BATTERY_PERCENT
 #endif
 
+#if defined(FETT263_EDIT_MODE_MENU) || defined(FETT263_EDIT_SETTINGS_MENU)
+#define FETT263_USE_SETTINGS_MENU
```
profezzorn, line 639: I recommend renaming this to FETT263__USE_SETTINGS_MENU (note, *two* underscores)
That way you can keep defines which should not be used in config files separate from defines that can/should be used in config files.

props/saber_fett263_buttons.h
```diff
@@ -2488,18 +2459,348 @@ SaberFett263Buttons() : PropBase() {}
...
-        DOVALIDATE(current_preset_);
-        current_preset_.Save();
-        DOVALIDATE(current_preset_);
-        show_preview_.Stop(blade_preview_);
-        UpdateStyle();
-        menu_type_ = MENU_COLOR_SUB;
-        MenuSelect();
-        choice_ = false;
-        blade_preview_ = 0;
+#endif
```
profezzorn, line 2462: Duplicating all this code is bad.
We need to find a way to do this that doesn't cause so much code duplication.

fdarosa2663: This isn't duplicated, I moved it from other locations in the same switch to group together.  You'll see corresponding code removed from the lines lower down.

profezzorn: Sorry about that, it's really hard to tell what's going on with this change since there is so much stuff moving around. Smaller changes would have been nice.

PR #495, opened 2023-01-01
EFFECT_FAST_OFF Update

fdarosa2663:
Update to allow for EFFECT_FAST_OFF to skip POSTOFF sound and effect, will also require separate change to prop.


profezzorn:
This change will make the off->postoff unlinked and not gapless, which is not ok.

fdarosa2663:
How can we separate the in and pstoff then?  I wasn't able to get it to work another way, maybe I missed something.  The goal is to be able to set a retraction (Off) without the pstoff playing.

profezzorn:
I think you're going to have to have OFF_NO_POSTOFF (which I think should be called OFF_FAST) set in->next to nullptr. (And set in->next to &SB_pstoff in all other cases.. (before the switch.))

fdarosa2663:
Thanks, I have updated.

common/saber_base.h
```diff
@@ -42,6 +42,7 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(QUOTE)			\
     DEFINE_EFFECT(SECONDARY_IGNITION)		\
     DEFINE_EFFECT(SECONDARY_RETRACTION)		\
+    DEFINE_EFFECT(OFF)                          \
```
profezzorn, line 45: Why do we need this.
We already have a way to trigger effects in InOutHelperTr

sound/hybrid_font.h
```diff
@@ -503,7 +506,7 @@ class HybridFont : public SaberBase, public Looper {
           PlayPolyphonic(&SFX_in);
 	  hum_fade_out_ = 0.2;
         }
-	check_postoff_ = !!SFX_pstoff;
+        if (off_type != OFF_FAST) check_postoff_ = !!SFX_pstoff;
```
profezzorn, line 509: probably better to always set it:

check_postoff_ = !!SFX_pstoff && off_type != OFF_FAST;

fdarosa2663: updated

common/saber_base.h
fdarosa2663: This is for use in TrDoEffect<> to activate normal Off() instead of FAST_OFF

profezzorn: A) It needs a better name.
B) It needs comments to explain what it's for.

profezzorn: Actually, let's leave this for a later PR.
This thing will need support from the prop to work properly, right?
Let's add it when that support is added, that way I can check it and understand it properly.

fdarosa2663: OK, removed.  Will add back with prop update and include comments.

PR #497, opened 2023-01-02
Improve false swing ignition from motion chip initialize

fdarosa2663:
Replacing fixed 2000ms prevention for swing ignition after Off() to check for motion chip initializing on boot and MOTION_WAKE instead.


props/saber_fett263_buttons.h
```diff
@@ -5945,7 +5950,7 @@ SaberFett263Buttons() : PropBase() {}
         if (!saved_gesture_control.gestureon) return true;
         if (!saved_gesture_control.swingon) return true;
         // Due to motion chip startup on boot creating false ignition we delay Swing On at boot or motion restart for 2000ms
-        if (!menu_ && millis() - saber_off_time_millis_ > 2000) {
+        if (!menu_ && motion_startup_) {
```
profezzorn, line 5953: can't you just do:

   if (!menu && millis() - saber_off_time_millis_ > 2000 && fusor.ready()) {

here?
Seems like very nearly the same thing (but simpler)

fdarosa2663: No, this adds a 2000ms delay on ANY retraction, users wanted to be able to immediately use the swing ignition after retraction and this prevented.  The motion_startup_ bool is only false during boot and after the MOTION_TIMEOUT occurs (which are when the false swings are occurring anyway) so all other retractions still allow an immediate swing ignition.

profezzorn: Ok, but with this change, it's essentially setting motion_startup_ to fusor.ready() all the time, except when:
A) swinging
B) 2 seconds after saber was turned off
C) 2 seconds after boot.

During these times motion_startup_ remains at it's previous value, which may be true or false, depending on what was going on.

Maybe this is what you want?

if (!menu_ && millis() > 2000 + PROFFIEOS_STARTUP_DELAY + PROFFIEOS_SD_STARTUP_DELAY && fusor.ready()) {

fdarosa2663: I'm not familiar with PROFFIEOS_STARTUP_DELAY and PROFFIEOS_SD_STARTUP_DELAY, do they apply only on boot or also after MOTION_WAKE (from MOTION_TIMEOUT).  There's two things at work here; when you first boot there's a ramp up of the motion chip that triggers a swing, the millis() > 2000ms handles fine BUT when we do SaberBase::RequestMotion(); for the MOTION_WAKE (at line 4771 for example), this also causes the motion chip to ramp up and trigger a swing.  The issue is FETT263_MOTION_WAKE_POWER_BUTTON is meant to just wake up motion but not ignite the blade, but the swing then ignites the blade so it causes this define to not work as intended.  fusor.ready() returns true too quickly in testing so I had to add a step in between using the bool so that this define and the initial boot can both happen without false swing ignitions.  
Open to ideas but it needs to handle both the initial boot and the MOTION_WAKE and all previous attempts only handled one or the other not both.  In testing, this solution handles both as expected without causing unnecessary swing ignition delays on normal retraction and operation which is the goal.

profezzorn: ok, but that's not what the code does.

I suggest putting this in loop then:

    if (!fusor.ready()) {
       fusor_ready_millis_ = millis();
    }

then change this if statement to:
   if (!menu_ && millis() - fusor_ready_millis_ > 2000) {
   }

This makes the delay 2000 ms after fusor.ready() returns true.
You can also put fusor_ready_millis_ = millis() in the code if you want to trigger the delay "manually".

profezzorn: Btw, it might be better to put this delay in the code that *detects* swings, instead of here.

fdarosa2663: OK thanks, I'll give that a test tomorrow and update accordingly.

fdarosa2663: Thanks, that worked, updated.

props/saber_fett263_buttons.h
```diff
@@ -1918,18 +1918,23 @@ SaberFett263Buttons() : PropBase() {}
...
         SaberBase::RequestMotion();
         if (swinging_ && fusor.swing_speed() < 90) {
           swinging_ = false;
         }
-        if (!swinging_ && fusor.swing_speed() > saved_gesture_control.swingonspeed) {
-          swinging_ = true;
-          Event(BUTTON_NONE, EVENT_SWING);
-        }
+        // Prevent false swing during motion startup
+        if (millis() - fusor_ready_millis_ > 2000) {
```
profezzorn, line 1932: if (A) {
    if (B) {
    }
}

should generally be if (A && B) { }

props/saber_fett263_buttons.h
```diff
@@ -6227,6 +6230,7 @@ SaberFett263Buttons() : PropBase() {}
   bool rehearse_ = false; // Rehearsal Mode active
   bool choreo_ = false; // Choreography Mode active
 #endif
+  uint32_t fusor_ready_millis_; // Fusor "Ready" timer
```
profezzorn, line 6233: initialize to zero

props/saber_fett263_buttons.h

props/saber_fett263_buttons.h

props/saber_fett263_buttons.h
fdarosa2663: Updated with new fusor.ready() improvement

PR #498, opened 2023-01-03
Add TrDoEfffect

fdarosa2663:


transitions/doeffect.h
```diff
@@ -0,0 +1,46 @@
...
+    wavnum_.run(blade);
+    float location = LOCATION;
+    if (location == -1) location = random(32768) / 32768;
+    if (begin_) {
+      int wavnum = wavnum_.calculate(blade);
+      SaberBase::DoEffect(EFFECT, location, wavnum);
+      begin_ = false;
+    }
+  }
+  bool done() {
```
profezzorn, line 36: This function is not needed.

transitions/doeffect.h
```diff
@@ -0,0 +1,46 @@
...
+public:
+  void begin() {
+    TRANSITION::begin();
+    begin_ = true;
+  }
+  void run(BladeBase* blade) {
+    TRANSITION::run(blade);
+    wavnum_.run(blade);
+    float location = LOCATION;
+    if (location == -1) location = random(32768) / 32768;
```
profezzorn, line 29: This needs to be random(32768) / 32768.0, otherwise it's always going to be zero or one.

transitions/doeffect.h
fdarosa2663: removed

transitions/doeffect.h
fdarosa2663: updated

PR #501, opened 2023-01-07
Add CAIWYN_BUTTON_LOCKUP, CAIWYN_BUTTON_CLASH, CAIWYN_SAVE_TRACKS, and CAIWYN_SAVE_TRACK_MODE

Caiwyn:
This adds four new definitions that can be used to customize the behavior of the prop file:
1. CAIWYN_BUTTON_LOCKUP allows the user to configure the power button to trigger a standard lockup when pressed while the saber is on.  This was previously the default behavior for this button, but because this is a niche use case, I have changed the default behavior to trigger a lightning block, to better serve the interests of the general public.  Users can still define this to generate clashes and lockups with a button.
2. CAIWYN_BUTTON_CLASH can be defined in conjunction with CAIWYN_BUTTON_LOCKUP to play a clash sound at the same time as the lockup, to smooth out the sound for short clashes.  Previously this was done by default and could be disabled by defining CAIWYN_BUTTON_NOCLASH, but I have reversed the behavior since CAIWYN_BUTTON_LOCKUP is no longer on by default.
3. CAIWYN_SAVE_TRACKS will update each preset with the currently selected track, so that changing presets or removing power does not force you to cycle through each track again afterwards.
4. CAIWYN_SAVE_TRACK_MODE creates a new ini file that stores the currently-selected track player mode and updates when the track player mode is changed.  This ensures the track player mode remains persistent across battery swaps/recharges.

Additionally, I have added diagnostic messages and error handling to the track player.

### DIFF
--- a/props/saber_caiwyn_buttons.h
+++ b/props/saber_caiwyn_buttons.h
@@ -92,6 +92,16 @@
 //                                  fonts, but if it sounds weird you can
 //                                  disable the clash sound and just use the
 //                                  lockup by itself.
+// #define CAIWYN_SAVE_TRACKS     - Automatically saves the selected track for
+//                                  each preset. Any time the user holds the
+//                                  Aux button for one second to change the
+//                                  active track, the preset is updated to
+//                                  default to that track. This ONLY happens
+//                                  when the user selects the track directly;
+//                                  if the Track Player Mode is set to repeat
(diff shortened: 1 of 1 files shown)


Caiwyn:
Also did a big cleanup of the track player, and tightened up the code.

PR #504, opened 2023-01-11
Update EFFECT_BATTERY_LEVEL

fdarosa2663:


profezzorn:
This is an interesting (and new?) approach.
Triggering the sound library from inside SB callback means that wavnum/wavlen won't work, but on the other hand it becomes possible to trigger "battery" updates from styles, which is nice.

I think it's good thing, but since it's a new approach, some extra testing might be in order.

PR #505, opened 2023-01-11
Add config used to "version"

NoSloppy:


ProffieOS.ino
```diff
@@ -1067,7 +1067,7 @@ class Commands : public CommandParser {
 
     if (!strcmp(cmd, "version")) {
       STDOUT << version
-      << "\nprop: "  TOSTRING(PROP_TYPE)  "\nbuttons: " TOSTRING(NUM_BUTTONS) "\ninstalled: " 
+      << "/n" TOSTRING(CONFIG_FILE) "\nprop: "  TOSTRING(PROP_TYPE)  "\nbuttons: " TOSTRING(NUM_BUTTONS) "\ninstalled: "
```
profezzorn, line 1070: Did you test this?
CONFIG_FILE is already a string, so no need to use TOSTRING
And I think you meant \n, not /n

NoSloppy: oops. Tested on local version, but then started looking at how to maybe strip the path off the config file name and then add a title of "config: ", but couldn't see how to do that, bailed, and went over-simplified-copy-paste and with mistakes as well.
Fixed.
Unless you think we should isolate the file name because you know how to easily.

PR #509, opened 2023-01-20
Add TrLoopN

fdarosa2663:


styles/length_finder.h
```diff
@@ -34,7 +34,7 @@ class LengthFinder {
     if (say_it_ && millis() - last_speak_ > 3000) {
       last_speak_ = millis();
       say_it_ = false;
-#ifdef ENABLE_AUDIO      
+#ifdef ENABLE_AUDIO
```
fdarosa2663, line 37: just removed spaces, it made it's way into my master during an edit

transitions/loop.h
```diff
@@ -16,7 +16,39 @@ class TrLoop : public TR {
...
+    return loops_ == num_loops_ && TRANSITION::done(); 
+  }
+
+private:
+  int loops_;
+  int num_loops_;
+  PONUA SVFWrapper<N> n_;
+};
+
+// Usage: TrLoopUntil<PULSE, TRANSITION, OUT>
```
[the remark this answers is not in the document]
profezzorn: I think this can be made smaller and more efficient by counting down, something like:

void run(BladeBase* blade) {
  if (loops_ < 0) loops_ = n_.calculate(blade);
  if (loops_ >  0 && TRANSITION::done()) {
     if (loops_ > 1) TRANSITION::begin();
     loops_--;
  }
  TRANSITION::run(blade);
}
viod begin() {
  TRANSITION::begin();
 loops_ = -1;
}
bool done() { return loops_ == 0; }

This way we don't need "num_loops_", we only call TRANSITION::done() once, and we only call n_.calculate() once per begin call.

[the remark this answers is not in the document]
fdarosa2663: Thanks, yup much simpler.  Updated and tested.

transitions/loop.h
```diff
@@ -16,7 +16,35 @@ class TrLoop : public TR {
...
 
-// Usage: TrLoop<PULSE, TRANSITION, OUT>
+// Usage: TrLoopN<N, TRANSITION>
+// N: FUNCTION (number of Loops)
+// TRANSITION: TRANSITION
+// Return Value: TRANSITION
+// Runs the specified transition N times.
+
+template<class N, class TRANSITION>
+class TrLoopN : public TRANSITION {
```
profezzorn, line 26: Actually, I guess this should be TrLoopNX, and we then add an alias called TrLoopN which takes an int.

fdarosa2663: Added

transitions/loop.h
```diff
@@ -16,7 +16,39 @@ class TrLoop : public TR {
...
+    TRANSITION::run(blade);
+  }
+  void begin() {
+    TRANSITION::begin();
+    loops_ = -1;
+  }
+  bool done() { return loops_ == 0; }
+
+private:
+  int loops_;
```
profezzorn, line 45: loops_ = 0;

fdarosa2663: fixed

PR #511, opened 2023-01-20
DISABLE_TALKIE

fdarosa2663:
Save 2336 bytes by replacing talkie functions with beeper, will create supporting POD to document

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -603,8 +603,15 @@ public:
   bad_blade:
     STDOUT.println("BAD BLADE");
 #ifdef ENABLE_AUDIO
+#ifdef DISABLE_TALKIE
+    beeper.Beep(0.25, 500);
+    beeper.Beep(0.5, 3000);
+    beeper.Beep(0.5, 1500);
+    beeper.Beep(0.25, 500);
+#else
     talkie.Say(talkie_error_in_15, 15);
     talkie.Say(talkie_blade_array_15, 15);
+#endif
 #endif
 
 #endif
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -606,7 +606,6 @@ public:
 #ifdef DISABLE_TALKIE
     beeper.Beep(0.25, 500);
     beeper.Beep(0.5, 3000);
-    beeper.Beep(0.5, 1500);
     beeper.Beep(0.25, 500);
 #else
     talkie.Say(talkie_error_in_15, 15);
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -114,6 +114,10 @@ private:
 };
 #endif
 
+#ifdef DISABLE_TALKIE
+#include "../sound/sound_library.h"
+#endif
+
 // Base class for props.
 class PropBase : CommandParser, Looper, protected SaberBase {
 public:
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -114,10 +114,6 @@ private:
 };
 #endif
 
-#ifdef DISABLE_TALKIE
-#include "../sound/sound_library.h"
-#endif
-
 // Base class for props.
 class PropBase : CommandParser, Looper, protected SaberBase {
 public:
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -685,10 +685,6 @@ CUSTOM SOUNDS SUPPORTED (add to font to enable):
 #define FETT263_USE_SETTINGS_MENU
 #endif
 
-#if defined(FETT263_SAY_COLOR_LIST) || defined(FETT263_SAY_COLOR_LIST_CC)
-#define SAY_COLOR_LIST
-#endif
-
 #if NUM_BUTTONS < 1
 #error /props/saber_fett263_buttons.h requires 1, 2 or 3 Buttons for operation
 #endif
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -685,6 +685,10 @@ CUSTOM SOUNDS SUPPORTED (add to font to enable):
 #define FETT263_USE_SETTINGS_MENU
 #endif
 
+#if defined(FETT263_SAY_COLOR_LIST) || defined(FETT263_SAY_COLOR_LIST_CC)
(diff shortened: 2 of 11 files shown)


sound/sound_library.h
```diff
@@ -8,6 +8,7 @@ enum SayType {
   SAY_WHOLE,
 };
 
+#ifdef SAY_COLOR_LIST
```
fdarosa2663, line 11: if sound_library is used and we don't use SAY_COLOR_LIST save memory for EFFECT and enum

styles/length_finder.h
```diff
@@ -34,7 +34,7 @@ class LengthFinder {
     if (say_it_ && millis() - last_speak_ > 3000) {
       last_speak_ = millis();
       say_it_ = false;
-#ifdef ENABLE_AUDIO      
+#if defined(ENABLE_AUDIO) && !defined(DISABLE_TALKIE)
```
profezzorn, line 37: This should be an error.
Lengthinder is useless without talkie.

sound/sound_library.h
```diff
@@ -265,9 +267,11 @@ class SoundLibrary {
   void SaySettingsMenu() { Play("msetsub.wav"); }
   void SayStyleSettings() { Play("stylstm.wav"); }
 
+#ifdef SAY_COLOR_LIST
```
profezzorn, line 270: This seems unrelated

props/prop_base.h
```diff
@@ -114,6 +114,10 @@ class SoundQueue {
 };
 #endif
 
+#ifdef DISABLE_TALKIE
+#include "../sound/sound_library.h"
```
profezzorn, line 118: Is this used anywhere?

props/prop_base.h
```diff
@@ -603,9 +607,16 @@ class PropBase : CommandParser, Looper, protected SaberBase {
   bad_blade:
     STDOUT.println("BAD BLADE");
 #ifdef ENABLE_AUDIO
+#ifdef DISABLE_TALKIE
```
profezzorn, line 610: These needs to be moved to a common location, and have comments that describe what they sound like.

props/saber_fett263_buttons.h
```diff
@@ -685,6 +685,10 @@ CUSTOM SOUNDS SUPPORTED (add to font to enable):
 #define FETT263_USE_SETTINGS_MENU
 #endif
 
+#if defined(FETT263_SAY_COLOR_LIST) || defined(FETT263_SAY_COLOR_LIST_CC)
+#define SAY_COLOR_LIST
```
profezzorn, line 689: unrelated?

profezzorn (commented):
So, it seems that this change doesn't actually disable the talkie class itself.
I suspect that much more memory would be saved if the talkie class was completely left out.

styles/length_finder.h
fdarosa2663: thought user could just use Serial Monitor in this case?

fdarosa2663:
I'll try removing talkie class completely and other suggestions and make updates

props/saber_fett263_buttons.h
fdarosa2663: it's an additional memory saving for sound_library when SAY_COLOR_LIST defines are not enabled.  So tangentially related?

props/prop_base.h
fdarosa2663: I assume you mean a comment explaining what the define does and the beeps that replace the messages?  Where would you recommend?

I had attempted to centralize all errors to a single file but it would not compile so I left the change to beeper in the existing location.

fdarosa2663:
Looks like completely removing Talkie class frees up 7576 bytes, so definitely better.

Should I go ahead and update the other props that are currently using talkie, as the define will now create a compile error when used - or - should I just notify them and leave it to those creators to handle?

Just need a few replies to above conversations to finish up.

profezzorn:
7.5k, not bad :)

I think it depends on the props, and how complicated they are and if they are actively maintained or not.
I think it's ok to have compile errors if you try to use a feature that needs talkie and also have DISABLE_TALKIE in your config file. We just need to warn people that such things can happen in the documentation for the define.  (LengthFinder would be a good example of this IMHO.)

For the most part, I think it's fine to leave the props as is and just let them cause errors with DISABLE_TALKIE, but if you want to make things better for users, you can:

1. Find cases where talkie could be replaced with beeps, and add code to the props for that.
2. Find cases where talkie is entirely optional, and just use #ifndef to make it not do that.
3. Find props where talkie is integral, and removing it will make no sense and add #error messages for them.

I would probably do all of this in a followup PR, rather than trying to include any of it in this PR though.

We may also be able to create a dummy talkie class which causes more readable errors if you try to call Say() on it, although I would have to read up on how to do that.

Looks like this is how you create custom error messages for method calls:
https://stackoverflow.com/questions/34745581/forbids-functions-with-static-assert/34746077#34746077

fdarosa2663:
OK, for now I'll leave the other props to handle how they want to.  I plan to do a write up on POD for the define with the replacement beeps for the respective talkie we replace.

I probably need help on the static assert, unfortunately I didn't really follow all of that.  I suppose I could just take the define our of lengthfinder and let the compile error handle if a user tried to use.

For a comment on the function where would you recommend it?  In the .ino?

sound/talkie.h
```diff
@@ -1,6 +1,13 @@
 #ifndef SOUND_TALKIE_H
 #define SOUND_TALKIE_H
 
+/* DISABLE_TALKIE define will save memory by replacing the default
```
fdarosa2663, line 4: Added comment to talkie.h for define, let me know if you would prefer somewhere else.

fdarosa2663:
Made updates, added comment for define in talkie.h but can move if you have a better location.

profezzorn:
I did a chunk of this work, not sure if it makes sense to continue this PR or if it's easier to create a new one.

fdarosa2663:
We can close this one, do you need me to do anything from my end, I should have time in the morning?

profezzorn:
Error.h still needs to be updated or DISABLE_TALKIE isn't going to actually work.

PR #514, opened 2023-01-26
SAY_COLOR_LIST and DISABLE_TALKIE support for prop

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -2346,8 +2350,12 @@ SaberFett263Buttons() : PropBase() {}
       StopTrackPlayer();
     }
     if (!SFX_medit) {
+#ifdef DISABLE_TALKIE
```
profezzorn, line 2353: This should call errors::error_in_font_directory() instead.

fdarosa2663: updated

PR #515, opened 2023-01-28
Add EFFECT_TRACK Support

fdarosa2663:


props/saber_fett263_buttons.h
```diff
@@ -4859,7 +4873,7 @@ SaberFett263Buttons() : PropBase() {}
         return true;
 
       case EVENTID(BUTTON_POWER, EVENT_SECOND_HELD_LONG, MODE_OFF):
-#if defined(FETT263_EDIT_MODE_MENU) || defined(FETT263_EDIT_SETTINGS_MENU
+#if defined(FETT263_EDIT_MODE_MENU) || defined(FETT263_EDIT_SETTINGS_MENU)
```
fdarosa2663, line 4876: missing ")" added

props/saber_fett263_buttons.h
```diff
@@ -2053,6 +2053,18 @@ SaberFett263Buttons() : PropBase() {}
     }
   }
 
+  void PlayEffectTrack() {
+    if (!track_player_->isPlaying()) {
+      num_tracks_ = RunCommandAndGetSingleLine("list_current_tracks", nullptr, 0, 0, 0);
```
profezzorn, line 2058: You should probably cache this so you don't need to do it while the saber is on.

fdarosa2663: Moved to chdir() override {}

PR #517, opened 2023-01-29
Errors msgs to display

NoSloppy:
missing something though...


NoSloppy:
Well, the error messages CAN show, or want to anyway, but do not have enough priority it seems.
At boot, audible errors play completely, followed by the SCREEN_STARTUP message "ProffieOS" and that's it. no error msg.
The blade style "name" message shows even during "Font directory not found" when doing next/prev preset.
In fact, the only time I can get the messages to show on their own is after doing Blade Detect, where it then preempts the style "name", and it takes a while before it shows as well.

display/ssd1306.h
```diff
@@ -488,7 +493,7 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
     }
   }
 
-  void SetMessage(const char* text) {
+  void SB_Message(const char* text) {
```
profezzorn, line 496: Just have SB_Message() call SetMessage instead.
Also: override.

display/ssd1306.h
```diff
@@ -732,6 +737,7 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
   Screen screen_ = SCREEN_STARTUP;
   Screen next_screen_ = SCREEN_UNSET;
   char message_[32];
+  bool from_sbmessage_ = false;
```
profezzorn, line 740: what does this do?

display/ssd1306.h
NoSloppy: void SB_Message(const char* text) override {
    SetMessage(message_);
  }
  
  In this file doesn't work. Hint?

profezzorn: Need to see the error message to give a hint.
(Did you put SetMessage() back?)

NoSloppy: I reverted all SB_Messages in the file back to SetMessage, yes.
Testing by sending `sd_card_not_found` in Serial Monitor.
No error message, just nothing displayed. Actually, I think I see garbage on the screen.
However, `setmessage hello`  does work. `setmessage hello\nworld` also, but line return ignored.

NoSloppy: Maybe I missed adding a DoMessage in saber_base?

display/ssd1306.h
NoSloppy: This is a prelude to having the wavlen stuff; a bypass if it's an error message. ...and now that you mention it, font.bmp syncing is prob not needed. 
Killing it.

display/ssd1306.h
NoSloppy: and just tested flipping all SetMessages in here to SB_Message, things work.

profezzorn:
I decided to do this with new effect enums instead.
That different display types can use different strings as needed.

NoSloppy:
> I decided to do this with new effect enums instead. That different display types can use different strings as needed.

So clean :)

PR #523, opened 2023-02-06
Updated add OLED images

NoSloppy:
Not sure why indention fixes appear as the same here. GitHub weirdness. As this screenshot shows, it's a correction.
Screen Shot 2023-02-06 at 2 35 52 AM

### DIFF
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -791,19 +791,19 @@ public:
     if (!frame_available_) {
       // STDERR << "ADVANCE=" << advance_ << " last_file_pos_= " << last_state_.file_pos << " ypos=" << last_state_.ypos << "\n";
       if (!advance_) {
-  file_.Seek(last_state_.file_pos);
-  ypos_ = last_state_.ypos;
+        file_.Seek(last_state_.file_pos);
+        ypos_ = last_state_.ypos;
       } else {
-  advance_ = false;
+        advance_ = false;
       }
       ReadState state;
       state.ypos = ypos_;
       state.file_pos = file_.Tell();
       if (ReadImage(&file_)) {
-  frame_available_ = true;
-  last_state_ = state;
+        frame_available_ = true;
+        last_state_ = state;
       } else {
-  STDERR << "read image fail\n";
+        STDERR << "read image fail\n";
         eof_ = true;
       }
     }
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -463,7 +463,7 @@ public:
 
   void SB_On2() override {
     if (IMG_out) {
-      SetFile(&IMG_out, round(SetDuration(font_config.ProffieOSOutImageDuration)));
+      ShowFileWithSoundLength(&IMG_out, font_config.ProffieOSOutImageDuration);
     }
   }
 
@@ -475,15 +475,16 @@ public:
     }
   }
 
-  float SetDuration(float image_duration) {
-  return (image_duration == 0) ? SaberBase::sound_length * 1000 : image_duration;
-}
+  void ShowFileWithSoundLength(Effect* e, float duration) {
+    float dur = (duration == 0) ? SaberBase::sound_length * 1000 : duration;
+    ShowFile(e, round(dur));
+  }
 
  void SB_Effect(EffectType effect, float location) override {
     switch (effect) {
       case EFFECT_BOOT:
         if (IMG_boot) {
-          ShowFile(&IMG_boot, round(SetDuration(font_config.ProffieOSBootImageDuration)));
+          ShowFileWithSoundLength(&IMG_boot, font_config.ProffieOSBootImageDuration);
         } else {
           SetScreenNow(SCREEN_STARTUP);
         }
@@ -492,7 +493,7 @@ public:
         looped_on_ = Tristate::Unknown;
         looped_idle_ = Tristate::Unknown;
         if (IMG_font) {
-          ShowFile(&IMG_font, round(SetDuration(font_config.ProffieOSFontImageDuration)));
+          ShowFileWithSoundLength(&IMG_font, font_config.ProffieOSFontImageDuration);
         } else if (prop.current_preset_name()) {
           SetMessage(prop.current_preset_name());
           SetScreenNow(SCREEN_MESSAGE);
@@ -533,39 +534,39 @@ public:
   void SB_Effect2(EffectType effect, float location) override {
     switch (effect) {
       case EFFECT_BLAST:
-      ShowFile(&IMG_blst, round(SetDuration(font_config.ProffieOSBlastImageDuration)));
+      ShowFileWithSoundLength(&IMG_blst, font_config.ProffieOSBlastImageDuration);
       return;
     case EFFECT_CLASH:
-      ShowFile(&IMG_clsh, round(SetDuration(font_config.ProffieOSClashImageDuration)));
+      ShowFileWithSoundLength(&IMG_clsh, font_config.ProffieOSClashImageDuration);
       return;
     case EFFECT_FORCE:
-      ShowFile(&IMG_force, round(SetDuration(font_config.ProffieOSForceImageDuration)));
+      ShowFileWithSoundLength(&IMG_force, font_config.ProffieOSForceImageDuration);
       return;
     case EFFECT_PREON:
       ShowFile(&IMG_preon, round(SaberBase::sound_length * 1000));
       return;
     case EFFECT_POSTOFF:
-      ShowFile(&IMG_pstoff, round(SetDuration(font_config.ProffieOSPstoffImageDuration)));
+      ShowFileWithSoundLength(&IMG_pstoff, font_config.ProffieOSPstoffImageDuration);
       return;
 /* To-Do, possibly differently
 #ifdef OLED_USE_BLASTER_IMAGES
     case EFFECT_RELOAD:
-      ShowFile(&IMG_reload, round(SetDuration(font_config.ProffieOSReloadImageDuration)));
+      ShowFileWithSoundLength(&IMG_reload, font_config.ProffieOSReloadImageDuration);
       return;
     case EFFECT_EMPTY:
-      ShowFile(&IMG_empty, round(SetDuration(font_config.ProffieOSEmptyImageDuration)));
+      ShowFileWithSoundLength(&IMG_empty, font_config.ProffieOSEmptyImageDuration);
       return;
     case EFFECT_JAM:
-      ShowFile(&IMG_jam, round(SetDuration(font_config.ProffieOSJamImageDuration)));
+      ShowFileWithSoundLength(&IMG_jam, font_config.ProffieOSJamImageDuration);
       return;
     case EFFECT_CLIP_IN:
-      ShowFile(&IMG_clipin, round(SetDuration(font_config.ProffieOSClipinImageDuration)));
+      ShowFileWithSoundLength(&IMG_clipin, font_config.ProffieOSClipinImageDuration);
       return;
     case EFFECT_CLIP_OUT:
-      ShowFile(&IMG_clipout, round(SetDuration(font_config.ProffieOSClipoutImageDuration)));
+      ShowFileWithSoundLength(&IMG_clipout, font_config.ProffieOSClipoutImageDuration);
       return;
     case EFFECT_DESTRUCT:
-      ShowFile(&IMG_destruct, round(SetDuration(font_config.ProffieOSDestructImageDuration)));
+      ShowFileWithSoundLength(&IMG_destruct, font_config.ProffieOSDestructImageDuration);
       return;
 #endif
 */
@@ -588,7 +589,7 @@ public:
     if (offtype == OFF_IDLE) {
       SetScreenNow(SCREEN_OFF);
     } else if (IMG_in) {
-      ShowFile(&IMG_in, round(SetDuration(font_config.ProffieOSDestructImageDuration)));
+      ShowFileWithSoundLength(&IMG_in, font_config.ProffieOSDestructImageDuration);
     } else if (IMG_idle) {
       ShowFile(&IMG_idle, 3600000.0);
     } else {
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -8,18 +8,19 @@
 #include "monoframe.h"
 
 // Images/animations
-IMAGE_FILESET(font);
 IMAGE_FILESET(boot);
-IMAGE_FILESET(on);
-IMAGE_FILESET(clsh);
+IMAGE_FILESET(font);
 IMAGE_FILESET(blst);
-IMAGE_FILESET(lock);
+IMAGE_FILESET(clsh);
 IMAGE_FILESET(force);
-IMAGE_FILESET(idle);
 IMAGE_FILESET(preon);
 IMAGE_FILESET(out);
 IMAGE_FILESET(in);
 IMAGE_FILESET(pstoff);
+IMAGE_FILESET(on);
+IMAGE_FILESET(lock);
+IMAGE_FILESET(idle);
+/* To-Do, possibly differently
 #ifdef OLED_USE_BLASTER_IMAGES
 IMAGE_FILESET(reload);
 IMAGE_FILESET(empty);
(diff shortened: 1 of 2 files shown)


display/ssd1306.h
```diff
@@ -330,47 +343,48 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
+          display_->DrawText("proffieos", 0,15, Starjedi10pt7bGlyphs);
+        }
+        display_->DrawText(version,0,31, Starjedi10pt7bGlyphs);
+        if (HEIGHT > 32) {
+          display_->DrawText("installed: ",0,47, Starjedi10pt7bGlyphs);
+          display_->DrawText(install_time,0,63, Starjedi10pt7bGlyphs);
+        }
         next_screen_ = SCREEN_PLI;
-        return font_config.ProffieOSFontImageDuration;
+        return font_config.ProffieOSBootImageDuration;
```
profezzorn, line 364: This needs to be backwards compatible, so if the "ProffieOSBootImageDuration" isn't set, then it should still use ProffieOSFontImageDuration.

display/ssd1306.h
```diff
@@ -445,66 +459,127 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
     }
   }
 
+  void SB_On2() override {
+    if (IMG_out) {
+      float dur = (font_config.ProffieOSOutImageDuration == 0) ? SaberBase::sound_length * 1000 : font_config.ProffieOSOutImageDuration;
```
profezzorn, line 464: This gets repeated a lot.
Can we make a helper function?

display/ssd1306.h
```diff
@@ -309,18 +322,18 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
     t_ = 0;
     if (SaberBase::IsOn()) {
       if (SaberBase::Lockup() && IMG_lock && !ignore_lockup) {
-	SetFile(&IMG_lock, 3600000.0);
-      } else if (looped_on_ == Tristate::True) {
-	SetFile(&IMG_on, font_config.ProffieOSOnImageDuration);
+        SetFile(&IMG_lock, 3600000.0);
+      } else if (looped_on_ != Tristate::False) {
```
profezzorn, line 326: This seems like a major change.
Why are we changing it?
Are you sure it's right?

profezzorn (commented):
Please note; no need to stress about getting this in to 7.x.
I know I just started the alpha testing, but we can definitely still add this when it's ready.

display/ssd1306.h
NoSloppy: Well I certainly didn't do it on my own. This was from a conversation with you about something with idle.bmp I believe. I took a look back through the commit history for ssd1306, through closed PRs, PMs, The Crucible, and TRA. 
I can not find the conversation.
I do not pretend to understand what this does, but if I recall, Tristate was added as a major rewrite by you to get idle.bmp, and the Mass Storage stuff worked in correctly.
Unless I changed it completely randomly, I don't see any other reason for the change other than you said to try that and it fixed something.

profezzorn: Thinking about it, it seem reasonable.
If we get here and we don't know if on is looping yet, then we should probably play it to find out, which is what this change does.

display/ssd1306.h
NoSloppy: But if we have a default Boot duration set in hybrid_font, wouldn't that take care of the case where BootImageDuration isn't set in the font config.ini?
Well regardless, in testing all this, setting this to a custom duration here actually seems pretty pointless, as the screen doesn't even initialize fast enough for most wav sounds anyway, and we should probably assure there's enough time to read the info....
So how about we just make it so that if the startup splash screen is going to show because there's no boot image, we just show it for like 4 seconds, period?

display/ssd1306.h
NoSloppy: 100% yes.
I just hadn't worked that out yet. Check now :)

display/ssd1306.h
NoSloppy: works with a looped image and single frame (plays for FontImageDuration), and idle.bmp works, so this seems correct.

display/ssd1306.h
```diff
@@ -445,69 +461,118 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
     if (EscapeIdleIfNeeded() && current_effect_ == &IMG_idle) {
       // We are idle-looping, and usb is connected. Time to stop.
-      SetMessage("usb");
+      SetMessage("    usb\nconnected");
       SetScreenNow(SCREEN_MESSAGE);
     }
   }
 
-  void SB_Effect(EffectType effect, float location) override {
+  float Set_Duration(float image_duration) {
```
profezzorn, line 478: This is a weird naming conversion.
The two "normal" ones are:

   snakeCaseNamingConvention

and

   underscore_naming_convention

Underscores *and* capital letters is a bit unusual.

Also, I think it would be better to have a helper function which wraps ShowFile.
Maybe something like:

void SHowFileWithSoundLength(Effect* e, float duration) { ... }

NoSloppy: ah.
Ok, I changed it to SetDuration() for the naming convention.
I'll need to work on how to incorporate ShowFile.

NoSloppy: > Also, I think it would be better to have a helper function which wraps ShowFile.

Ok, edited.

display/ssd1306.h
```diff
@@ -330,47 +345,48 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
+          display_->DrawText(install_time,0,63, Starjedi10pt7bGlyphs);
+        }
         next_screen_ = SCREEN_PLI;
-        return font_config.ProffieOSFontImageDuration;
+        return 4000;
 
       case SCREEN_PLI:
-        if (!SaberBase::IsOn() && t_ > PLI_OFF_TIME) {
+        if (!SaberBase::IsOn() && t_ >= PLI_OFF_TIME) {
+          STDOUT << "**Time = " << t_ << ". PLI_OFF_TIME expired.\n"; // BC add
```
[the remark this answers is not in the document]
NoSloppy: Changed. 
For using in the future, could you clarify this?
I guess I'm incorrect in assuming STDERR is for errors.

display/ssd1306.h
```diff
@@ -330,47 +345,48 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
+          display_->DrawText("proffieos", 0,15, Starjedi10pt7bGlyphs);
+        }
+        display_->DrawText(version,0,31, Starjedi10pt7bGlyphs);
+        if (HEIGHT > 32) {
+          display_->DrawText("installed: ",0,47, Starjedi10pt7bGlyphs);
+          display_->DrawText(install_time,0,63, Starjedi10pt7bGlyphs);
+        }
         next_screen_ = SCREEN_PLI;
-        return font_config.ProffieOSFontImageDuration;
+        return 4000;
```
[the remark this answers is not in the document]
NoSloppy: In repeated testing, I personally found this to be a good number. Not too short not too long.
It's arbitrary. Whatever you think is best.

display/ssd1306.h
```diff
@@ -379,61 +395,61 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
         } else {
-	  // centered
+        // centered
           display_->DrawText(message_, 0, HEIGHT / 2 + 7, font);
         }
         next_screen_ = SCREEN_DEFAULT;
-	// STDERR << "MESSAGE, millis = " << font_config.ProffieOSFontImageDuration << "\n";
-        return font_config.ProffieOSFontImageDuration;
+        // STDERR << "MESSAGE, millis = " << font_config.ProffieOSFontImageDuration << "\n";
+        return 3000;
```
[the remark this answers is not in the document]
NoSloppy: In repeated testing, I personally found this to be a good number. Not too short not too long.
It's arbitrary. Whatever you think is best.

display/ssd1306.h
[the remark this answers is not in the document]
profezzorn: STDERR always prints to the "default" serial port, while STDOUT prints to the currently active serial port.
Printing to STDOUT in places outside of commands can disrupt the proffieos workbench.

display/ssd1306.h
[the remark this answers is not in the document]
profezzorn: The problem is that it's neither backwards compatible, nor configurable.
If someone was using ProffieOSFontImageDuration for this and relied on it, it won't work anymore and they can't change it. (Without changing the code.)

Not saying that 3000 is a bad choice really,  I just want to think through the alternatives, like maybe another configuration variable?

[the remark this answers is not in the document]
profezzorn: (and whether backwards compatibility would be good or bad.)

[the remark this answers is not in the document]
NoSloppy: I'm trying to remember why I made this change. I think it had to do with the Error messages. Since they use this to display, but there's no wav for them, if the fontconfig duration was set to zero for auto sync, it wouldn't show the message.
Can we check if the incoming message is from ProffieOSErrors then use ProffieOSFontImageDuration, else 3000?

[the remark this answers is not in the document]
profezzorn: That seems more like a problem with the autosync feature.
Like maybe we should change autosync so that if you specify a negative number, then we use the wav length, and if there is no wav file, we negate the number and use that?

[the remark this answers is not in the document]
NoSloppy: Of course that could work, I feel it's a little awkward setting -3000, considering the standard for other things to sync (IgnitionTime<>) is zero. Maybe -1, but still.

> like maybe another configuration variable?

This probably makes the most sense; to break text messages out of the Image Duration anyway as they're 2 different things anyway really, right?
We could just add something like font_config.ProffieOSTextMessageDuration(?) that is a time for all OLED printed messages.
And how about meet in the middle and make it's default 3500ms?
Just tested that and it works great and seems plausible to me.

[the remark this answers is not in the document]
profezzorn: Seems fairly reasonable.
And if that's not set we can fall back to ProffieOSFontImageDuration for backwards compatibility.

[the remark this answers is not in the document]
NoSloppy: I don't know the syntax to check if it's set in font_config vs in hybrid_font.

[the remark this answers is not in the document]
profezzorn: The easiest way is to make the default 0 in hybrid_font, then you can use if statements to see if it's zero or not.

display/ssd1306.h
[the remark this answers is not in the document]
NoSloppy: addressed below.

display/ssd1306.h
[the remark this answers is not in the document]
NoSloppy: Hmm

> if (font_config.ProffieOSTextMessageDuration) is not set we can fall back to ProffieOSFontImageDuration for backwards compatibility.

Hmm, this puts us back to the problem of if FontImageDuration is set to Zero for syncing to wav.

Unless we do an else-if and then double fall back to a static value like 3500?
So:
Set TextMessageDuration default to 0.
Then if font_config.ProffieOSTextMessageDuration !=0, use it
else if font_config.ProffieOSFontImageDuration is not Zero, use it,
else 3500.
Does that work?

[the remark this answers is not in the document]
profezzorn: How about we just make the default -1?
That way we can tell the difference between zero and unset.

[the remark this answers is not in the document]
NoSloppy: Meaning so that zero is a useable option to (not) display text?
ok. But I don't see the point.

display/ssd1306.h
```diff
@@ -445,69 +461,122 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
     switch (effect) {
       case EFFECT_BOOT:
-	if (IMG_boot) {
-	  ShowFile(&IMG_boot, font_config.ProffieOSFontImageDuration);
-	} else if (IMG_idle) {
-	  ShowFile(&IMG_idle, 3600000.0);
-	}
-	return;
+        if (IMG_boot) {
+          ShowFileWithSoundLength(&IMG_boot, font_config.ProffieOSBootImageDuration);
```
profezzorn, line 487: This should SB_Effect2 to be guaranteed to work.

display/ssd1306.h
```diff
@@ -330,47 +345,48 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
+          display_->DrawText("proffieos", 0,15, Starjedi10pt7bGlyphs);
+        }
+        display_->DrawText(version,0,31, Starjedi10pt7bGlyphs);
+        if (HEIGHT > 32) {
+          display_->DrawText("installed: ",0,47, Starjedi10pt7bGlyphs);
+          display_->DrawText(install_time,0,63, Starjedi10pt7bGlyphs);
+        }
         next_screen_ = SCREEN_PLI;
-        return font_config.ProffieOSFontImageDuration;
+        return font_config.ProffieOSTextMessageDuration;
```
profezzorn, line 366: backwards compatibility?
Something like:

if (font_config.ProffieOSTextMessageDuration > 0) return font_config.ProffieOSTextMessageDuration;
return font_config.ProffieOSFontImageDuration;

?

display/ssd1306.h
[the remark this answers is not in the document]
NoSloppy: This is done, i left off testing and didn't get back here yet.
If ProffieOSFontImageDuration has been set to 0 for sync purposes, then it would need a fallback default time.
This covers the below comment for STARTUP_SCREEN as well.
The -1 feels clunky though.

display/ssd1306.h
NoSloppy: moved.

display/ssd1306.h
NoSloppy: addressed below.

display/ssd1306.h
```diff
@@ -445,70 +473,124 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
-      case EFFECT_CLASH:
-	ShowFile(&IMG_clsh, font_config.ProffieOSClashImageDuration);
-	return;
-      case EFFECT_FORCE:
-	ShowFile(&IMG_force, font_config.ProffieOSForceImageDuration);
-	return;
+        looped_on_ = Tristate::Unknown;
+        looped_idle_ = Tristate::Unknown;
+        if (IMG_font) {
+          ShowFileWithSoundLength(&IMG_font, font_config.ProffieOSFontImageDuration);
```
profezzorn, line 501: Needs to move to SB_Effect2

display/ssd1306.h
```diff
@@ -522,6 +604,8 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
   void SB_Off(OffType offtype) override {
```
profezzorn, line 604: SB_Off2

display/ssd1306.h
```diff
@@ -445,70 +473,124 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
...
     }
   }
 
-  void SB_Effect(EffectType effect, float location) override {
+  void ShowFileWithSoundLength(Effect* e, float duration) {
+    float dur = (duration == 0) ? SaberBase::sound_length * 1000 : duration;
+    ShowFile(e, round(dur));
+  }
+
+ void SB_Effect(EffectType effect, float location) override {
```
profezzorn, line 495: Might be easier to just move everything into SB_effect2

display/ssd1306.h
NoSloppy: My comment seems to keep disappearing.
This is addresses above.

display/ssd1306.h
NoSloppy: I honestly don't quite get what that is/does/difference. 
I will remove All things then from SB_EFffect into SB_Effect2 ?

NoSloppy: moved. SB_Off2 also.

profezzorn: SB_****2 was added specifically so that SB_**** can run *first* and start any sounds that needs to play.
Afterwards we call SB_****2, which will be able to know what sound was started, how long it is etc.

NoSloppy: Perfect, thanks.

PR #526, opened 2023-02-07
Fix CAIWYN_SAVE_TRACKS by saving after track change instead of before

Caiwyn:
This fixes a bug in CAIWYN_SAVE_TRACKS.  The save is meant to occur whenever a new track is manually selected, but it was happening before the new track was in memory, causing it to save the previous track instead of the newly-selected track.

### DIFF
--- a/props/saber_caiwyn_buttons.h
+++ b/props/saber_caiwyn_buttons.h
@@ -92,6 +92,16 @@
 //                                  fonts, but if it sounds weird you can
 //                                  disable the clash sound and just use the
 //                                  lockup by itself.
+// #define CAIWYN_SAVE_TRACKS     - Automatically saves the selected track for
+//                                  each preset. Any time the user holds the
+//                                  Aux button for one second to change the
+//                                  active track, the preset is updated to
+//                                  default to that track. This ONLY happens
+//                                  when the user selects the track directly;
+//                                  if the Track Player Mode is set to repeat
+//                                  all tracks, the saved track is not updated
+//                                  when the next track is automatically
+//                                  played.
 
 #ifndef PROPS_CAIWYN_BUTTONS_H
 #define PROPS_CAIWYN_BUTTONS_H
@@ -327,10 +337,6 @@ public:
         RunCommandAndFindNextSortedLine<128>("list_current_tracks", nullptr, nullptr, next_track, false);
       }
       strcpy(current_track_, next_track);
-#ifdef SAVE_TRACK
-      current_preset_.track = mkstr(current_track_);
-      current_preset_.Save();
-#endif
       PlayTrack();
     } else {
       StartOrStopTrack();
@@ -460,6 +466,10 @@ public:
 #endif
         if (track_player_) {
           NextTrack();
+#ifdef CAIWYN_SAVE_TRACKS
+          current_preset_.track = mkstr(current_track_);
+          current_preset_.Save();
+#endif
         } else {
           next_preset();
           strcpy(current_track_,current_preset_.track.get());
--- a/props/saber_caiwyn_buttons.h
+++ b/props/saber_caiwyn_buttons.h
@@ -68,8 +68,7 @@
 // monce.wav                - Set Track Player to play a single track one time
 // mloop.wav                - Set Track Player to repeat a single track
 // mrotate.wav              - Set Track Player to repeat all tracks
-// mselect.wav              - Next Track (omit if you don't want a confirmation
-//                                        sound when the track is changed)
+// mselect.wav              - Next Track (can be omitted if you prefer)
 // ccbegin.wav              - Enter Color Change Mode
 // ccend01.wav              - Save Color and Exit Color Change Mode
 // ccend02.wav              - Reset Color and Exit Color Change Mode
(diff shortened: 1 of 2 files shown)


props/saber_caiwyn_buttons.h
```diff
@@ -369,24 +368,22 @@ class CaiwynButtons : public PropBase {
...
-#ifdef CAIWYN_SAVE_TRACKS
-      current_preset_.track = mkstr(current_track_);
-      current_preset_.Save();
-#endif
-      if (SFX_mselect) {
-        hybrid_font.PlayPolyphonic(&SFX_mselect);
-      }
-    }
+    if (track_player_->isPlaying()) track_player_->Stop();
+    if (track_player_) track_player_.Free();
```
profezzorn, line 372: why free the track player?
Aren't you just going to use it again anyways?

Caiwyn: Good question.  Mainly because PlayTrack() invokes GetFreeWavPlayer() but I suppose I could move that elsewhere... let me see what I can do with that.

Caiwyn: When a track finishes playing on its own, does the track player get freed automatically?  I've built my track player on the assumption that it does...

profezzorn: The track player is a RefPtr<>, which would normally mean that the answer is no, it does not get freed automatically. However there is a bit of code in PropBase::Loop() that frees the track player once it stops playing.

props/saber_caiwyn_buttons.h
```diff
@@ -369,24 +368,22 @@ class CaiwynButtons : public PropBase {
...
-      track_player_.Free();
-#ifdef CAIWYN_SAVE_TRACKS
-      current_preset_.track = mkstr(current_track_);
-      current_preset_.Save();
-#endif
-      if (SFX_mselect) {
-        hybrid_font.PlayPolyphonic(&SFX_mselect);
-      }
-    }
+    if (track_player_->isPlaying()) track_player_->Stop();
```
profezzorn, line 371: Btw, this will crash if track_player_ is null.

Caiwyn: That would be the reason, then.  NextTrack() can be called by a button press, or it can be called when a track finishes playing and the track player mode is PLAYBACK_ROTATE.  But PropBase::Loop() frees the track player when a track finishes, so I have to call GetWavPlayer() to set it back up again.  NextTrack() stops and frees the track player if it's currently playing in anticipation of that.

I'm wondering: could I avoid this by not calling PropBase::Loop() from my own Loop() override?  Could I instead just replicate everything in PropBase::Loop() except the code that frees the tracks?  Like so:

  void Loop() override {
    PropBase::CallMotion();
    if (clash_pending1_) {
      clash_pending1_ = false;
      PropBase::Clash(pending_clash_is_stab1_, pending_clash_strength1_);
    }
    if (clash_pending_ && millis() - last_clash_ >= clash_timeout_) {
      clash_pending_ = false;
      PropBase::Clash2(pending_clash_is_stab_, pending_clash_strength_);
    }
    PropBase::PollScanId();
    PropBase::CheckLowBattery();
  }

profezzorn: You could, but why?
If you do that you would reserve one of the 7 wav players forever, even when you don't need it.

I think what I would do is to change PlayTrack to do:

if (!track_player_) track_player_ = GetFreeTrackPlayer();

That way, it would re-use the same track player if there already is one, or try to get a new one if there isn't.

Then there would be no reason to free the track player.

Caiwyn:
Created a PollTrackPlayer() virtual function in prop_base.h and moved the code that frees the track player when it's done playing to that.  PropBase::Loop() now invokes PollTrackPlayer(), which can be overridden by individual prop files when creating their own track players.  This enables me to change the behavior without breaking existing functionality for anyone else.

profezzorn:
That's a good way to stop the PropBase from freeing the track_player_, but I'm still not clear on why you're doing it.
Why do you need the track player to not be freed?

Caiwyn:
Well, you had asked why there's a routine in NextTrack() to free the track player if I was just going to use it again, and I thought it was a good question, and that led to me thinking, "why free the track player at all until I'm actually done with it?"  Freeing the track_player_ automatically after a track is done playing makes sense when you only have a single track and you're only playing it once.  But when you're playing the same track on repeat or multiple tracks in a row, freeing and reinitializing the track_player_ on each track seems unnecessary.

However, after tinkering with the code, I realize there's not an easy way to distinguish between stopping the track to advance to the next track and stopping the track to shut off the track player, at least not without freeing the track player.  Every time I stop the track, my own PollTrackPlayer() function, which runs in the loop, thinks this means the track is done playing and it's time to stop and free the track player anyway.  So I'm reverting the changes to prop_base.h and I've cleaned up the track player in my prop file to just let PropBase::Loop() handle freeing the track player.

props/saber_caiwyn_buttons.h
```diff
@@ -338,92 +311,76 @@ class CaiwynButtons : public PropBase {
     track_player_ = GetFreeWavPlayer();
```
profezzorn, line 311: Since NextTrack doesn't free the track player anymore, there is a now a minor problem here.
GetFreeWavPlayer() can fail to allocate a player even if we already have one, because it's reserved and already stored in track_player_.

I recommend:

if (!track_player_) track_player_ = GetFreeWavPlayer();

Caiwyn: I assumed PropBase::Loop() would resolve that by freeing the track player after it was stopped.  Is that not the case?

Caiwyn: This is done.

PR #527, opened 2023-02-10
Add EFFECT_OFF_CLASH support

fdarosa2663:
This was missed, adding support for Alpha


props/saber_fett263_buttons.h
```diff
@@ -4733,7 +4733,7 @@ SaberFett263Buttons() : PropBase() {}
     }
 #elif defined(FETT263_SAY_BATTERY_PERCENT)
     sound_library_.SayBatteryPercent();
-#elif defined(FETT263_SAY_BATTERY_VOLTAGE)
+#elif defined(FETT263_SAY_BATTERY_VOLTS)
```
fdarosa2663, line 4736: typo in define

profezzorn: One PR per feature would be better, but I'll let it slide this time.

PR #531, opened 2023-02-19
add IDLE_OFF_TIME expired printout

NoSloppy:


props/prop_base.h
```diff
@@ -1154,6 +1154,7 @@ class PropBase : CommandParser, Looper, protected SaberBase {
       last_on_time_ = millis();
     }
     if (millis() - last_on_time_ > IDLE_OFF_TIME) {
+      STDERR << "Time = " << millis() - last_on_time_ << ". IDLE_OFF_TIME expired.\n";
```
profezzorn, line 1157: This seems like something that wouldn't be used very often, which kind of makes it a waste of space.
Why do we need it?

NoSloppy: I've found it useful. These are rhetorical questions to answer you:
To be able to look back at the log and see events of when/why accents and OLED have turned off etc...? 
Verify settings are working?
There's a Motion timeout printout, why not a FET timeout printout?
It seems you approved of the PLI timeout message?

"Need" is subjective. Just seemed helpful to me.
By all means, it's not that important that the OS has it.

profezzorn: Debug printouts like these can also be added as needed, when a problem is suspected.
Far more frugal.
Although, maybe we should have something like a VERBOSE define that adds more printouts, that would make it easy to have lots of printouts and remove them when we want to save memory.

NoSloppy: hmm. OK, well this can go away for now.
If I get a chance to scour the OS for omittable messages via define and check the memory savings, I'll let you know.

NoSloppy: issue added as reminder

PR #538, opened 2023-03-06
Allow smoothswings to sync with hum

NoSloppy:
Small enough to incorporate?
Realizing now this isn't a bugfix.

### DIFF
--- a/sound/smooth_swing_v2.h
+++ b/sound/smooth_swing_v2.h
@@ -61,11 +61,13 @@ public:
   void PickRandomSwing() {
     if (!on_) return;
     uint32_t m = millis();
+    RefPtr<BufferedWavPlayer> humplayer = GetWavPlayerPlaying(hybrid_font.getHum());
+    if (!humplayer) return;
+    float start = (font_config.ProffieOSSmoothSwingHumstart == 0) ? m / 1000.0 : humplayer->pos();
     // No point in picking a new random so soon after picking one.
     if (A.player && m - last_random_ < 1000) return;
     last_random_ = m;
     int swing = random(L->files_found());
-    float start = m / 1000.0;
     A.Stop();
     B.Stop();
     L->Select(swing);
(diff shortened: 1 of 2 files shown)


sound/smooth_swing_v2.h
```diff
@@ -61,11 +61,13 @@ class SmoothSwingV2 : public SaberBasePassThrough {
   void PickRandomSwing() {
     if (!on_) return;
     uint32_t m = millis();
+    RefPtr<BufferedWavPlayer> humplayer = GetWavPlayerPlaying(hybrid_font.getHum());
+    if (!humplayer) return;
```
profezzorn, line 65: So, if you swing the saber around while activating, then the smoothswing won't start until you stop swinging, because that's normally when PickRandomSwing() is called.  Which is weird.
On the other hand, if we start swingl/swingh in the middle of a swing, they will start abruptly, which is also weird...

NoSloppy: I will test these scenarios tomorrow.

profezzorn: Any updates?

NoSloppy: Just tested this.
Made a hum.wav that is just myself quickly counting 0-10.
Have 1 smoothswing pair (the musical one, so for testing the "1" downbeat is clear when it happens).
I spun the hilt (no blade) like a baton and ignited during the spins. 
Every single time, both the smoothswings and the hum started at their beginning. Adjusted ducking and physical spin speed to find that good range where I could hear both blended well to confirm.
The only unexpected "side effect" is that the hum restarts at the beginning with every blast or clash (although there are no sounds in the test font). Maybe it does that always anyway and I just never noticed because normal hums drone.

profezzorn: sounds like you made a monophonic font...

PR #539, opened 2023-03-06
Return false on EVENT_TWIST_RIGHT/LEFT events

fdarosa2663:
To handle users moving saber during other events like clash or stab where a twist isn't really applicable but may interfere with other events.


props/saber_fett263_buttons.h
```diff
@@ -5645,32 +5645,36 @@ SaberFett263Buttons() : PropBase() {}
 
 #ifdef FETT263_SPECIAL_ABILITIES
       case EVENTID(BUTTON_NONE, EVENT_TWIST_RIGHT, MODE_ON | BUTTON_AUX):
-        if (!menu_) {
+        if (!menu_ && !swinging_) {
```
profezzorn, line 5648: Do you really want to update current_menu_angle_ when swinging?

props/saber_fett263_buttons.h
```diff
@@ -5792,24 +5799,33 @@ SaberFett263Buttons() : PropBase() {}
...
           current_menu_angle_ = fusor.angle2();
           return true;
         }
-        if (menu_) MenuDial(-1);
-        return true;
+        if (menu_) {
+          MenuDial(-1);
+          return true;
+        }
+        return false;
```
profezzorn, line 5828: I guess this eats modifier buttons if you return true?

fdarosa2663: It was in some of the Serial Monitor outputs users shared, if you were moving the hilt in a way that it detected the turn it would get in the way of other events like clashes, etc.

props/saber_fett263_buttons.h
fdarosa2663: Going to look at just ignoring the DetectMenuTurn() while the saber is swinging altogether.  Just need to test, will update once I make sure it works as intended.

fdarosa2663:
OK, I've updated DetectMenuTurn() so the EVENTs only generate if the saber isn't swinging.  In some of the reports users had "EVENT MENU TURN ..." in Serial Monitor in places I think they were trying to clash or do other action but the menu turn was detected first and interfering.  Now DetectMenuTurn will only do EVENTs when the saber is still, since swinging can involve turning the hilt at times.  
I'm leaving the "return false" for the individual cases as a fall through when we're not expecting to need that EVENT anyway, to prevent it from interfering elsewhere based on how the user may be moving the saber at the time.

PR #542, opened 2023-03-09
Fix dupe case in BC_buttons

NoSloppy:


NoSloppy:
Mistakenly had been testing with wrong file. Should be all fixed now.

PR #544, opened 2023-03-13
More EVENT_TWIST_RIGHT/LEFT return false additions

fdarosa2663:
Return false for remainder of TWIST_RIGHT/LEFT events, when not applicable, to prevent interfering with EVENT_TWIST detection.


fdarosa2663:
Just pinging for 7.3 zip file

PR #547, opened 2023-03-31
Added the DEFAULT_BLASTER_MODE define.

baldusi:
Now users can just use the define to set the default mode for their blasters.

### DIFF
--- a/props/blaster.h
+++ b/props/blaster.h
@@ -2,8 +2,10 @@
 
 Default Startup Mode = STUN
 Add the following to your config file if so desired:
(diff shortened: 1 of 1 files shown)


props/blaster.h
```diff
@@ -66,8 +66,12 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
     MODE_AUTO
   };
 
-  BlasterMode blaster_mode = MODE_STUN;
-
+  #ifdef DEFAULT_BLASTER_MODE
```
profezzorn, line 69: We should also add a comment about this at the top.

PR #550, opened 2023-04-22
Delete dead detonator code.

baldusi:
I thin NoSnoopy was trying to add the self-destruct feature, and the code ended there, but there's simply no way to interact with it.


props/blaster.h
```diff
@@ -200,76 +200,6 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
-#endif
-    SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
-    if (armed_) {
-      SetNextActionF(NEXT_ACTION_BLOW, len);
-    } else {
-      SetNextAction(NEXT_ACTION_NOTHING, 0);
-    }
-  }
-
-  void Loop() override {
```
profezzorn, line 269: It might be better to only remove the Loop() function.

That way, a prop that inherits from blaster.h could just add PollNextAction() to their loop function, and call SelfDestruct() from Event2() to implement selfdestruct.

Code that is never called anywhere is optimized out, so the rest of this code would just get optimized away if nothing calls it.

We have similar things going on in saber.h, there are lots of functions in there which are never called from saber.h, but called from lots of subclasses.

props/blaster.h
```diff
@@ -200,7 +200,8 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
     }
   }
 
-  // Self-destruct pulled from detonator 
+  // Self-destruct pulled from Detonator. Any prop that inherits from blaster.h could just add PollNextAction() to their loop function, and call SelfDestruct() from Event2() to implement selfdestruct.
```
profezzorn, line 203: Line is too long.

baldusi (commented):
Shortened the line.

PR #559, opened 2023-05-25
Power Manager

RSX-Engineering:
Proffie OSx's power manager ported on ProffieOS. Handles peripherals power and deep sleep.
Only tested on ProffieBoard v2:  ~0.5 mA in deep sleep

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.git_*
+.vscode
+.build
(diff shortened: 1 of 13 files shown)


profezzorn:
This is a new feature, it should go ono the master branch, not v7.x branch.
v7.x is only for bugfixes.

PR #560, opened 2023-05-26
Power Manager

RSX-Engineering:
Proffie OSx's power manager ported on ProffieOS. Handles peripherals power and deep sleep.
Only tested on ProffieBoard v2: ~0.5 mA in deep sleep


ProffieOS.ino
```diff
@@ -65,7 +65,7 @@
 #define SAVE_VOLUME
 #define SAVE_PRESET
 #define SAVE_COLOR_CHANGE
-#define SAVE_DYNAMIC_DIMMING
```
profezzorn, line 68: This change does not belong in this PR

ProffieOS.ino
```diff
@@ -426,6 +430,9 @@ const char* previous_current_directory(const char* dir) {
   }
 }
 
+#if defined(XPOWERMAN)
+#include "common/xPowerManager.h"
```
profezzorn, line 434: Why is it called xPowerManager?
Can we drop the "x" ?

common/i2cbus.h
```diff
@@ -99,6 +110,9 @@ class I2CBus : Looper, StateMachine {
   int i;
   bool clock_detected, data_detected;
   bool i2c_detected_ = false;
+#ifdef XPOWERMAN
+  uint32_t shtimeMS = 2000;
```
profezzorn, line 114: I don't understand this variable name, and it's in a very different format from everything else.
I think it should be called something ending in _time_millis_, but I don't know what, because I don't know what it does.

common/sd_card.h
```diff
@@ -6,31 +6,83 @@
 #if defined(ENABLE_SD) && defined(ARDUINO_ARCH_STM32L4)
 
 // Unmount sdcard when we don't need it anymore.
+#if defined(XPOWERMAN) 
+class SDCard : Looper , xPowerSubscriber {
+#else // nULTRA_PROFFIE
```
profezzorn, line 11: These ULTRA_PROFFIE comments are incorrect.

profezzorn (commented):
This is a large PR, and it's going to take me a while to go through it properly.
I did scan through it, and here are my initial comments on it.

common/xPowerManager.h
```diff
@@ -0,0 +1,769 @@
...
+
+            armv7m_systick_enable(); 
+        }
+
+    } powerman;     // global object!
+
+    // POWER DOMAINS
+    // ==================================================================================
+    // Power domain objects must inherit from this class to get registered. 
+    class xPowerDomain {
```
profezzorn, line 270: Since all the power domains are hardcoded, it seems unnecessary to dynamically build a list of powerdomain classes. Seems like we could have an array of them, or even just hardcode them all in the power manager itself.

Alternatively, maybe there should be one power-manager per domain. Using bitfields to turn things on/off doesn't seem all that useful.

common/xPowerManager.h
```diff
@@ -0,0 +1,769 @@
...
+                countdownTimer = 0;
+                return true; 
+            }    
+            countdownTimer -= loopTime;             // not timed out, update timer
+            return false;                 
+        }               
+
+    };
+
+    class PowerDomain_Pixel : public xPowerDomain {
```
profezzorn, line 320: This power domain doesn't seem to do anything, What is it for?

common/xPowerManager.h
```diff
@@ -0,0 +1,769 @@
...
+        }                      
+                                
+        // Check if all the subscribed domains are active
+        bool IsOn() {  return ((powerman.powerState & subscribedDomains) == subscribedDomains); }
+
+        // Keep power ON until timeout on all subscribed domains. Turns domains ON if needed.
+        // Returns true if subscriber went on just now
+        // If specified, timeout_ must point to a vector of uint32_t with the tineout value in millis for each subscribed domain, e.g. 'RequestPower(&(const uint32_t t=100));' or 'RequestPower(&(const uint32_t[] t={100,200}));'
+        // If not specified, domain-specific timeout will be applied to each subscribed domain
+        bool RequestPower(uint32_t* timeout_ = 0) {
```
profezzorn, line 462: Why do you have to be a subscriber to request power from a domain?
It seems like several places in the code inherit from xPowerSubscriber just to keep things alive, but it could just as easily be done with a call like:

  powerman.RequestPower(pwr4_Booster);

Also, the way to specify timeouts seems really weird, and is only ever used with a single timeout, so why not just keep it simple and have *one* timeout?

Finally, code style is to use ending underscore for class variables, not local variables.

profezzorn (commented):
Getting a reasonable grasp on what this does.
Some stuff is quite nice, but there are some assumptions in this code that are not always safe to make.

Anyways, here is another set of comments.

profezzorn:
One more question: How much memory does this take?

profezzorn:
This seems to be a fairly modest power saving btw.
I measured a V2 proffieboard, and it goes down to 0.73mA when not doing anything, without this change.

ProffieOS.ino
RSX-Engineering: Not sure why it appears as a change in the PR, it's 'SAVE BLADE DIMMING' both in the master branch and our commit.

common/xPowerManager.h
```diff
@@ -0,0 +1,769 @@
...
+
+        // Turn ON power domains 
+        bool Activate(PDType_base startUpDomains = PWRMAN_STARTON);            // delayed implementation, see end of file.
+        bool Parse(const char* cmd, const char* arg) override;   // delayed implementation, see end of file.    
+
+    private:
+        static void initWakeups()
+        {
+            // set btn and VUSB pin for for wakeup 
+            stm32l4_exti_notify(&stm32l4_exti, g_APinDescription[powerButtonPin].pin,
```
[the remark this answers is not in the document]
RSX-Engineering: It doesn't really matter because with no button the processor will not go in deep sleep - it is protected by prop.HoldPower(). The assumption here is that if at least one button in installed, it will be tied at powerButtonPin.

common/xPowerManager.h
RSX-Engineering: The idea was to ease integration.  Hardcoding or even a static list would require maintenance, the linked list allows adding power domains by simply inheriting from a class. 

One power manager per domain wouldn't work, its job is to handle the overall interaction between domains and subscribers. Bitfields (or something to that effect) are needed to allow more than one-on-one connections between domains and subscribers - for instance the dac to subscribe to amplifier and booster, or  multiple subscribers to subscribe to the cpu domain.

common/xPowerManager.h
RSX-Engineering: Without subscribers as distinct objects we would loose the callbacks and HoldPower functionality.

The weird specification is what allows a subscriber to request different timeouts for different domains. It's for future extension, feel free to remove it if you don't find any use for it.

RSX-Engineering:
> One more question: How much memory does this take?
872 extra bytes



> This seems to be a fairly modest power saving btw. I measured a V2 proffieboard, and it goes down to 0.73mA when not doing anything, without this change.
We measured 1.1 and were hoping to get less, but I guess that's the best the hardware can support.

common/xPowerManager.h
```diff
@@ -0,0 +1,769 @@
...
+
+            armv7m_systick_disable();               // disable systick 
+            
+            savePORTState(&portA, GPIOA_BASE);
+            savePORTState(&portB, GPIOB_BASE);
+            savePORTState(&portC, GPIOC_BASE);
+            savePORTState(&portH, GPIOH_BASE);
+            
+            // put all pins of port A to default, except :
+            defaultPortState(GPIOA_BASE, 0b1111111111111111);
```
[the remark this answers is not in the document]
RSX-Engineering: Which pins do you think are unsafe and should remain untouched?

common/xPowerManager.h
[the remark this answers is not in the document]
profezzorn: What I mean is that someone could be driving a LED from that pin, or use it as a touch button, or something completely different. We can't assume that it's a power button unless there is a Button class tied to it.

common/xPowerManager.h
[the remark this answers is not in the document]
profezzorn: All of them?
Which ones have a positive impact on power draw?

common/xPowerManager.h
profezzorn: You already have a static list in the bitfield though, so it's already hardcoded anyways.
Using linked lists just takes up code space.

common/xPowerManager.h
profezzorn: I'm not saying we shouldn't have subscribers, it's a useful function.
I'm saying that some classes don't need it, because all they want is to keep some power domain on.

profezzorn:
872 bytes is better than I expected, this change looks like it would take up a lot of space, but I guess it also saves some.

So do we know which parts of the power manager actually saves power?
How much power do we save by stopping the CPU?
How much power do we save by setting the pins back to their original state?
How much power do we save by turning off the power to the SD card?

To put it another way: Do we actually need all of these things, or can we get away with something smaller and simpler?

common/xPowerManager.h
profezzorn: > The weird specification is what allows a subscriber to request different timeouts for different domains. It's for future 
> extension, feel free to remove it if you don't find any use for it.

I don't want to check in things that we don't need unless they are implemented in such a way that they don't use any flash memory.  (Adding a function/method that we never call is fine, but a function with arguments that we never use is not.)

RSX-Engineering:
`To put it another way: Do we actually need all of these things, or can we get away with something smaller and simpler?`

No we don't, and in the light of the latest conversation and measurements, it looks like it would actually be unwise.  The centralized power manager works well on UltraProffie because the hardware configurations are fixed, but since ProffieBoard supports a wider range of installations, distributed power management is more appropriate. It's hard to put the CPU in deep sleep without centralized power management, but here's what we measured:
- 0.52 mA with the full power manager
- 1.075 mA with the original code (no power manager)
- **0.65 mA with the original code but clocked down at 200 kHz instead of 16 MHz**

Saving 130 microamps doesn't worth the extra 876 bytes and the trouble of fixing all the raised concerns, when very similar results can be achieved much easier. We will close this pull request and raise another one, with just that minor change.

PR #561, opened 2023-05-29
Enable standard features by default

allir:
As mentioned in the ProffieOS documentation to make these defines not needed. Changed them so they are still able to be disabled via defines.

### DIFF
--- a/ProffieOS.ino
+++ b/ProffieOS.ino
@@ -57,6 +57,22 @@
 #include CONFIG_FILE
 #undef CONFIG_TOP
 
+#if !defined(ENABLE_AUDIO) && !defined(DISABLE_AUDIO)
+#define ENABLE_AUDIO
+#endif
+
+#if !defined(ENABLE_MOTION) && !defined(DISABLE_MOTION)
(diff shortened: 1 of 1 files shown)


allir:
Tested on Proffieboard v2.2, the only version I have. :P

profezzorn:
There is a much easier way to do this, something like:

#if !defined(ENABLE_SD) && !defined(DISABLE_SD)
#define ENABLE_SD
#endif

No need to change 49 different files.

allir:
Yes, makes sense. Updated setting this for each of the standard features.

ProffieOS.ino
```diff
@@ -57,6 +57,22 @@
 #include CONFIG_FILE
 #undef CONFIG_TOP
 
+#if !defined(ENABLE_ENABLE_AUDIO) && !defined(DISABLE_AUDIO)
```
profezzorn, line 60: ENABLE_ENABLE_AUDIO -> ENABLE_AUDIO

allir: Fixed.

PR #562, opened 2023-05-30
Clock down in standby

RSX-Engineering:
Reduced system clock to 200 kHz instead of 16 MHz when in standby, to reduce power consumption.


common/clock_control.h
```diff
@@ -29,7 +29,7 @@ class ClockControl : public Looper {
 #ifdef PROFFIEOS_VERSION
       stm32l4_system_sysclk_configure(1000000, 500000, 500000);
 #else
-      stm32l4_system_sysclk_configure(16000000, 8000000, 8000000);
+      stm32l4_system_sysclk_configure(200000, 100000, 100000);
```
profezzorn, line 32: So, this side of the #ifdef is meant for older versions of the ardino-proffieboard plugin.
I think since version 2.2.0, the code should hit the version on line 30, which clocks everything down to 1Mhz.
(Not sure why I didn't set it lower though.)

Maybe this is why you got 1.1mA instead of 0.7mA?

RSX-Engineering: There's no PROFFIEOS_VERSION definition in either the plugin or main code, it runs line 32.  
Since there's no other use of this define, maybe we should drop it.

profezzorn: Well I'll be darn.
I think this was meant to be PROFFIEBOARD_VERSION

RSX-Engineering: PROFFIEBOARD_VERSION it is.

profezzorn: So I did some testing, and I didn't see any difference between using a frequency of 200000Hz and 1000000Hz.
(In fact, it kind of seems like 1000000Hz uses less power than 200000Hz)
I checked in the fix for PROFFIEBOARD_VERSION now, thanks for helping me find that problem.

profezzorn:
Fixed defined outside of this PR.

PR #568, opened 2023-07-17
PVLOG effect.h

NoSloppy:


sound/effect.h
```diff
@@ -234,24 +234,24 @@ class Effect {
 
   void Show() {
     if (files_found()) {
-      STDOUT.print("Found ");
-      STDOUT.print(name_);
-      STDOUT.print(" files: ");
+      PVLOG_STATUS.print("Found ");
```
profezzorn, line 237: This are all essential printouts and should not use PVLOG.
Disabling these can only be done by disabling the whole command.

NoSloppy:
ok.

PR #569, opened 2023-07-19
New Config Options for Disabling Preset Changing Props

bmartinson:
# What's New?
The main purpose of this update is to provide two new defines for personal saber configurations that allow users to easily disable preset changing props that may be found irritable during normal saber operation. The name of each of these configuration options were chosen to match similar naming conventions found in both affected prop implementations.

| Deifne | Affected Prop | Description     |
|------------------|-----------|-----------|
| `BC_NO_QUICK_PRESET_CHANGE` | BC Button Props (`props/saber_BC_buttons.h`) | Disable changing presets by twisting when twist gestures are enabled. |
| `DISABLE_FAST_PRESET_CHANGE` | All Props (`props/prop_base.h`) | Disable changing presets quickly while the blade is on. |

This PR also includes some simple file cleanup based on the Google CPP Format Style for affected prop header files.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -182,7 +182,7 @@ public:
 #else
     // No sound means no preon.
     FastOn();
-#endif    
+#endif
   }
 
   void FastOn() {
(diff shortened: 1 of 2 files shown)


profezzorn:
This code is messy, contains lots of extra changes and does two things for no reason.
I'm not sure if this is a good idea or not yet, but we definitely need to split these two defines into two separate PRs so that they can be judged on their own merits.

Also, try to make sure that the PRs don't have a bunch of unrelated changes in it.

bmartinson:
@profezzorn, I've opened the following so they may be considered on their own merits:
#575
#576

I've left the inconsistent code style guidelines and trailing whitespace throughout the files alone for some later pull request.

PR #572, opened 2023-07-27
saber_BC_buttons spoken battery level tweaks

NoSloppy:


props/saber_BC_buttons.h
```diff
@@ -1169,6 +1183,8 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
   bool min_vol_reached_ = false;
   bool sequential_quote_ = false;
   bool spam_blast_ = false;
+  bool speaking_ = false;
```
profezzorn, line 1186: I'm not sure i understand this variable.
Can we add a comment that explains what it's for?

NoSloppy: Comment added.
It's to avoid overlap of battery.wav playing from On-Demand battery level when doing Spoken Battery Level as they both invoke EFFECT_BATTERY LEVEL.

props/saber_BC_buttons.h
```diff
@@ -1115,6 +1126,10 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
         return;
       // On-Demand Battery Level
       case EFFECT_BATTERY_LEVEL:
+        if (speaking_) {
+          speaking_ = false;
```
profezzorn, line 1130: Wouldn't it make more sense to say the battery level here?

NoSloppy: 3 options can be used.
Spoken level in volts, spoken level in percentage, or On-Demand which is just showing the level on the blade with battery.wav sound played.
The way it's submitted is the intended usage.

PR #575, opened 2023-07-28
Add Option for Disabling Fast Preset Change

bmartinson:
Have you ever been incredibly annoyed that you changed your blade style while trying to retract it because of your hold angle? `DISABLE_FAST_PRESET_CHANGE` is a new define that allows you to disable changing presets quickly while the blade is on, which in my experience is a completely undesirable feature.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -504,7 +504,10 @@ public:
 #ifdef SAVE_PRESET
     SaveState(current_preset_.preset_num + 1);
 #endif
+
+#ifndef DISABLE_FAST_PRESET_CHANGE
     SetPresetFast(current_preset_.preset_num + 1);
+#endif // DISABLE_FAST_PRESET_CHANGE
   }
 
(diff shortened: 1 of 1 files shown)


NoSloppy:
I don't see how this would happen using saber_BC_buttons, as you would need to release or double click and release to switch presets, and while pointing up.
Retraction is just holding the button until the blade turns off.
... unless this happens when using a different prop?

props/prop_base.h
```diff
@@ -504,7 +504,10 @@ class PropBase : CommandParser, Looper, protected SaberBase {
 #ifdef SAVE_PRESET
     SaveState(current_preset_.preset_num + 1);
 #endif
+
+#ifndef DISABLE_FAST_PRESET_CHANGE
```
profezzorn, line 508: This breaks the next_preset_fast() contract, and is not the right way to disable functionality.  If you don't want to go to the next preset, then don't call this function.

PR #576, opened 2023-07-28
Add Option for Disabling Preset Change via Gesture

bmartinson:
Have you ever just wanted to ignite or retract your blade through gestures but instead changed presets because of your positioning? `BC_NO_QUICK_PRESET_CHANGE` is a new define that allows you to disable changing presets by twisting when twist gestures are enabled so that you only control preset selection via your button control and it is not tightly linked to gestures.


NoSloppy:
I just submitted #577 which will simply remove this option. I don't think it's needed anyway.

profezzorn:
I think NoSloppy just removed this functionality completely.

bmartinson:
> I just submitted #577 which will simply remove this option. I don't think it's needed anyway.

Thanks for the removal. I look forward to using that standardization in my fork, as well.

PR #578, opened 2023-07-28
just whitespace fixes

NoSloppy:


props/prop_base.h
```diff
@@ -360,14 +360,14 @@ class PropBase : CommandParser, Looper, protected SaberBase {
   void SaveVolumeIfNeeded() {
     if (0
 #ifdef SAVE_VOLUME
-      || dynamic_mixer.get_volume() != saved_global_state.volume
+			|| dynamic_mixer.get_volume() != saved_global_state.volume
```
profezzorn, line 363: Some of these look weird, but I'm not sure if it's just github making it look weird or not....

bmartinson: It's GitHub's interpretation of tab usage instead of spaces for indentation and likely because there is mixed usage in the file.

NoSloppy: This has come up before, but I don't remember the decision. 
Do we want to universalize to using tabs or using spaces for indentation throughout code?

NoSloppy: Ok, so I generated this PR based on the diff I saw from #569, as the hunting had been done already.
Looking now, this function particularly didn't need any editing, but messing with the indentation caused all this.
TL;DR I just edited it back to what should be correct I think,.

bmartinson: As someone who just recently (within the past couple of months) became aware of ProffieOS and is willing to contribute to the project, it would certainly be useful to have some guidelines or rails in place for those things so as to mitigate friction. (I don't want to be a bother and simply assist what I think is a great and fun project).

That being said, my interpretation of reading thru the code is that it more or less follows the [Google C++ style guidelines](https://google.github.io/styleguide/cppguide.html), which would dictate 2 space indentation. Is that an accurate assessment of the guidelines one should follow here?

NoSloppy: Well I just learned you can add "draw_white_space": "all" to settings.
That certainly makes life easier. Hence a few follow up commits, but i think we're straight now.

bmartinson: Yeah, I had an auto formatter that was not initially set to extend Google's and that may have introduced the tabs. I didn't run the formatter on the entire file, though, which you can probably hunt out a few more. i.e. there are some alignment issues near some of the new PVLOG statements, etc. if we're really trying to clean that up.

bmartinson: @NoSloppy if you're at all interested, here's a diff with some others from what was meant to be a quick pass minimally invasive look thru this space for prop_base.h: https://github.com/NoSloppy/ProffieOS-1/compare/BC-buttons-whitespace-fixes...bmartinson:ProffieOS:prop-base-cleanup in my workspace.

NoSloppy: That looks all good to me, except maybe line 791. 
I think the + is meant to align under the v for readability.
If you submit that, i don't need to do anything else;)

profezzorn:
conflicts

NoSloppy:
resolved

PR #579, opened 2023-07-29
Orientation update

NoSloppy:
Updated to use DATA1 and SPKR.
Leaving SDA and SERIAL as well for backward compatibility.


profezzorn:
This will probably be wrong for some upcoming proffieboard....

PR #582, opened 2023-08-01
Add blaster_BC_buttons

NoSloppy:


common/saber_base.h
```diff
@@ -77,6 +77,7 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(UNJAM)			\
     DEFINE_EFFECT(PLI_ON)			\
     DEFINE_EFFECT(PLI_OFF)                      \
+    DEFINE_EFFECT(DESTRUCT)                     \
```
profezzorn, line 80: self-destruct is usually an off-type rather than an effect...

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,861 @@
...
+EFFECT(vmbegin);    // for Begin Volume Menu
+EFFECT(vmend);      // for End Volume Menu
+EFFECT(volup);      // for increse volume
+EFFECT(voldown);    // for decrease volume
+EFFECT(volmin);     // for minimum volume reached
+EFFECT(volmax);     // for maximum volume reached
+EFFECT(quote);      // for playing quotes
+#endif
+
+class BlasterBC : public PROP_INHERIT_PREFIX PropBase {
```
profezzorn, line 184: It can't inherit from the blaster prop?
(If not that's ok, but it should be considered if it will reduce code duplication)

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,861 @@
...
+      }
+    }
+  }
+
+  // Make swings do nothing
+  void DoMotion(const Vec3& motion, bool clear) override {
+    PropBase::DoMotion(Vec3(0), clear);
+  }
+
+RefPtr<BufferedWavPlayer> wav_player;
```
profezzorn, line 520: variables should generally go at the bottom (and be indented)

profezzorn (commented):
Just some initial comments.
IF this things were to inherit from the blaster prop, then there will be lots of changes, so I want to get that out of the way before I review in more detail.

common/saber_base.h
NoSloppy: It is,
```
        case NEXT_ACTION_BLOW:
          Off(OFF_BLAST);
          break;
  ```
  But having EFFECT_DESTRUCT allows for blade animations.

File: props/blaster_BC_buttons.h (hunk 2 above)
Comment (NoSloppy) [reply props/blaster_BC_buttons.h:184]: I am unclear on this. Since only one blaster prop file is included at a time, the other isn't even compiled, no?
Isn't it code substitution as opposed to code duplication?
I can imagine if there were a blaster_base.h maybe...

File: props/blaster_BC_buttons.h (hunk 3 above)
Comment (NoSloppy) [reply props/blaster_BC_buttons.h:520]: I put it above Event2 where I saw it in other props, and
I just added more bools to the spot that blaster.h has them.
So should the RefPtr (and all the other declarations) be moved to the bottom of the file?

Comment (profezzorn) [reply props/blaster_BC_buttons.h:520]: Yes.

File: props/blaster_BC_buttons.h (hunk 2 above)
Comment (NoSloppy) [reply props/blaster_BC_buttons.h:184]: Well, here's a reverse diff reporting what's the same between blaster.h and my prop. 271 same lines.
Is this good/worth making a blaster_base for?
[blaster_props_shared_code.txt](https://github.com/profezzorn/ProffieOS/files/12334977/blaster_props_shared_code.txt)

Comment (profezzorn) [reply props/blaster_BC_buttons.h:184]: yes, absolutely.

**Comment by NoSloppy** [comment]
> > Needs fixes for dedicated power and reload buttons, esp if dual_prop used.
> > Guidance?
> 
> It would be easier to provide guidance if I knew what the problem was...

dual_prop doesn't account for these buttons I think was / is the issue.

File: props/blaster_BC_buttons.h (hunk 3 above)
Comment (NoSloppy) [reply props/blaster_BC_buttons.h:520]: Working on it.

File: props/blaster_BC_buttons.h (hunk 2 above)
Comment (NoSloppy) [reply props/blaster_BC_buttons.h:184]: Well, i'm modifying it now, assuming we're leaving it named blaster.h, not making a new file blaster_base.h, right?

Comment (NoSloppy) [reply props/blaster_BC_buttons.h]: ok. inheritance complete. How about now?

File: props/blaster_BC_buttons.h (hunk 3 above)
Comment (NoSloppy) [reply props/blaster_BC_buttons.h]: done

File: props/blaster_BC_buttons.h [hunk 4]
```diff
@@ -649,7 +649,7 @@ void Loop() override {
   }
 
   // Blaster effects, auto fire is handled by begin/end lockup
-  void SB_Effect(EffectType effect, float location) override {
+  void SB_Effect2(EffectType effect, float location) {
```
profezzorn, line 652: Still needs "override"

NoSloppy: oh, I see now what the issue was. It wasn't compiling, said it was marked override but didn't override. Seemed odd since it's overriding already in the Blaster class and we're inheriting from there. But I found it worked as SB_Effect2, so thought that was the right move.
The actual issue was the addition of EffectLocation since this was last worked on. changing float to EffectLocation makes things right now.
If I understand correctly, it only "needs" to be SB_Effect2 if we were trying to get sound lengths for example.

profezzorn: Correct.
Generally SB_Effect() is the right place to play sounds, and SB_Effect2() is the right place to do something after that.

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,720 @@
...
+
+class BlasterBCButtons : public Blaster {
+public:
+  BlasterBCButtons() : Blaster() {}
+  const char* name() override { return "BlasterBCButtons"; }
+
+  void NextBlasterMode() override {
+    switch(blaster_mode) {
+      case MODE_STUN:
+        stun_mode_ = false;
```
profezzorn, line 185: stun_mode_ seems redundant, why not just check if blaster_mode == MODE_STUN ?

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,720 @@
...
+
+
+  void selfDestruct() override {
+        SaberBase::DoEffect(EFFECT_DESTRUCT, 0);
+    float len = hybrid_font.GetCurrentEffectLength();
+    SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+      SetNextActionF(NEXT_ACTION_BLOW, len);
+  }
+// Volume Menu
+    void VolumeMenu() {
```
profezzorn, line 332: Use a mode?

common/saber_base.h
profezzorn: We might want to add a way for TrInOut to handle EFFECT_DESTRUCT so we can have a transition from the on to the off color.

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,720 @@
...
+#endif
+        return;
+      case MODE_AUTO:
+        stun_mode_ = true;
+        SetBlasterMode(MODE_STUN);
+        return;
+    }
+  }
+
+  bool IsEmpty() const {
```
profezzorn, line 204: This should probably be added to blaster.h

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,720 @@
...
+    return max_shots_ != -1 && shots_fired_ >= max_shots_;
+  }
+
+void Fire() {
```
profezzorn, line 208: Instead of overrinding the whole Fire function, how about splitting Blaster::Fire into smaller functions and then overriding the ones that are actually different from the base.

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,720 @@
...
+    // This will mimic the hammer running on no bullets at the same rate it was firing.
+    // if there are different length auto.wavs, the last played's length is used.
+    if (!auto_firing_ && IsEmpty() && trigger_is_pressed_ && (millis() - auto_time_) > auto_sound_length) {
+      SaberBase::DoEffect(EFFECT_EMPTY, 0);
+      auto_time_ = millis(); // Reset for the next empty sound
+    }
+  }
+}
+
+  void Reload() override {
```
profezzorn, line 285: The only difference seems to be that you set is_jammed_ to false.
Not sure if we should just do that in Blaster.
If not, the function should probably call the base, like:

void Reload() override {
   Blaster::Reload();
   is_jammed_ = false;
}

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,720 @@
...
+    shots_fired_ = 0;
+    is_jammed_ = false;
+    no_clip_ = false;
+  }
+
+  // Pull in parent's SetPreset, but turn the blaster on.
+  void SetPreset(int preset_num, bool announce) override {
+    PropBase::SetPreset(preset_num, announce);
+      // Just always default to on.
+       if (!SaberBase::IsOn()) On();
```
profezzorn, line 308: Not sure why always doing this is better than doing it *if* there is an SFX_poweron?

profezzorn (commented):
> dual_prop doesn't account for these buttons I think was / is the issue.

I guess the question is: what do we want it to do?

common/saber_base.h
NoSloppy: Hmm. oh, you mean like an "explosion". Wouldn't that just be the OUT->IN transition, something like
```cpp
InOutTrL<TrWipe<300>,TrConcat<TrInstant,Blinking<White,Black,75,500>,TrFade<600>>>
```
That would play with boom.wav.
I'm not sure what you're saying the integration with EFFECT_DESTRUCT should be.

props/blaster_BC_buttons.h
NoSloppy: Yes, now I need to update this ancient PR.

props/blaster_BC_buttons.h
NoSloppy: I guess I thought it was weird to use the Plecter Monophonic Filename for saber ignition as a basis for whether the blaster has a power on ability? Unless it's supposed to be "advertised" as a supported blaster font sound?
https://pod.hubbe.net/sound/sound-font-configuration.html

NoSloppy:
> > > Needs fixes for dedicated power and reload buttons, esp if dual_prop used.
> > > Guidance?
> > 
> > 
> > It would be easier to provide guidance if I knew what the problem was...
> 
> dual_prop doesn't account for these buttons I think was / is the issue. How would that work if we already have a "power" button that gets swapped to a button named BUTTON_FIRE? Would we just add the 2 additional non-remapped buttons something mostly like this?
> 
> ```
> #ifdef CONFIG_BUTTONS
> Button PowerButton(BUTTON_POWER, powerButtonPin, "pow"); 
> Button AuxButton(BUTTON_AUX, auxPin, "aux");
> Button ReloadButton(BUTTON_RELOAD, aux2Pin, "reload");
> Button ????Button (BUTTON_???? , 17, "???"); // using data3 on v2.2 or TX on v3.9 Proffieboards
> #endif
> ```

I think I'll just make AUX2 assignable as either RELOAD or POWER and hope no one needs a 4 button blaster.

props/blaster_BC_buttons.h
NoSloppy: I think this is to play dedicated stun reload and clip in/out sounds. It's used as a condition in SB_Effect.

NoSloppy: removed, used blaster_mode == MODE_STUN instead in the conditional checks.

props/blaster_BC_buttons.h
NoSloppy: done

props/blaster_BC_buttons.h
NoSloppy: done

props/blaster_BC_buttons.h
NoSloppy: done

props/blaster_BC_buttons.h
NoSloppy: done

props/blaster_BC_buttons.h
NoSloppy: done

props/blaster.h
```diff
@@ -87,6 +87,8 @@ mdauto		Sound made when switching to AUTO mode
 #ifndef PROPS_BLASTER_H
 #define PROPS_BLASTER_H
 
+#define PROP_HAS_GETBLASTERMODE
```
NoSloppy, line 90: I don't remember what this was for. OLED maybe? Also not sure why I have it in my local copy but it's not in the master.

NoSloppy:
Still making changes as I keep testing and finding things. Getting closer.

common/saber_base.h
NoSloppy: oh well that's not right. It works for Destruct, but also shows if powering off normally.
So what did you have in mind?
I glanced at styles/inout_helper.h, but do not know what should happen here.
Maybe if EFFECT_DESTRUCT happens, we somehow bypass `in_tr_.begin();` ?

props/blaster_BC_buttons.h
NoSloppy: Ok so the reason this was confusing is because it was explained backwards in the instruction?
According to the actual code, the blaster will not turn ON by default if there is a poweron.wav
```cpp
  // Pull in parent's SetPreset, but turn the blaster on.
  void SetPreset(int preset_num, bool announce) override {
    PropBase::SetPreset(preset_num, announce);

    if (!SFX_poweron) {
      if (!SaberBase::IsOn()) {
        On();
      }
    }
  }
```
Which now makes sense, as in you would want to use your POWER button to control it.
I've incorporated it.

props/blaster.h
```diff
@@ -484,7 +499,16 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
       break;
     }
   }
-};
 
+#ifdef BLASTER_SHOTS_UNTIL_EMPTY
+  const int max_shots_ = BLASTER_SHOTS_UNTIL_EMPTY;
```
profezzorn, line 504: Constants don't usually go at the bottom with the variables.

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,788 @@
...
+private:
+  bool no_clip_ = false;
+  bool is_speaking_ = false; // Prevent overlap
+  bool sequential_quote_ = false;
+  bool trigger_is_pressed_ = false;
+  int auto_sound_length = 0;
+  uint32_t auto_time_;
+  RefPtr<BufferedWavPlayer> auto_player_;
+  RefPtr<BufferedWavPlayer> wav_player;
+
```
profezzorn, line 784: Extra empty line.

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,788 @@
...
+struct BCBlasterMenuSpec {
+  typedef BCBlasterVolumeMode<SPEC> BCBlasterVolumeMenu;
+  typedef mode::SelectCancelMode SelectCancelMode;
+  typedef mode::SteppedMode<SPEC> SteppedMode;
+  typedef mode::SteppedModeBase<SPEC> SteppedModeBase;
+  typedef mode::MenuBase<SPEC> MenuBase;
+  typedef SoundLibraryV2 SoundLibrary;
+};
+
+  void Setup() {
```
profezzorn, line 340: Why is this global, and indented?

props/blaster.h
profezzorn: It's used here:
https://github.com/profezzorn/ProffieOS/blob/b88b82da47b2c5b02691bc216c150edf2613a53b/ProffieOS.ino#L695
which is in turn used by BlasterModeF

profezzorn:
> > > > Needs fixes for dedicated power and reload buttons, esp if dual_prop used.
> > > > Guidance?
> > > 
> > > 
> > > It would be easier to provide guidance if I knew what the problem was...
> > 
> > 
> > dual_prop doesn't account for these buttons I think was / is the issue. How would that work if we already have a "power" button that gets swapped to a button named BUTTON_FIRE? Would we just add the 2 additional non-remapped buttons something mostly like this?
> > ```
> > #ifdef CONFIG_BUTTONS
> > Button PowerButton(BUTTON_POWER, powerButtonPin, "pow"); 
> > Button AuxButton(BUTTON_AUX, auxPin, "aux");
> > Button ReloadButton(BUTTON_RELOAD, aux2Pin, "reload");
> > Button ????Button (BUTTON_???? , 17, "???"); // using data3 on v2.2 or TX on v3.9 Proffieboards
> > #endif
> > ```
> 
> Changes to blaster.h make this easier now I think. Working on it.
> 
> I'll just make the RELOAD button (AUX2) have the option to be dedicated POWER and hope no one brings a 4 button blaster :)

I think we would need to add some code to dual_prop to handle these kind of scenarios.

props/blaster.h
NoSloppy: ah yes I remember that. Cool. I guess it just never made it into the blaster prop?

props/blaster_BC_buttons.h
NoSloppy: Do you mean it's global, so fix the indent, or is it wrong to be global?

props/blaster_BC_buttons.h
NoSloppy: fixed.

props/blaster_BC_buttons.h
profezzorn: I think it should be inside the prop. (and I think you want an "override" on it)

props/blaster.h
NoSloppy: moved.

props/blaster_BC_buttons.h
NoSloppy: Done, thanks. Not too aware of the startup routine yet.

common/saber_base.h
NoSloppy: Or maybe we just have an EFFECT_BOOM that layers in the blade style on top of the InOutTrL ?

NoSloppy: Added EFFECT_BOOM, tested, works well for Boom, unless you had a different idea?

props/blaster.h
```diff
@@ -159,59 +158,81 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
   }
 
   virtual bool CheckJam(int percent) {
     int random = rand() % 100;
     return random < percent;
   }
 
-  virtual void Fire() {
+  bool DoJam() {
 #ifdef ENABLE_MOTION
```
profezzorn, line 185: #if defined(ENABLE_MOTION) && defined(BLASTER_JAM_PERCENTAGE)

props/blaster.h
```diff
@@ -159,59 +158,81 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
 #endif
+  }
 
-    if (max_shots_ != -1) {
-      if (shots_fired_ >= max_shots_) {
-        SaberBase::DoEffect(EFFECT_EMPTY, 0);
-        return;
-      }
-    }
+  virtual int GetBulletCount() {
```
profezzorn, line 200: Seems the same as the inherited version, so we don't need it, right?

props/blaster.h
profezzorn: it doesn't seem moved...
(It should be near the top of the class.)

NoSloppy:
I thought I moved that because I remember saying to myself "ok I'll just put it back where it was." 
Now you're saying top, so I just put it there now, up above the first function.

props/blaster.h
NoSloppy: I don't see anything overriding this. 
Did you mean we don't need the virtual?
I think that's correct.

props/blaster.h
NoSloppy: Yup.
Done.

props/blaster.h
NoSloppy: > it doesn't seem moved... (It should be near the top of the class.)

I thought I moved that because I remember saying to myself "ok I'll just put it back where it was." 
Now you're saying top, so I just put it there now, up above the first function.

*EDIT* - wait. I *did* move them? I thought so. Irrelevant now, just weird that they were on the bottom again just 20 minutes ago.
https://github.com/profezzorn/ProffieOS/pull/582/commits/73c533cbded12b5f74aad403242fdbd547486023

props/blaster.h
profezzorn: Sorry, I missed what file this was in.
It shows up as new because the diff is dumb. Maybe just move it up to line 172 to make the diff nicer.

NoSloppy: outdated

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,787 @@
...
+  }
+
+ void DoAutoFire() override {
+    if (!SFX_auto) return;
+    // First time through, let bgnauto finish first if we have one
+    if (SFX_bgnauto && GetWavPlayerPlaying(&SFX_bgnauto)) return;
+
+    // No bgnauto or bgnauto has finished playing, so initialize and start auto-firing
+    if (auto_firing_ && !auto_player_) {
+      auto_player_ = GetWavPlayerPlaying(&SFX_auto);
```
profezzorn, line 388: Should SFX_auto "follow" SFX_bgnauto?

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,787 @@
...
+        auto_firing_ = false;
+        auto_player_.Free();
+        auto_time_ = millis(); // Reset for the first empty sound
+        return;
+      }
+      auto_time_ = millis();
+      PVLOG_NORMAL << "******** AUTOFIRING - Remaining shots = " << GetBulletCount() << "\n";
+    }
+
+    // This will mimic the hammer running on no bullets at the same rate it was firing.
```
profezzorn, line 409: This is cool, but only guns with electric ignition would do this right?
Most auto-firing guns would only do this once.
Maybe a define?

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,787 @@
...
+  void ClipIn() override {
+    Blaster::ClipIn();
+    is_jammed_ = false;
+    no_clip_ = false;
+  }
+
+  // Pull in parent's SetPreset, but turn the blaster on.
+  void SetPreset(int preset_num, bool announce) override {
+    PropBase::SetPreset(preset_num, announce);
+     if (!SFX_poweron && !SaberBase::IsOn()) {
```
profezzorn, line 447: indentation

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,787 @@
...
+  }
+
+  void ClipIn() override {
+    Blaster::ClipIn();
+    is_jammed_ = false;
+    no_clip_ = false;
+  }
+
+  // Pull in parent's SetPreset, but turn the blaster on.
+  void SetPreset(int preset_num, bool announce) override {
```
profezzorn, line 445: This seems to exactly the same thing as Blaster::SetPreset(), is there a reason for this override?

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,787 @@
...
+  }
+
+  void selfDestruct() override {
+    Blaster::selfDestruct();
+    SaberBase::DoEffect(EFFECT_DESTRUCT, 0);
+    SetNextActionF(NEXT_ACTION_BLOW, hybrid_font.GetCurrentEffectLength());
+  }
+
+  // Clash to unjam or Enter/Exit Volume Menu.
+  void Clash(bool stab, float strength) override {
```
profezzorn, line 459: same as override?

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,787 @@
...
+
+  void DoQuote() {
+    if (SFX_quote) {
+      if (GetWavPlayerPlaying(&SFX_quote)) return;  // Simple prevention of quote overlap
+      sequential_quote_ ? SFX_quote.SelectNext() : SFX_quote.Select(-1);
+      hybrid_font.PlayCommon(&SFX_quote);
+    }
+  }
+
+  void ToggleSequentialQuote() {
```
profezzorn, line 516: Seems like a settings menu thing (optional)

props/blaster.h
```diff
@@ -129,7 +128,13 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
   };
 
   BlasterMode blaster_mode = BLASTER_DEFAULT_MODE;
-	
+
```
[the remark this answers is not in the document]
NoSloppy: The comments at line 32-34 have shifted everything down one line. So this blank line has not moved alone. The change is that on my end, the original has a dash(looks like a tab). There are more on line 401 just before #ifdef BLADE_DETECT_PIN, as well as a ton in the comments (see image)
So this change just removes the whitespace from the blank line.
Screenshot 2025-02-17 at 6 40 15 PM

props/blaster.h
```diff
@@ -159,59 +164,73 @@ class Blaster : public PROP_INHERIT_PREFIX PropBase {
...
     is_jammed_ = is_jammed_ ? true : CheckJam(BLASTER_JAM_PERCENTAGE);
 
     if (is_jammed_) {
       SaberBase::DoEffect(EFFECT_JAM, 0);
-      return;
+      return true;
+    } else {
+      return false;
     }
 #endif
```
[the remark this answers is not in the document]
NoSloppy: Didn't give an error, but I added a return false to make sure something is returned.

sound/hybrid_font.h
```diff
@@ -526,13 +526,15 @@ class HybridFont : public SaberBase, public Looper {
...
       case OFF_BLAST:
+        SaberBase::DoEffect(EFFECT_BOOM, 0);
         if (monophonic_hum_) {
           if (SFX_boom) PlayMonophonic(getNext(hum_player_, &SFX_boom), NULL);
           else PlayMonophonic(getNext(hum_player_, &SFX_clash), NULL);  // Thermal-D fallback
         } else {
           state_ = STATE_HUM_FADE_OUT;
           PlayPolyphonic(getNext(lock_player_, &SFX_boom));
         }
+        PVLOG_NORMAL << "+++++ BOOM!! +++++\n";
```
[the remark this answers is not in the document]
NoSloppy: done.

sound/hybrid_font.h
```diff
@@ -526,13 +526,15 @@ class HybridFont : public SaberBase, public Looper {
 	}
         break;
       case OFF_BLAST:
+        SaberBase::DoEffect(EFFECT_BOOM, 0);
```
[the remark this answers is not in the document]
NoSloppy: ah ok, done. tested works well.

props/blaster_BC_buttons.h
NoSloppy: Well, If it's up to me because it's the BC prop, I'd like to leave it like this. I like it and have pride in the auto-timing thing, plus I feel like more sounds playing the better instead of one "click - you're empty".
Also, it's a long time ago in a galaxy far, far away. Who knows how blasters actually worked? :)

props/blaster_BC_buttons.h
NoSloppy: fixed.

props/blaster_BC_buttons.h
NoSloppy: Ah. It wasn't using the poweron sound as a determining factor before, but somebody talked me into it lately.
So now, yes it's the same.
Removed from here, and edited the base file's version to consolidate the conditions to one neater line.

props/blaster_BC_buttons.h
NoSloppy: Yeah, I guess in the base file's updating, they ended up the same now.
Removed from here. Thanks for the catch.

props/blaster_BC_buttons.h
NoSloppy: Maybe if there were more actions and/or buttons involved it could be, sure. But if it's ok as a simple toggle here like this, I'll choose the option to leave it.

props/blaster_BC_buttons.h
NoSloppy: Hmm.. Well it looks like SB_BeginLockup() currently handles setting all the bgn sounds to their lockup sounds.
I presume in your suggestion that you mean to utilize EFFECT2, which I assume would work the same? But then are we adjusting all of SB_BeginLockup() as well? I;m not sure how much refactoring you want to do, if it's needed, or just a better way.

Also, this opens another question for auto-fire. Should we add a check that auto sounds exist even though the user may have defined auto mode? Then if no auto exists, should we use a blast sound instead? Or throw an error and go right to the next blaster mode, something like
```cpp
  virtual void NextBlasterMode() {
............
      case MODE_KILL:
#if defined (ENABLE_BLASTER_AUTO) || defined (BLASTER_ENABLE_AUTO)
        if (SFX_auto) {
          SetBlasterMode(MODE_AUTO);
        } else {
          ProffieOSErrors::error_in_font_directory();
          SetBlasterMode(MODE_STUN);
        }
#else
        SetBlasterMode(MODE_STUN);
#endif
        return;
      case MODE_AUTO:
        SetBlasterMode(MODE_STUN);
        return;
    }
  }
  ```

File: props/blaster.h (hunk 23 above)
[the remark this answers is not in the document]
Comment (NoSloppy) [reply props/blaster.h:131]: I cleaned them up in #740

File: props/blaster_BC_buttons.h [hunk 27]
```diff
@@ -0,0 +1,767 @@
...
+#endif
+
+#define PROP_TYPE BlasterBCButtons
+
+#ifndef BUTTON_HELD_LONG_TIMEOUT
+#define BUTTON_HELD_LONG_TIMEOUT 1200
+#endif
+
+// Additional effects for BC buttons prop.
+EFFECT(clipins);    // s for stun mode
```
profezzorn, line 196: Seems like it would be better to use alt modes for stun/kill/auto, that way *every* effect can be individualized if you want to, without adding any more code.

props/blaster_BC_buttons.h
profezzorn: I think checking SFX_auto exists is optional, however, it would be kind of nice, because then you could possibly get rid of the define and have enable/disable auto entirely based on what files exists.

props/blaster_BC_buttons.h
NoSloppy: Sorry, I guess I don't know what you mean by every effect could be individualized.
Unless you mean to have 3 sets of clipin, clipout, and reload in the font, then call the appropriate alt for each mode kill, stun, auto?
Again, would that be in the prop? I thought alt sounds were "meant" to be changed via blade style code.

profezzorn: https://pod.hubbe.net/sound/alt_sounds.html#have-the-prop-control-the-alt

props/blaster_BC_buttons.h
profezzorn: If you're going to require SFX_auto, the earlier you can give the error/warning the better.
(Like in SetPreset maybe?)
Also, you would need to print something to let people know what the actual problem is.

Since BeginLockup already handles that part, why mess with it?

NoSloppy: ok maybe I'm reading things wrong then. Let me regroup.

1. You mentioned "Should SFX_auto "follow" SFX_bgnauto?" Well, I see they are set as the play once and play looped sounds in SB_BeginLockup already, but I wrote this "wait for bgnauto to finish" code here for a reason, but I can't remember why. I assume the sounds were overlapping?
So, now thinking, I should be able to just add the "following" feature via EFFECT2, and remove the waiting code here. 
Does that sound right?

2. When you said " possibly get rid of the define and have enable/disable auto entirely based on what files exists", That sounds good for in my prop. The base blaster prop still uses the define, and that's not up to me to keep or not.
But SetPreset() can still check for the sound, like:
```cpp
#if defined (ENABLE_BLASTER_AUTO) || defined (BLASTER_ENABLE_AUTO)
    if (!SFX_auto) {
      ProffieOSErrors::error_in_font_directory();
      PVLOG_NORMAL << "** ERROR - BLASTER_ENABLE_AUTO active, but no auto.wav files exist!\n";
    }
#endif
```
and then the mode can skip right to stun as in the code above. sound ok?

profezzorn: I think you should implement the auto-skip or the "error in font directory", not both.
For the "following" thing, I think it's probably best to just leave it alone for now.

NoSloppy: > I think you should implement the auto-skip or the "error in font directory", not both. 

Copy Gold Leader.

NoSloppy: > 1. I wrote this "wait for bgnauto to finish" code here for a reason, but I can't remember why. I assume the sounds were overlapping?

So I see now why this was here.
The autofiring starts while bgnauto is playing, and since no auto sound plays until bgnauto finishes, there's no wav length to get, so it just blew through all the ammo and was empty by the end of bgnauto.wav.
I simplified it.
Regarding "autofollowing", it already does because it's a lockup type.
I think we're good now.

props/blaster_BC_buttons.h
NoSloppy: > Seems like it would be better to use alt modes for stun/kill/auto, that way _every_ effect can be individualized if you want to, without adding any more code.

Done.

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,801 @@
...
+      beeper.Beep(0.1, 300);
+    }
+    Off();
+  }
+
+  virtual void NextBlasterMode() override {
+    switch(blaster_mode) {
+      case MODE_STUN:
+        SetBlasterMode(MODE_KILL);
+        SaberBase::DoEffect(EFFECT_ALT_SOUND, 0.0, 0);
```
profezzorn, line 545: Why not put this in SetBlasterMode?

props/blaster_BC_buttons.h
```diff
@@ -0,0 +1,801 @@
...
+  // Blaster effects, auto fire is handled by begin/end lockup
+  void SB_Effect(EffectType effect, EffectLocation location) override {
+    Blaster::SB_Effect(effect, location);
+    switch (effect) {
+      default: return;
+      case EFFECT_STUN: hybrid_font.PlayCommon(&SFX_stun); return;
+      case EFFECT_FIRE: hybrid_font.PlayCommon(&SFX_blast); return;
+      case EFFECT_CLIP_IN: 
+        if (blaster_mode == MODE_STUN && SFX_clipins) {
+          hybrid_font.PlayCommon(&SFX_clipins); return;
```
profezzorn, line 743: Now that we have ALT modes, do you need clipins?

NoSloppy: I deleted this and bumped undo accidentally. the whole function will just call the base version, then add in only the custom cases.
Also, I need to have alt sounds choice not reset to KILL when doing SetPreset() for some reason, even though everything else is still in the correct mode.
... I'm working on it as we speak.

NoSloppy: Ok I figured out what I was doing wrong, plus greatly simplified by using SetBlasterMode instead. Thanks.
Question. If a derived class calls the base class’ version of a virtual function directly (like Blaster::SetBlasterMode for example), is there any need to override it explicitly? Or is that just cleaner coding.

profezzorn: Looks like current_alternative is currently being reset in Effect::ScanCurrentDirectory().
We could probably lift that out and do it in the prop, and then make it overridable.

props/blaster_BC_buttons.h
NoSloppy: Yes. Way way better, thank you.

props/blaster_BC_buttons.h
profezzorn: Blaster::SetBlasterMode will always call SetBlasterMode in "Blaster", so making it virtual and/or overriding it will not make any difference.  Not sure if that was the actual question or not though.

NoSloppy: > Looks like current_alternative is currently being reset in Effect::ScanCurrentDirectory(). We could probably lift that out and do it in the prop, and then make it overridable.

I see that. Well, I just set the alt to the current mode in SetPreset(), which I think makes sense for a blaster, but for sabers it might make more sense the way it works now, like the expectation when you start a new font is that the alts are the "default" alt000.

Moving `current_alternative = 0` out of ScanCurrentDirectory would change that for all props, unless we set it explicitly in all prop files, no?
Or is that what you just said :upside_down_face:

NoSloppy: > Blaster::SetBlasterMode will always call SetBlasterMode in "Blaster", so making it virtual and/or overriding it will not make any difference. Not sure if that was the actual question or not though.

Yes that's the question, to the degree that I see it would behave identically, but wanted to know if you prefer the override to conform to most of the rest of the code base.

profezzorn: > Moving current_alternative = 0 out of ScanCurrentDirectory would change that for all props, unless we set it explicitly in all prop files, no?

I was thinking that we would create a virtual function in PropBase for resetting the alt, then put a call to that function in chdir(), which is the only place ScanCurrentDirectory is called, then props can override that function.

> Yes that's the question, to the degree that I see it would behave identically, but wanted to know if you prefer the override to conform to most of the rest of the code base.

No, virtual functions costs a small amount of memory and cpu, so if they don't need to be virtual, they shouldn't be. Normally, we use virtual functions because the base class calls it, not the derived class.  However, if you think that there may be a class that derives from your prop in the future, then you may want to make something virtual to allow the new class to override it...

NoSloppy: ok. So What I'm hearing is that if a function is already virtual the "damage is done" to memory already, so whether or not a derived class function overrides it or not is not very pertinent to memory. 
Specifically to this case, since both my SetBasterMode() and SetPreset() functions directly call the parent virtual version, they could be non-overriding, they'd still behave the same, but it wouldn't matter much to resources use.
So might as well leave them overriding as most other similar things do (like SetPreset() in blaster.h for example.)

profezzorn: So I'm a bit confused.
Overriding a function uses additional memory, in addition to the memory used by making it virtual.
You shouldn't override SetPreset() unless you need to do something in addition to what Blaster::SetPreset() already does.

If you do override a function to make it do additional things, it then it should be virtual.
Technically if doesn't need to be virtual if the base class(es) never calls it, but since SetPreset is definitely called by the base class, it would have to be virtual.   (If it wasn't virtual, then calls to SetPreset would just go to the SetPreset in that file, not the overridden one.)

Now, if you want to add something to a function, but the base class *doesn't* call it, then we don't need to make it virtual, but we also don't need to make it the same name. In this case it would be better to make a BlasterSetPreset() or something, which then calls SetPreset().

NoSloppy: Sorry for the confusion. I think it's right, if not, you can point it out in a review of the code line itself.

I'll work on moving the current_alternative in a separate PR.

profezzorn:
It only took a year and a half....

profezzorn:
PS: I'll need another blurb for the 8.x post on the crucible...

PR #588, opened 2023-08-14
Blade id constant monitoring fixes

NoSloppy:
Needs a follow up indentation and whitespace cleanup.


props/prop_base.h
```diff
@@ -542,17 +542,32 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+  // Prevent BLADE ID: from spamming console with every BLADE_SCAN_MILLIS
+  bool blade_id_should_report_ = true;
+
   // Measure and return the blade identifier resistor.
   float id() {
     EnableBooster();
     BLADE_ID_CLASS_INTERNAL blade_id;
     float ret = blade_id.id();
+#ifdef BLADE_ID_SCAN_MILLIS
+    if (blade_id_should_report_) {
```
profezzorn, line 554: While this work, it is not the best way to change the behavior of a function.
We should do one of:

1: add an argument to id() that specifies whether to do it quietly or not
2: split id into two functions, one quiet one and one noisy one, then call the right function for the right job

Using a global variable to change how a function works should only be used when 1 and 2 doesn't work for some reason.

props/prop_base.h
```diff
@@ -542,17 +542,32 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
 
+  // Prevent BLADE ID: from spamming console with every BLADE_SCAN_MILLIS
+  bool blade_id_should_report_ = true;
+
   // Measure and return the blade identifier resistor.
   float id() {
     EnableBooster();
     BLADE_ID_CLASS_INTERNAL blade_id;
     float ret = blade_id.id();
+#ifdef BLADE_ID_SCAN_MILLIS
```
profezzorn, line 553: Nesting #Ifdefs and duplicating code is not very clean.

props/prop_base.h
```diff
@@ -607,7 +622,11 @@ class PropBase : CommandParser, Looper, protected SaberBase {
   void PollScanId() {
     if (find_blade_again_pending_) {
       find_blade_again_pending_ = false;
+      blade_id_should_report_= true;
       FindBladeAgain();
+      // Indicate what happened when different Blade ID detected
+      SaberBase::DoNewFont();
```
profezzorn, line 628: Why is this needed?
Doesn't FindBladeAgain() already do this? (or something similar?)

props/prop_base.h
```diff
@@ -607,7 +622,11 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+      STDOUT << "DISPLAY: " << current_preset_name() << "\n";
```
profezzorn, line 629: Why does this say DISPLAY?
Why is it needed?
Why isn't it a PVLOG?

props/prop_base.h
NoSloppy: FindBladeAgain() was just silently running.
This was the only indication something occurred:
```
22:56:52.786 -> blade = 0
22:56:52.786 -> WS2811 Blade with 11 leds.
22:56:52.786 -> WS2811 Blade with 5 leds.
22:56:52.786 -> WS2811 Blade with 11 leds.
22:56:52.786 -> Simple Blade
22:56:52.786 -> Simple Blade
22:56:52.786 -> Style RAM = 72
22:56:52.786 -> Style RAM = 72
22:56:52.786 -> Style RAM = 72
22:56:52.786 -> Style RAM = 72
22:56:52.786 -> Style RAM = 72
22:56:52.786 -> Scanning sound font: testfont2b
22:56:52.888 ->  done
22:56:52.888 -> Scanning sound font: common
22:56:52.888 ->  done
22:56:52.888 -> Scanning sound font: commonBU
22:56:52.961 ->  done
22:56:53.140 -> Activating polyphonic font.
22:56:53.242 -> Activating SmoothSwing V2
22:56:53.242 -> Accent Swings Enabled.
22:56:53.242 -> Polyphonic swings: 1
22:56:53.242 -> Monophonic swings: 0
22:56:53.242 -> Accent Slashes Enabled.
22:56:53.242 -> Polyphonic slashes: 1
```
so it made sense to play the font.wav and show the preset name, no?
Strange though, I think FindBladeAgain USED to do that though?

props/prop_base.h
NoSloppy: DISPLAY was there already (well not HERE , but I reused this line)
See above for why.
Should be PVLOG, yes. Again, i just reused the line.

props/prop_base.h
NoSloppy: Update - I think just calling SetPreset(0, true) does this instead. 
Edited.

props/prop_base.h
NoSloppy: ok, how about now? simpler approach.

props/prop_base.h
NoSloppy: gone.

props/prop_base.h
NoSloppy: updated.
SetPreset()'s  line made PVLOG

NoSloppy:
What I am NOT sure about is how much we need to backward support a NON-constant monitoring process.
Would someone prefer new blade scan only to happen at boot, like it used to be?

props/prop_base.h
```diff
@@ -563,6 +557,16 @@ class PropBase : CommandParser, Looper, protected SaberBase {
     return ret;
   }
 
+  void PrintAndSpeakId() {
```
profezzorn, line 560: Make this return current_id, then make FindBladeAgain() call this function, that way we don't have to call id() twice, and we don't need to pepper the code with PrintAndSpeakId() calls.

props/prop_base.h
```diff
@@ -547,12 +547,6 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
     float ret = blade_id.id();
-    PVLOG_STATUS << "BLADE ID: " << ret << "\n";
-#ifdef SPEAK_BLADE_ID
-    talkie.Say(spI);
-    talkie.Say(spD);
-    talkie.SayNumber((int)ret);
-#endif
 #ifdef BLADE_DETECT_PIN
     if (!blade_detected_) {
       STDOUT << "NO ";
```
profezzorn, line 552: This printout (and the one on 561) needs to be moved to PrintAndSpeakId() as well.

NoSloppy: ok done see update

props/prop_base.h
NoSloppy: the trick is that FindBladeAgain runs FindBestConfig, which needs to retrieve the id().
If that gets it from PrintAndSpeak() as current_id instead, then we;re back to the original issue where the console is spammed with printed (or worse, also spoken) BLADE ID: every loop.
So I'm not sure why we'd return current_id in PrintAndSpeakId(). Future flexibility?

How does it look now?

NoSloppy:
Also, I thought it would be good to use sound_library for the speaking, and check if DISABLE_TALKIE is active.
Adding sound_library not working off the bat. It has to do with SoundQueue.

props/prop_base.h
NoSloppy: oh wait. Just realized I'm hearing 2 boot.wavs now.
It's caused by the call to SetPreset() in FindBlade(). It was false (announce) before I made it true to play font.wav when Blade ID changed.
Why would SetPreset be playing boot.wav on top of actual boot?
I guess that's why it was false announce originally?
It seems like a weird thing that SetPreset would play boot.wav, and not just font.wav on preset change (or blade ID change as was the intent here).
Adding a delay "fixes" it where only one boot.wav plays, but I feel like that's a hack-patch.
```
    if (millis() < 3000) {
      SetPreset(0, false);
    } else {
      SetPreset(0, true);
    }
```

There's this from hybrid_font maybe why boot.wav is playing from blade ID routines, but nothing's calling BladeDetect in FindBladeAgain()...or is there?
```
  void SB_BladeDetect(bool detected) {
    Effect &X(detected ? SFX_bladein : SFX_bladeout);
    if (X) {
      PlayPolyphonic(&X);
      return;
    }
    if (detected && SFX_boot) {
      PlayPolyphonic(&SFX_boot);
      return;
    }
    beeper.Beep(0.05, 2000.0);
  }
  ```

File: props/prop_base.h [hunk 7]
```diff
@@ -443,7 +443,7 @@ class PropBase : CommandParser, Looper, protected SaberBase {
     chdir(current_preset_.font.get());
     if (on) On();
     if (announce) {
-      STDOUT << "DISPLAY: " << current_preset_name() << "\n";
+      PVLOG_STATUS << "DISPLAY: " << current_preset_name() << "\n";
```
profezzorn, line 446: Change "DISPLAY: " to "Current Preset: "

props/prop_base.h
```diff
@@ -543,32 +543,43 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
+    return current_id; // not sure why we're returning the ID value here
+  }
+
   size_t FindBestConfig() {
     static_assert(NELEM(blades) > 0, "blades array cannot be empty");
 
     size_t best_config = 0;
     if (NELEM(blades) > 1) {
-      float resistor = id();
+      float resistor = RetrieveBladeId();
```
profezzorn, line 582: I guess I didn't think this through.
FindBestConfig() needs a parameter (either "quietly" or "announce")
Then we can call id() (or whatever we want to call it) with that parameter, no need for two functions I guess.
(Sorry about the confusion)

props/prop_base.h
NoSloppy: done

props/prop_base.h
NoSloppy: Alright, how about now?
Calling id(true) will print (and speak) ID, else quiet.
FindBladeAgain() is announced along with the SetPreset routine being run. (except for first 3 seconds of boot to avoid 2 boot.wavs being played)
FindBestConfig() runs quietly otherwise.
"scanid" calls FindBladeAgain(), 
"id" calls id(true) directly, so just an announce.
If BLADE_ID_SCAN_MILLIS is not defined, "scanid" detects the new blade ID.

It's working.
Just that dual boot.wav thing seems it should be addressed.

props/prop_base.h
```diff
@@ -734,6 +741,7 @@ class PropBase : CommandParser, Looper, protected SaberBase {
 
     ONCEPERBLADE(DEACTIVATE);
     SaveVolumeIfNeeded();
+    FindBestConfig(true);
```
profezzorn, line 744: Is the only reason you need this for the printouts?
Is the only reason you need those printouts because of the change in FindBlade() ?

NoSloppy: Originally,FindBlade() called SetPreset(0, false), so there was no printout of current preset and no call to DoNewFont(). There was no indication something happened when a blade was inserted.
I thought playing the font.wav made sense to know something happened, and to indicate what preset was loaded, so I changed it to SetPreset(0, true). However, that caused 2 boot.wavs to play simultaneously. I don't know why SetPreset would play boot.wav. But as a workaround, I added the SetPreset(0, false) back in, but in the condition of only the first 3 seconds so a second boot.wav wasn't triggered.

The code currently works as I stated above, which was the intent. Is it bad?

> Calling id(true) will print (and speak) ID, else quiet.
> FindBladeAgain() is announced along with the SetPreset routine being run. (except for first 3 seconds of boot to avoid 2 boot.wavs being played)
> FindBestConfig() runs quietly otherwise.
> "scanid" calls FindBladeAgain(),
> "id" calls id(true) directly, so just an announce.
> If BLADE_ID_SCAN_MILLIS is not defined, "scanid" detects the new blade ID.

profezzorn: > Originally,FindBlade() called SetPreset(0, false), so there was no printout of current preset and no call to DoNewFont(). There was no indication something happened when a blade was inserted.
> I thought playing the font.wav made sense to know something happened, and to indicate what preset was loaded, so I changed it to SetPreset(0, true). However, that caused 2 boot.wavs to play simultaneously.

So far so good.

> I don't know why SetPreset would play boot.wav. But as a workaround, I added the SetPreset(0, false) back in, but in the condition of only the first 3 seconds so a second boot.wav wasn't triggered.

This is bad.
1. it's always easier to fix a problem you understand
2. functions should not do different things depending on what time it is, that makes them hard to use and unpredictable. In this particular case, it will lead to people who use a delay during boot to hear boot sounds twice.
3. Maybe it would be better to fix the *other* place that plays a boot sound?

So, good intentions, bad implementation.

NoSloppy: 🤦‍♂️
It's not 2 boots. It's boot.wav and font.wav. I had the same sound for both in a Dooku font.
Lesson learned, always use my testfont.
Well at least that makes more sense.
Ok, working on it.

NoSloppy: ok fixed.

props/prop_base.h
NoSloppy: Note - it wasn't 2 boot.wavs. It was font.wav over boot.wav because of the edit to SetPreset(0, true) all the time.
Resolved by having it call SetPreset with the announce variable instead of the delay hack-workaround.

NoSloppy:
I got it to do bladein and bladeout too, but currently it plays simultaneously and overlaps with font.wav.
Another reason sound_library would be good, to queue up the sounds.

sound/sound_library.h
```diff
@@ -9,39 +9,38 @@ enum SayType {
...
-  COLOR_ICEWHITE = 22,
-  COLOR_LIGHTCYAN = 23,   
-  COLOR_MOCCASIN = 24,
-  COLOR_LEMONCHIFFON = 25,
-  COLOR_NAVAJOWHITE = 26,
-  COLOR_WHITE = 27,
-};
-
+  // New colors should be added at end of enum and assigned numbers for each COLOR_ should not be changed.
+  enum ColorNumber {
```
profezzorn, line 15: What is this doing in this PR?
This is never going to get merged if you keep adding more and more to it.

NoSloppy: You're right, too much at once. It's part of sound library use for speak blade ID. Related, but not directly involved. should have waited.
if it's ok, i'll close submit this less haphazardly, now that i understand what's actually going on.(mostly)

NoSloppy:
Resubmitting cleaner

PR #589, opened 2023-08-14
Boot and version info

NoSloppy:
Too much?


profezzorn:
Yes, way way too much, and some of those things do not belong in "version" at all. (like current volume)

PR #591, opened 2023-08-19
Blade ID v2

NoSloppy:
This PR:
- Real Time Blade ID to work with dual prop and V3 boards.
- bladein and bladeout sounds work with Blade ID, but not play over boot.wav.

ToDo: 
- Replace Talkie with Sound library for SPEAK_BLADE_ID if feasible ?
-Update POD


blades/ws2811_blade.h
```diff
@@ -253,11 +253,16 @@ WS2811_Blade(WS2811PIN* pin,
       if (!current_style_ || !run_) {
 	loop_counter_.Reset();
 #ifdef BLADE_ID_SCAN_MILLIS
-	if (pin_->pin() == bladeIdentifyPin && ScanBladeIdNow()) {
-	  pin_->Enable(powered_);
-	  SLEEP(1);
-	}
-#endif      
+#if VERSION_MAJOR <= 5
+        if (pin_->pin() == bladeIdentifyPin && ScanBladeIdNow())
```
profezzorn, line 257: for now I think we just replace bladeIdentifyPin with bladePin for everybody,  no need for an #ifdef

props/dual_prop.h
```diff
@@ -112,24 +120,24 @@ class SaberBlasterProp : public virtual Saber, public virtual Blaster {
...
-	Saber::Off();
-	Saber::blade_detected_ = false;
-	Saber::FindBladeAgain();
-	SaberBase::DoBladeDetect(false);
-	current_modifiers = 0;
-	return true;
+        Saber::blade_detected_  = false;
+        SaberBase::DoBladeDetect(false);
       }
+    Saber::Off();
```
profezzorn, line 133: Maybe we should just set the saber to off quietly instead of calling Off() ?

props/dual_prop.h
```diff
@@ -112,24 +120,24 @@ class SaberBlasterProp : public virtual Saber, public virtual Blaster {
...
+    Saber::FindBladeAgain();
```
profezzorn, line 134: Why are checking the blade id() here?

props/dual_prop.h
```diff
@@ -112,24 +120,24 @@ class SaberBlasterProp : public virtual Saber, public virtual Blaster {
...
+    current_modifiers = 0;
```
profezzorn, line 135: Why are we dropping the modifiers here?

profezzorn:
> Replace Talkie with Sound library for SPEAK_BLADE_ID.

That's not going to fly if it makes all props require the sound library.

props/dual_prop.h
```diff
@@ -3,31 +3,39 @@
...
-to the file name of the prop you choose.
-
-Then setup your config file for 2 states:
-Bladein preset bank, and a BladeOut preset bank.
-Give the BladeConfig 2 sets of descriptions, and
-#define BLADE_DETECT_PIN properly.
+** Note the prop file names (saber_sa22c_buttons.h and blaster.h)
+as well as the class names would change based on the prop you choose.
+
+Now setup your CONFIG_PRESETS section to have multiple preset banks.
```
profezzorn, line 28: I think we should write this as two options: either use blade detect or blade id, then describe how to set up each option.

props/dual_prop.h
```diff
@@ -112,24 +120,24 @@ class SaberBlasterProp : public virtual Saber, public virtual Blaster {
...
+        SaberBase::DoBladeDetect(true);
       } else  if (event == EVENT_LATCH_OFF) {
-	Saber::Off();
-	Saber::blade_detected_ = false;
-	Saber::FindBladeAgain();
-	SaberBase::DoBladeDetect(false);
-	current_modifiers = 0;
-	return true;
+        Saber::blade_detected_  = false;
+        SaberBase::DoBladeDetect(false);
```
profezzorn, line 131: I think DoBladeDetect need to be done after FindBladeAgain()

blades/ws2811_blade.h
NoSloppy: ok this part is done.

NoSloppy:
> > Replace Talkie with Sound library for SPEAK_BLADE_ID.
> 
> That's not going to fly if it makes all props require the sound library.

What do you think we should do? What if the user defines DISABLE_TALKIE?
Wouldn't it work for all props if it's in prop_base since it's ....the base? 
I am slowly gaining more understanding on how things inherit, so forgive my naivete.
I suppose we could just do `#error You cannot define both DISABLE_TALKIE and SPEAK_BLADE_ID`
but it would be cool to utilize the sound_library, no?

props/dual_prop.h
NoSloppy: indeed.

props/dual_prop.h
NoSloppy: done.
The POD will need updating next.

NoSloppy:
Bump for attention.

profezzorn:
I don't think there is much reason to use the sound library for blade ID, since that's generally only used for debugging and initial configuration.

NoSloppy:
> I don't think there is much reason to use the sound library for blade ID, since that's generally only used for debugging and initial configuration.

Agreed, but I think this was to avoid a conflict if `#define DISABLE_TALKIE` is enabled.

profezzorn:
> Agreed, but I think this was to avoid a conflict if #define DISABLE_TALKIE is enabled.

Which would be totally reasonable if it didn't cost anything when not in use.

props/prop_base.h
```diff
@@ -631,8 +637,18 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
 #else
-    SetPreset(0, false);
-#endif
+    SetPreset(0, false); // DoNewFont or not? set to announce if yes desired.
+    // Have Blade ID play bladein and bladeout sounds.
+#endif // SAVE_PRESET
+#ifndef BLADE_DETECT_PIN
+    PVLOG_DEBUG << "************** Blade status = " << (bladestatus ? "IN" : "OUT") << "\n";
+    if (system_booted) {
+      SaberBase::DoBladeDetect(bladestatus);
```
profezzorn, line 646: I think we can simplify this by putting this line in PollScanId(), right after FindBlade

props/prop_base.h
```diff
@@ -612,14 +617,15 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
 
   // Called from setup to identify the blade and select the right
   // Blade driver, style and sound font.
-  void FindBlade() {
-    size_t best_config = FindBestConfig();
+  void FindBlade(bool announce = false) {
+    size_t best_config = FindBestConfig(announce);
     PVLOG_STATUS << "blade = " << best_config << "\n";
     current_config = blades + best_config;
+    bool bladestatus = (announce && best_config > 0);
```
profezzorn, line 628: I think this should be bladestatus = current_config->id >= NO_BLADE

NoSloppy: but in testing, blade out condition doesn't read as NO_BLADE value. Right now I'm getting ~750000. 1st blade I have here comes in at ~595000, and a 2nd blade shows 5500.
 Changing the code to your suggestion as `bool bladestatus = id(announce) >= NO_BLADE;` and making the no blade definition `NO_BLADE` causes it to always do bladeout sound, and the wrong choice is made as well.
 It's been a while since I was active with this code so i'm refreshing myself as we go back in now.
 
It's working as is, but are we trying to avoid requiring the "no blade" definition to be the first element in the blades array?

profezzorn: I've added a new define: NO_BLADE_ID_RANGE

Try #define NO_BLADE_ID_RANGE 700000,800000

This will make the blade ID class for that range return NO_BLADE

profezzorn: I think we should add a helper function:

bool blade_inserted() { return current_config->id >= NO_BLADE; }

We can then set DUAL_PROP_CONDITION to blade_inserted(), as that should work for both blade id and blade detect.

NoSloppy:
> > Agreed, but I think this was to avoid a conflict if #define DISABLE_TALKIE is enabled.
> 
> Which would be totally reasonable if it didn't cost anything when not in use.

how about ?
```
#ifdef SPEAK_BLADE_ID
#ifndef DISABLE_TALKIE
      talkie.Say(spI);
      talkie.Say(spD);
      talkie.SayNumber((int)ret);
#else
      if (&SFX_mnum) {
        sound_library_.SayNumber(ret, SAY_WHOLE);
      } else {
        beeper.Beep(0.05, 2000.0);
      }
#endif // DISABLE_TALKIE
#endif // SPEAK_BLADE_ID
```

profezzorn:
>         sound_library_.SayNumber(ret, SAY_WHOLE);

There is no sound_library_ variable in the base prop though.
I think the #error idea was better.

props/prop_base.h
NoSloppy: > Try #define NO_BLADE_ID_RANGE 700000,800000


Ok this is great.
I was getting random reports because apparently the readings dip down to ~689000 and were triggering a new read. Changing the range to 685000,1000000000 is rock solid.

Just an observation, BLADE ID: 1000000000.00 speaks as "zero" probably just because our numbers don't go that high.

NoSloppy: Oh. I meant to add that the >= needs to be <, as in
`    bool bladestatus = id(announce) < NO_BLADE; `

NoSloppy:
> There is no sound_library_ variable in the base prop though. I think the #error idea was better.

Would you be willing to have
```
#include "../sound/sound_library.h"
extern SoundLibrary sound_library_;
```
added right above `class PropBase `?

Otherwise, the error is working fine .

props/prop_base.h
```diff
@@ -1073,6 +1089,7 @@ class PropBase : CommandParser, Looper, protected SaberBase {
   volatile bool clash_pending1_ = false;
   volatile bool pending_clash_is_stab1_ = false;
   volatile float pending_clash_strength1_ = 0.0;
+  bool idle_off_time_reported_ = false;                            // BC IDLE_OFF_TIME expired printout only once mod
```
profezzorn, line 1092: whut?

props/prop_base.h
```diff
@@ -612,14 +617,15 @@ class PropBase : CommandParser, Looper, protected SaberBase {
...
 
   // Called from setup to identify the blade and select the right
   // Blade driver, style and sound font.
-  void FindBlade() {
-    size_t best_config = FindBestConfig();
+  void FindBlade(bool announce = false) {
+    size_t best_config = FindBestConfig(announce);
     PVLOG_STATUS << "blade = " << best_config << "\n";
     current_config = blades + best_config;
+    bool bladestatus = id() < NO_BLADE;
```
profezzorn, line 628: id() is expensive, so let's not call it again, current_config->id < NO_BLADE is better.

profezzorn: I updated prop_base to do this part, it will probably cause some conflicts with this change though.
Please let me know if you need help resolving.

PR #594, opened 2023-08-30
Bug fix for TWIST_RIGHT and LEFT preventing Drag and Melt

fdarosa2663:
I had to increase the Twist Right and Left angle during normal operation for Special Abilities, it made stab for Drag and Melt too hard to trigger as it was registering the Twist before the Stab for users.


fdarosa2663:
Just pinging, a few reports in OS7.10 of Melt and Drag not working when Twist Control is enabled (FETT263_SPECIAL_ABILITIES define).  Would like to get this and previous bug fix in my prop implemented for 7.11

PR #596, opened 2023-10-19
Bug fix and adjustment for Special Abilities

fdarosa2663:
Minor bug where Special Abilities would not trigger when wav_player was in use, conflicting with several effects.  Minor adjustment to Twist Right/Left sensitivity for users having difficulty triggering Special Abilities, will probably make it a define in future.


props/saber_fett263_buttons.h
```diff
@@ -6099,7 +6101,7 @@ SaberFett263Buttons() : PropBase() {}
 #ifdef FETT263_DUAL_MODE_SOUND
           SelectIgnitionSound();
 #endif
-	  wav_player.Free();
+	  if (!wav_player->isPlaying()) wav_player.Free();
```
profezzorn, line 6104: This can crash if wav_player is null.

Maybe it would be better to have a helper function, something like:

void FreeWavplayerIfPossible() {
   if (!wav_player) return;
   if (menu_) return;
   if (wav_player->isPlaying()) return;
   wav_player.Free();
}

This could then be called from loop() to release the wav player as needed.
It could also be called here, but I'm not sure if that's actually needed or not.

fdarosa2663: Thanks, I'll give this a try this weekend and update.

fdarosa2663: I replaced wav_player.Free() instances with FreeWavplayerIfPossible() in Loop(), works as expected no issues found.  Thanks.

PR #597, opened 2023-10-30
Update leds.h

ShtokyD:
Added templates for Bluetooth module and various colours 3mm leds.

### DIFF
--- a/blades/leds.h
+++ b/blades/leds.h
@@ -137,15 +137,71 @@ struct CreeXPLTemplate {
 
 using CreeXPL = CreeXPLTemplate<>;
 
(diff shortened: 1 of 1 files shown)


profezzorn:
1. I want part numbers and data sheets for the LEDs that you measured.
2. BT module should probably just use CH1LED

ShtokyD:
Those measurements are generic for those colours.

profezzorn:
Based on what though?
Did you measure it?
If so, how?
On what type of LEDs?

Tijgerd:
I agree with @profezzorn , there are so many LEDs available (also in 3mm sizes, they can have way different specs)
There are no real generic values

PR #605, opened 2023-12-09
fix DUAL_PROP_CONDITION

NoSloppy:
Avoid error: there are no arguments to 'blade_present' that depend on a template parameter, so a declaration of 'blade_present' must be available [-fpermissive]


profezzorn:
This change seem reasonable, but it tells me there is a conflict, not sure why...

NoSloppy:
> This change seem reasonable, but it tells me there is a conflict, not sure why...

Probably bad timing. Resolved now it seems.

PR #606, opened 2023-12-10
Prop base use sound lib

NoSloppy:
Open for discussion. It brings consistency to the UI, especially for Blade ID when FEMALE_TALKIE_VOICE is active.
As "I.D." is only available in male talkie... you wind up with 3 different voices speaking at you.

### DIFF
--- a/sound/sound_library.h
+++ b/sound/sound_library.h
@@ -115,20 +115,44 @@ public:
         if (hundredths != 0) Play(SoundToPlay(&SFX_mnum, hundredths - 1));
         break;
       case SAY_WHOLE:
+        if (number == 0) {
+          Play("mzero.wav");
+          break;
+        }
+#ifdef BLADE_ID_SCAN_MILLIS
+        // Billions spoken as "thousand million"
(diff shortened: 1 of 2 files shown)


sound/sound_library.h
```diff
@@ -115,20 +115,44 @@ class SoundLibrary {
...
+        }
+        // Millions
+        if (million > 0) {
+          TensValue(million);
+          Play("million.wav");
+        }
+#endif
         // Thousands
+        if (thousand >= 100) {
+          /* Support hundreds thousands
```
NoSloppy, line 137: I do not know where this version of Thousand came from, but it's in my local copy.
It works fine and I am only testing the official version as we speak.
Does it look like a good change? Any idea where this came from somewhere along the way?

profezzorn:
This feels like painting a hammer.
When I use a hammer, I care about the results, not how it looks. (or sounds)
SPEAK_BLADE_ID is a tool IMHO, not UI

NoSloppy:
ok!

PR #608, opened 2023-12-13
Update prop_base.h

NoSloppy:
In testing, this causes the board to play font.wav "for no reason". Although the board may have detected a change, if the result after averaging reads is the same state as before, it should just be quiet. Serial monitor still reports activity.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -628,8 +628,6 @@ public:
       bool blade_present_after = blade_present();
       if (blade_present_before != blade_present_after) {
         SaberBase::DoBladeDetect(blade_present_after);
-      } else {
-	SaberBase::DoNewFont();
       }
     }
   }


profezzorn:
If we get here and select the same font, then blade ID scanning isn't working properly.
Also, we will have torn down the styles and the blade structs, and then rebuilt them, which will cause glitches (restarts) on the blade. Removing this code may hide the problem, but it doesn't fix it.

NoSloppy:
I'm not trying to complicate things here. I think this is a rare case though where:
- SnapShot BladeID is used, so values are very close.
- find_blade_again_pending is coming into play and shifting the outcome by one read.
- I'm not sure what the "else" case should be or why we're Doing DoNewFont if nothing has changed.

Here's what I did. 
set `#define NO_BLADE_ID_RANGE 90,1000000000`

I have a no blade reading of ~91, but it dips to as low as 89 sometimes.
to keep an eye on what's going on, I added a printout in the scan averaging routine in blade_id.h like so
```
struct AverageBladeIDNTimes {
  float id() {
    float id = 0;
    BLADE_ID blade_id;
    for (int i = 0; i < N; i++) id += blade_id.id();
PVLOG_STATUS << "**** Scanned ID to average: " << id / N << "\n";
    return id / N;
  }
};
```
This gives me output where I can see the dip below 90 which causes font.wav to play, which looks like this:
15:43:14.789 -> **** Scanned ID to average: 91.84
15:43:15.782 -> **** Scanned ID to average: 91.72
15:43:16.791 -> **** Scanned ID to average: 91.72
15:43:17.780 -> **** Scanned ID to average: 89.68
15:43:17.822 -> **** Scanned ID to average: 96.80
15:43:17.822 -> BLADE ID: 1000000000.00
15:43:17.822 -> blade = 0
15:43:17.822 -> WS2811 Blade with 90 leds.
15:43:17.822 -> WS2811 Blade with 7 leds.
15:43:17.822 -> WS2811 Blade with 8 leds.
15:43:17.822 -> Style RAM = 1128
15:43:17.822 -> Style RAM = 44
15:43:17.822 -> Style RAM = 100
15:43:17.822 -> Scanning sound font: CodeFulcrumAhsoka
15:43:17.902 ->  done
15:43:17.902 -> Scanning sound font: common
15:43:17.987 ->  done
15:43:18.030 -> Activating polyphonic font.
15:43:18.030 -> Activating SmoothSwing V2
15:43:18.030 -> Accent Swings Enabled.
15:43:18.030 -> Polyphonic swings: 16
15:43:18.030 -> Monophonic swings: 0
15:43:18.030 -> Accent Slashes NOT Detected: 
15:43:18.073 -> unit = 0 vol = 0.50, Playing CodeFulcrumAhsoka/font/font.wav
15:43:18.073 -> channels: 1 rate: 44100 bits: 16
15:43:18.797 -> **** Scanned ID to average: 91.56
15:43:19.789 -> **** Scanned ID to average: 91.68
15:43:20.791 -> **** Scanned ID to average: 91.88

Notice how there's one additional reading after the 89.68 that falls back in the set range.
So the 89 triggers a difference of best config, so we get  find_blade_again_pending_ = true.
Then we check blade_present() but the additional read is back in the range, so we get the same blade, and therefore DONewFont...

It's like there's just one thing slightly off,  I can't tell what though.

If the check for if (blade_present_before != blade_present_after) comes back false, we would just move along, right?
Why the else DoNewFont if the blade_present() has no change?

Or maybe the the single out-of-range reading might need to be forgiven in the sea of correct reads?

Again, this is all rare occurrence due to poor blade resistance, BladeIDClass, range settings.
Should we just leave it and the answer is just setup better conditions?

or

profezzorn:
What is the highest reading you get with a blade inserted?

NoSloppy:
> What is the highest reading you get with a blade inserted?

It seems to change each boot and never settles anywhere the same.
Understandable as I'm still testing using SnapshotBladeId and no blade ID resistors.
With this one particular blade, when blade is off readings are consistently in the 79's maxing at 79.68.
When blade is turned on 85-87ish, max i see currently is 88.36.

Again, using a more accurate method like BridgedPullup would address all of this.

I'm just trying to make sure the process in place is correct for what we want it to do.
Like what are we doing NewFont for if there's no change in the blade? I must be missing something in the way things flow.

NoSloppy:
Ok. Recurring theme seems to be just have a more stable Blade ID via either resistors, BladeIDClass, or both.
For some reason my vision is still clouded as to why we'd DoNewFont if the check for 
`if (blade_present_before != blade_present_after)` fails.  But just moving on.

NoSloppy:
closing

profezzorn:
BladeID could be used in ways where blade/no blade doesn't really make sense.
Like maybe to identify a crystal in the crystal chamber.
In that case, none of the entries in the blade array will have NO_BLADE in it, but we would still want to play the font sound when that changes.

At least that was my thinking when I put it in there...

PR #613, opened 2023-12-25
BC_buttons - replace faston with fastout

NoSloppy:


profezzorn:
IMHO, faston.wav should be retired.
That means adding:

#ifdef ENABLE_FASTON
#warning This feature will be disabled unless people go to the crucible and tell us why it is needed
/* code here */
#endif

Then we can remove it in ProffieOS 9.x

NoSloppy:
> #ifdef ENABLE_FASTON

Sorry, I don't follow exactly.
You suggest I add this to the BC prop? and but should the file name be updated to fastout?
And this should be in a new PR?

profezzorn:
The idea here is to remove the existing faston.wav functionality, since it's kind of silly.
However, since none of us really know if there are people who rely on it or not, I suggest putting inside an ifdef, and pair it with a warning, and then actually delete the code *later*.

The idea would be to do this in any prop that supports faston.wav.
If everybody agrees, we can use ENABLE_FASTON, otherwise it would have to be BC_ENABLE_FASTON.

So, no, the file name would not be updated.

NoSloppy:
> So, no, the file name would not be updated.

But we just changed it to use "fastout" instead of "faston"? That's the name update I was referring to.

profezzorn:
Then I don't understand the question.

NoSloppy:
> Then I don't understand the question.

I'm sorry. I am confused, so we must have a communication gap.
The initial issue was that faston.wav was playing concurrently with out.wav when it was called with FastOn().
In discussing how it might be resolved, you mentioned that "Since “fast on” is working it’s way into the base prop, we might want to add the sound effect for it in hybrid_font instead of a specific prop." 
So then over the course of the day we sorted out the operation of faston.wav, and in doing so, changed the working name to be "fastout.wav". It is working well now. 
I think the only outstanding question was "how much is this used anyway?" to which Fett263 replied that it is used in several ways throughout his prop and styles from his Style Library, and I also presented a use case that I employ.

It seems sudden that you want to retire the faston functionality, especially since of today's update?

I thought we were all set going forward with "fastout.wav" for gesture ignitions and such, and would be encouraging the sound to be used more in the future.
I had just updated my SoundFontNamingConverter to move CFX's poweronf to faston, since that board has a similar motion activated ignition.
So hence this PR, moving the BC prop into the current "fastout.wav" stage.
Please let me know if I am off-base on something, I thought there was only one page to be on here.

profezzorn:
I think I see the problem...

My intention is to only deprecate "faston.wav", and any code involved with playing that particular sound.
EFFECT_FAST_ON, fastout.wav and everything else will still be there, with no changes in behavior.

PR #619, opened 2024-01-05
Clean up usages

NoSloppy:
Just making the Usage comments of the header files more uniform, including a few typo fixes along the way.
To be honest, this was primarily to make them more easily scannable with a script i used to generate the summary list.


functions/ramp.h
```diff
@@ -1,9 +1,9 @@
 #ifndef FUNCTIONS_RAMP_H
 #define FUNCTIONS_RAMP_H
 
+// Usage: Remap<RampF,COLOR>
```
profezzorn, line 4: No, it should be:
Usage: RampF

Also, the description should be updated to:
Returns 0 at base and 32768 at tip.
Example: Mix<RampF, COLOR1, COLOR2, COLOR3>
This would do the same thing as Gradient<COLOR1, COLOR2, COLOR3>

styles/pov.h
```diff
@@ -4,38 +4,40 @@
 // TODO(hubbe): Make it work with FastLED
 #if defined(ENABLE_WS2811)
 
-/* 
+
+// Usage: StylePOV<>
```
profezzorn, line 8: I think you made it more confusing since StylePOV and ContinousPOV now have been moved away from their descriptions.

transitions/wave.h
```diff
@@ -1,6 +1,9 @@
 #ifndef TRANSITIONS_WAVE_H
 #define TRANSITIONS_WAVE_H
 
+// Usage: TrWaveX<COLOR, FADEOUT_MS, WAVE_SIZE, WAVE_MS, WAVE_CENTER>
+// COLOR: COLOR
+// FADEOUT_MS, WAVE_SIZE, WAVE_MS, WAVE_CENTER: FUNCTIONS
```
profezzorn, line 6: s/FUNCTIONS/FUNCTION/  (here and in other places)

NoSloppy: sorry, huh?

functions/ramp.h
NoSloppy: done.

transitions/wave.h
NoSloppy: Done.

styles/pov.h
NoSloppy: Restored

transitions/wave.h
profezzorn: Sorry, s/A/B/ is search-and-replace in ed, sed and vi, which are ancient unix text editors of various kinds.

NoSloppy: I use find with sed to batch replace text in my config files 😃

PR #620, opened 2024-01-06
also update SB_Postoff() with EffectLocation location

NoSloppy:


profezzorn:
It doesn't need it, so it's better to remove it from the call site. (done)

PR #624, opened 2024-01-24
rotate_presets -  no beep if font.wav

NoSloppy:


props/prop_base.h
```diff
@@ -538,7 +538,9 @@ class PropBase : CommandParser, Looper, protected SaberBase {
     last_on_time_ = millis();
 #endif
 #ifdef ENABLE_AUDIO
-    beeper.Beep(0.05, 2000.0);
+    if (!SFX_font) {
```
profezzorn, line 541: This won't work right, because it checks if SFX_font has files *before* we switch to a new preset.
Seems like we should just do set "announce" to true instead.

NoSloppy: hmmm. I see how `SetPreset()` isn't until after this, so we'd be checking the *current* font mistakenly, but I'm not following how an announce bool would help here, unless you mean we'd need to call `rotate_presets(true)` in the prop?
What if we just moved the check to after we SetPreset instead, like
cpp
```
  virtual void rotate_presets() {
#ifdef IDLE_OFF_TIME
    last_on_time_ = millis();
#endif
    LOCK_SD(true);
    current_preset_.Load(-1);  // load last preset
    current_preset_.SaveAt(0); // save in first position, shifting all other presets down
    LOCK_SD(false);
    SetPreset(0, true);
#ifdef ENABLE_AUDIO
    if (!SFX_font) beeper.Beep(0.05, 2000.0);
#endif
  }
  ```

profezzorn: I mean that it's really up to hybrid_font to play a sound or a beep, and SetPreset() does that if the announce boolean argument is set to true I think. And if that is the case, then the beeper call here should just be removed.

NoSloppy: Removed.

PR #626, opened 2024-02-21
dial menu adjustment

NoSloppy:
use same sensitivity for both vol menu and scroll presets modes.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -581,9 +581,9 @@ public:
 
   void DetectMenuTurn() {
     float a = fusor.angle2() - current_menu_angle_;
+    if (a > M_PI) a-=M_PI*2;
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -581,9 +581,9 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
 
   void DetectMenuTurn() {
     float a = fusor.angle2() - current_menu_angle_;
+    if (a > M_PI) a-=M_PI*2;
```
profezzorn, line 584: This code has nothing to do with sensitivity though, it controls the wrap-around.

NoSloppy: ok. it should still be the same for both modes.

NoSloppy: conflict resolved.

PR #627, opened 2024-02-21
add pointing up helper function

NoSloppy:
in prep for dual blades support

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -606,7 +606,7 @@ public:
   bool is_pointing_up() {
     if (fusor.angle1() > M_PI / 3) return true;
 #ifdef DUAL_BLADES
-    if (fusor.angle1() < -M_PI / 4) return true;
+    if (fusor.angle1() < -M_PI / 3) return true;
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -603,6 +603,14 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     }
   }
 
+  bool is_pointing_up() {
+    if (fusor.angle1() > M_PI / 3) return true;
+#ifdef DUAL_BLADES
+    if (fusor.angle1() < -M_PI / 4) return true;
```
profezzorn, line 609: It's kind of weird that one blade has a more sensetive "up" than the other.
Shouldn't they both be /3 or /4 ?

NoSloppy: I don't disagree, I just used the same conditions for pointing up and pointing down that have been floating around since the SA22C prop was first made. I assume there was a reason, probably to be more forgiving for dragging and/or wrist angle?
I'll change it to both be 30 degrees, thanks.

PR #628, opened 2024-03-03
Add on / off check bypass preon / postoff

NoSloppy:


props/saber_BC_buttons.h
```diff
@@ -611,6 +611,23 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     return false;
   }
 
+  void OnCheckBypassPreon() {
```
profezzorn, line 614: This name is confusing.
"OnSOMETHING()" is usually taken to mean "SOMETHING" happened, and this is the function that is called when that happens. For example, "OnPowerButtonPressed()".

I don't think that is what this function does, right?
This looks like a helper function to either turn saber on, or turn it on fast.
Maybe just call it "TurnOn()" ?
Or "TurnOnHelper()" ?

NoSloppy: Ok. TurnOnHelper() and TurnOffHelper() now.

PR #629, opened 2024-03-03
Add scroll preset menu beeps

NoSloppy:


props/saber_BC_buttons.h
```diff
@@ -463,7 +463,23 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
       } else {
         thrust_begin_millis_ = millis();
       }
-  }
+
+    // Scroll Presets timer check - avoid beep/wav overlap
+    if (scroll_presets_timer_trigger_ && DelayTimer(scroll_presets_timer_trigger_, last_timer_time_, beep_seq_duration_ * 1000)) {
```
profezzorn, line 468: Why check the trigger before calling DelayTimer() when DelayTimer will also check the flag?

props/saber_BC_buttons.h
```diff
@@ -463,7 +463,23 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
+        if (scroll_presets_) {
+            SaberBase::DoEffect(EFFECT_NEWFONT, 0);
+        }
+        scroll_presets_timer_trigger_ = false; // Reset the timer flag
```
profezzorn, line 472: Looks like DelayTimer will already set the flag to false before you get here.

props/saber_BC_buttons.h
profezzorn, line 468: What's the point of using a float for beep_seq_duration_ if you're just going to use it as millis anyways?

props/saber_BC_buttons.h
```diff
@@ -463,7 +463,23 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
+    }
+  }  // Loop()
+
+    bool DelayTimer(bool& trigger_flag, uint32_t& last_timer_time, float duration) {
```
profezzorn, line 476: It might be better to make this into a class so you don't have to pass in a bunch of arguments, kind of like:

class DelayTimer {
   public:
     bool run();
     void trigger();
  private:
     bool triggered_;
     uint32_t trigger_time_;
     uint32_t duration_;
};

NoSloppy: ok should be better now?

props/saber_BC_buttons.h
```diff
@@ -1208,6 +1265,8 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
   }
 
 private:
+  DelayTimer scrollPresetsTimer;
```
profezzorn, line 1268: scroll_presets_timer_

props/saber_BC_buttons.h
profezzorn: Yes, it looks better, but the DelayTimer() function can be removed now, right?

props/saber_BC_buttons.h
NoSloppy: done

NoSloppy:
previous DelayTimer() function removed

props/saber_BC_buttons.h
```diff
@@ -463,7 +489,12 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
       } else {
         thrust_begin_millis_ = millis();
       }
-  }
+
+    // Scroll Presets timer check - avoid beep/wav overlap
+    if (scroll_presets_timer_.timerCheck() && scroll_presets_) {
```
profezzorn, line 494: I'm thinking that it might be cleaner to clear the "triggered" flag when scroll_presets_ becomes false rather than checking it here.

NoSloppy: Well, the timer is not ongoing while in the scroll presets mode, but rather just at the entry point to manage timing for the audio cue.
`triggered_` is set to false as soon as the timer expires.
The additional check for `&& scroll_presets` is just there for the off-chance the mode is exited super quickly and avoid playing font.wav unnecessarily.

NoSloppy: So anything else?

PR #630, opened 2024-03-06
add missing return

NoSloppy:
Appeazes Arduino warning
"ProffieOS/modes/mode.h:8:82: warning: no return statement in function returning non-void [-Wreturn-type]
8 | virtual bool mode_Event2(enum BUTTON button, EVENT event, uint32_t modifiers) {}"

### DIFF
--- a/modes/mode.h
+++ b/modes/mode.h
@@ -5,7 +5,9 @@
 // allows you to temporarily override buttons and other types of events.
 struct ModeInterface {
   virtual void mode_activate(bool onreturn) {}
-  virtual bool mode_Event2(enum BUTTON button, EVENT event, uint32_t modifiers) {}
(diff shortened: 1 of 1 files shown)


modes/mode.h
```diff
@@ -5,7 +5,9 @@
 // allows you to temporarily override buttons and other types of events.
 struct ModeInterface {
   virtual void mode_activate(bool onreturn) {}
-  virtual bool mode_Event2(enum BUTTON button, EVENT event, uint32_t modifiers) {}
+  virtual bool mode_Event2(enum BUTTON button, EVENT event, uint32_t modifiers) {
+    return true;
```
profezzorn, line 9: Should be "return false;"

NoSloppy: ugh. Always go with your first choice.

PR #633, opened 2024-03-17
Bc buttons. add mute sound

NoSloppy:
Plays mute.wav if it exists before muted ignition to avoid confusion.


props/saber_BC_buttons.h
```diff
@@ -490,12 +490,31 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
     if (scroll_presets_timer_.timerCheck() && scroll_presets_) {
         SaberBase::DoEffect(EFFECT_NEWFONT, 0);
     }
   }  // Loop()
 
+
+  float GetSoundLengthNow(Effect* effect) {
+    RefPtr<BufferedWavPlayer> tmp = GetWavPlayerPlaying(effect);
+    if (tmp) {
+      tmp->UpdateSaberBaseSoundInfo();
```
profezzorn, line 510: This is only needed when a sound is started automatically started, and we want to update the SaberBaseSoundInfo after the fact.  When you call PlayCommon(), the sound info will already be updated.

props/saber_BC_buttons.h
```diff
@@ -775,15 +794,14 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
-    case EVENTID(BUTTON_POWER, EVENT_FOURTH_HELD, MODE_OFF):
-      if (!mode_volume_) {
-        if (SetMute(true)) {
-          unmute_on_deactivation_ = true;
-          scroll_presets_ = false;
-          On();
-        }
+case EVENTID(BUTTON_POWER, EVENT_FOURTH_HELD, MODE_OFF):
+    if (!mode_volume_) {
+      if (SFX_mute) {
```
profezzorn, line 799: This will do nothing if there is no mute sound, I don't think that's what you want.

NoSloppy: ah I didn't test with no mute sound.
ok I'll be back.

props/saber_BC_buttons.h
NoSloppy: ah. yes that's easier, thanks. changes made.

props/saber_BC_buttons.h
NoSloppy: ok fixed.

PR #635, opened 2024-03-24
ssd1306.h. fix on.bmp for blaster use (always on)

NoSloppy:
Delay on.bmp until boot,font, or name message has been displayed

### DIFF
--- a/display/ssd1306.h
+++ b/display/ssd1306.h
@@ -474,6 +474,10 @@ public:
   }
 
   void SB_On(EffectLocation location) override {
+    // Delay on.bmp until boot,font, or name message has been displayed for its full duration
+    if (current_effect_ == &IMG_font && t_ < font_config.ProffieOSFontImageDuration) return;
+    if (current_effect_ == &IMG_boot && t_ < font_config.ProffieOSBootImageDuration) return;
+    if (screen_ == SCREEN_MESSAGE && t_ < font_config.ProffieOSTexttImageDuration) return;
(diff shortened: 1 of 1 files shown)


display/ssd1306.h
```diff
@@ -474,6 +474,10 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, Sa
   }
 
   void SB_On(EffectLocation location) override {
+    // Delay on.bmp until boot,font, or name message has been displayed for its full duration
```
profezzorn, line 477: This doesn't actually "delay" anything, it just skips it.

NoSloppy: How about "avoid preempting images for boot, font, or name message, primarily when used with a blaster prop"

profezzorn: I guess the question is really about how we want it to work?

NoSloppy: Yes. Would it be safe to assume that when switching presets the expectation would be to display the preset name or image, just like it does with a saber? And if a boot image exists it's likely intentional for the purpose of viewing when the blaster boots up. It's completely optional as always by just not having those images available if you don't want to see them. I think this fix is just aligning the behavior with what was intended initially and just never fleshed out? So how it's supposed to work seems rather intuitive to me. 🤷

profezzorn: I was thinking that this might skip some step somewhere, but now that I look at it, I think it does not, so my concern isn't really valid.

PR #644, opened 2024-03-30
Add dual blade support bc prop

NoSloppy:
Somewhat large due to controls interference and changes.
The top comments section has adds pertaining to Dual Blades, but I also cleaned up a bit while in there, as well as a whitespace or indention fix here and there. If those changes *need* to go into a separate PR, I could do that.


NoSloppy:
Having 2nd thoughts.
I'll be back.

props/saber_BC_buttons.h
```diff
@@ -638,6 +699,53 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     }
   }
 
+#ifdef DUAL_BLADES
+  void ControlBladeState(bool& this_blade_on_, bool other_blade_on_, bool& this_blade_first_, bool& other_blade_first_, uint16_t THIS_BLADE, uint16_t OTHER_BLADE, bool is_main_blade_) {
```
profezzorn, line 703: Didn't you make a class for grouping the bools that belong to one blade together?

props/saber_BC_buttons.h
```diff
@@ -638,6 +699,53 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     }
   }
 
+#ifdef DUAL_BLADES
```
profezzorn, line 702: What about just using defined(BC_MAIN_BLADE) && defined(BC_SECOND_BLADE) ?

props/saber_BC_buttons.h
```diff
@@ -1278,6 +1558,12 @@ case EVENTID(BUTTON_POWER, EVENT_FOURTH_HELD, MODE_OFF):
   bool battle_mode_ = false;
   bool max_vol_reached_ = false;
   bool min_vol_reached_ = false;
+
+  bool main_blade_on_ = false;
```
profezzorn, line 1556: SaberBase actually keeps track of which blades are on, so we should be able to replace these bools with a function or macro that asks saberbase what is already on.

That will be easier if we make a class for it though.

props/saber_BC_buttons.h
```diff
@@ -479,16 +540,16 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
     // EVENT_THRUST
-      if (mss.y * mss.y + mss.z * mss.z < 16.0 &&
-          mss.x > 14  &&
-          fusor.swing_speed() < 150) {
-        if (millis() - thrust_begin_millis_ > 15) {
-          Event(BUTTON_NONE, EVENT_THRUST);
-          thrust_begin_millis_ = millis();
-        }
-      } else {
+    if (mss.y * mss.y + mss.z * mss.z < 16.0 &&
```
profezzorn, line 543: Some of thsee changes seem to be cosmetic, or just re-indenting something.
They slow down reviewing quite a bit, because it's hard to tell if something changed or not.
In the future, try to keep functional and non-functional changes separate.

props/saber_BC_buttons.h
```diff
@@ -364,8 +413,20 @@ Turn OFF without postoff - Turn OFF while pointing up.
...
+#ifndef MAIN_BLADE
+#define MAIN_BLADE 1
+#endif
+
+#ifndef SECOND_BLADE
+#define SECOND_BLADE 2
+#endif
+
+#if defined(MAIN_BLADE) && defined(SECOND_BLADE)
+static_assert(MAIN_BLADE != SECOND_BLADE, "MAIN_BLADE and SECOND_BLADE cannot be the same");
```
profezzorn, line 425: While this works, I think it might be cleaner to just do:

#if MAIN_BLADE == SECOND_BLADE
#error ....
#endif

props/saber_BC_buttons.h
profezzorn, line 703: So does this function turn blades on? off? toggle?
Neither the name, nor the arguments is telling me what it really does.

props/saber_BC_buttons.h
```diff
@@ -638,6 +699,53 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
+        PVLOG_NORMAL << "** All Blades Deactivated\n";
+        this_blade_first_ = false;
+        other_blade_first_ = false;
+      }
+      this_blade_on_ = !this_blade_on_;
+      saber_off_time_ = millis();
+    }
+  }
+
+  void TurnMainBladeOnOff() {
```
profezzorn, line 740: Does this toggle the main blade?
Is that really what you want?

profezzorn (commented):
First read-through, I expect to find more things in follow-up reviews since this is fairly large.

props/saber_BC_buttons.h
NoSloppy: Not sure if you still see these now that this is closed, but yes, that's what it does and that was the intention.
Should it be ToggleMainBladeOnOff instead?
ToggleMainBladePower? ToggleMainBladeState?
I had thought the name was clear.

profezzorn: The name is clear, my surprise was really that such a function was ever helpful.
Normally, we need separate on and off functions since they are called from different places, but perhaps that's not the case for dual blades...

PR #645, opened 2024-03-31
Add BlasterModeF

NoSloppy:


ProffieOS.ino
```diff
@@ -705,6 +706,10 @@ int prop_GetBulletCount() {
 }
 #endif
 
+int prop_GetBlasterMode() {
+    return prop.GetBlasterMode();
```
profezzorn, line 710: This won't compile if the prop doesn't have a GetBlasterMode function.
To fix that, we should move this function from ProffieOS.ino to props/blaster.h

functions/blaster_mode.h
```diff
@@ -0,0 +1,28 @@
+// blaster_mode.h
+
+#ifndef FUNCTIONS_BLASTER_MODE_H
```
profezzorn, line 3: This should be the first line.

ProffieOS.ino
NoSloppy: I tried that, but there's an issue with the order of things. It seems 'prop' variable isn't available to blaster.h

profezzorn: Right, good point.
How about we add a #define PROP_HAS_GETBLASTERMODE in blaster.h then, and then we'll use an ifdef here.

NoSloppy: I'll work on that. 
Also, (off, but similar topic) if a non-blaster prop is included in the user config, 
but also #define BLASTER_SHOTS_UNTIL_EMPTY is active, an error:
"class SaberBCButtons' has no member named 'GetBulletCount'"

Now, why would someone have a bullet count define for a saber config is the question, but maybe some other error should show just in case they do? like "Blaster defines are not supported in saber configs" or something?

functions/blaster_mode.h
NoSloppy: ok done

ProffieOS.ino
NoSloppy: > add a #define PROP_HAS_GETBLASTERMODE in blaster.h

ok done.

profezzorn: > Now, why would someone have a bullet count define for a saber config is the question,

I'm more worried about the opposite.
Not all blasters have limited number of shots.

ProffieOS.ino
```diff
@@ -705,6 +706,12 @@ int prop_GetBulletCount() {
 }
 #endif
 
+#ifdef PROP_HAS_GETBLASTERMODE
```
profezzorn, line 709: This should be defined in props/blaster.h

NoSloppy: oops. Edit was sitting there, I didn't hit go.
Done.

PR #650, opened 2024-04-04
Diagnostic Delay at boot

NoSloppy:
This allows for the absolute beginning of messages to be monitored with Serial Monitor. Normally, messages prior to the Welcome are not shown.

### DIFF
--- a/ProffieOS.ino
+++ b/ProffieOS.ino
@@ -1633,6 +1633,10 @@ void setup() {
 #endif
 
   Looper::DoSetup();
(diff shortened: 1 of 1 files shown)


profezzorn:
Doesn't CONFIG_STARTUP_DELAY do the same thing basically?

NoSloppy:
> Doesn't CONFIG_STARTUP_DELAY do the same thing basically?
Oh yeah we did this already.
Yes, it does. Removing.
Can we leave the Booting Up printout though?

PR #653, opened 2024-04-04
Fix sound_queue

NoSloppy:
`busy_` was always true. This seems to work now.


sound/sound_queue.h
```diff
@@ -46,6 +46,8 @@ class SoundQueue {
       queue_[0].Play(player.get());
       sounds_--;
       for (int i = 0; i < sounds_; i++) queue_[i] = queue_[i+1];
+      // Reset busy_ when no sounds are left to play
+      if (sounds_ == 0) busy_ = false;
```
profezzorn, line 50: We just started playing a sound, how can we be not busy?

NoSloppy: Isn't that once the queue is exhausted and we're done?

profezzorn: busy() should only return true if the sound queue is empty AND no sound is playing.

profezzorn: Fixed elsewhere.

PR #657, opened 2024-04-30
Fix compile error, faulty blst in saber_fett263_buttons.h

pedker:


props/saber_fett263_buttons.h
```diff
@@ -5548,7 +5548,6 @@ SaberFett263Buttons() : PropBase() {}
         }
 
       case EVENTID(BUTTON_AUX, EVENT_CLICK_SHORT, MODE_ON):
-      case EVENTID(BUTTON_AUX, EVENT_DOUBLE_CLICK, MODE_ON):
```
fdarosa2663, line 5551: This is intentionally there for button "spamming", it would not cause a compile error.  This should not be removed from my prop.

PR #658, opened 2024-05-05
add quote

NoSloppy:
Time to make this official yet?
Saberbase supports the blade effect.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1286,6 +1286,10 @@ public:
       SaberBase::DoBlast();
       return true;
     }
+    if (!strcmp(cmd, "quote")) {
+      SaberBase::DoEffect(EFFECT_QUOTE, 0);
+      return true;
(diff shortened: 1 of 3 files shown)


profezzorn:
We'll also need some support in hybrid_font.h then.

NoSloppy:
> We'll also need some support in hybrid_font.h then.

oh?
Not sure in what capacity you mean.
Props handle calling it...

profezzorn:
If props handle playing the effect, then they should also declare the effect.

If we're going to have a global EFFECT(quote) then I think hybrid font ought to have a:

   case EFFECT_QUOTE: PlayCommon(&SFX_quote); 

or something similar as well.

NoSloppy:
> : PlayCommon(&SFX_quote);

Ah yes, forgot this was handled in the prop in SB_Effect override.
How about now?

PR #659, opened 2024-05-07
Update NO_REPEAT_RANDOM to use Sample Without Replacement

pedker:
With Sample Without Replacement, for each Effect, once a file is randomly selected it cannot be selected again. This resets whenever all files have been selected. 

The improvement of using this is most notable when using Effects that have very distinctly-sounding files, like Quotes. 

This also supports sub-files: picking particular subdirectories and picking files inside of a particular subdirectory are all tracked independently using SWR.

### DIFF
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -141,7 +141,16 @@ class Effect {
     reset();
   }
 
+  ~Effect() {
+#ifdef NO_REPEAT_RANDOM
+    clean();
+#endif
+  }
+
   void reset() {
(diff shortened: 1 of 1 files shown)


profezzorn:
This uses dynamic allocations and takes up too much RAM.

pedker:
If a new version of this was made without using dynamic memory, would it then be acceptable? Are there any other considerations that would also disqualify the feature?

profezzorn:
Any solution that needs one byte per file (or more) will not be acceptable.

PR #660, opened 2024-05-16
Add BC_ to define

NoSloppy:


profezzorn:
Seems like it's a little late to change this now.
(It will break backwards compatibility.)

NoSloppy:
That's fine. Just noticed it and figured keep things uniform. I don't think it's used much anyway, as it was a request from one person.

PR #664, opened 2024-06-20
adjust ProffieOSHumDelay description

NoSloppy:


sound/hybrid_font.h
```diff
@@ -102,7 +102,7 @@ class FontConfigFile : public ConfigFile {
   int volEff;
   // Milliseconds from beginning of out.wav to delay hum.
   // If not specified or set to 0 (no delay), humStart will be used.
```
profezzorn, line 104: How about we fix the line before it too then?

PR #665, opened 2024-06-23
Temp_PR-BC_WIP

NoSloppy:


NoSloppy:
This isn't right. I tested a different version. I'll be back.

NoSloppy:
ok

props/BC_Working_temp_relevant_only.h
```diff
@@ -28,6 +28,105 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
   SaberBCButtons() : PropBase() {}
   const char* name() override { return "SaberBCButtons"; }
 
+BLADE_ID_CLASS_INTERNAL blade_id;  // still needed?
```
profezzorn, line 31: What is this for?

props/BC_Working_temp_relevant_only.h
```diff
@@ -76,103 +202,123 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
 
-  void PrintBladesStatus() {
-    PVLOG_DEBUG << "***** Updated. Current blade on: " 
-           << (SaberBase::OnBlades()[BC_MAIN_BLADE] ? " MAIN " : "") 
-           << (SaberBase::OnBlades()[BC_SECOND_BLADE] ? " SECOND " : "") 
-           << "\n";
-  }
-
+  
   void TurnBladeOn(BladeSet target_blade) {  // receives opposite blade (MAIN or SECOND) so we can exclude it
```
profezzorn, line 214: Is this comment still accurate?
If so, you should fix that.
If not, remove the comment.

profezzorn: I would recommend doing this:

// Add all non-controlled blades.
target_blade = target_blade | ~controlled_blades_;

That way you don't have to feed in the "opposite blade" anymore.

props/BC_Working_temp_relevant_only.h
NoSloppy: Perfect question, and hence the comment to myself. Searching The Crucible for it, it looks like it's a remnant from the first attempt of implementing constant BladeID monitoring.
Gone now.

PR #668, opened 2024-06-29
add ComboBlade and associated wrappers

KeeganMorrow:
Here is the combo blade code that I came up with. I admit that my current test case of combining the two crystal chamber LEDs is about as simple as it gets, but it seems to work pretty nicely for my use case.

The ComboBladeWrapper is used to send out calls to the the actual blades controlling real hardware. The style on the actual blades is set to a ComboBladeStyleWrapper, which forwards the calls for the on to the actual style code, but running against the original ComboBladeWrapper. This allows the ComboBladeWrapper to dispatch the led control calls on to the actual blade hardware while simulating a single blade with the length of both blades put together.

Also see discussion on [crucible](https://crucible.hubbe.net/t/combining-leds-on-different-data-lines-into-a-single-blade/5461)

### DIFF
--- a/blades/combo_blade.h
+++ b/blades/combo_blade.h
@@ -1,0 +1,191 @@
+#ifndef BLADES_COMBO_BLADE_H
+#define BLADES_COMBO_BLADE_H
+
+/*
+Usage: ComboBlade(blade1, blade2)
+This allows multiple blades to be combined into a single linear blade even if
+they use different data lines for control.
+
+The resulting blade will have a length of the number of leds in the first and
+second blade combined. When the style is running, the second blade's LEDs will
+be controlled as though they were in line and after the first blade's LEDs
+
+For a simple example, if you have a saber with 2 crystal LEDs running on different
+data lines rather than in line with one another, you might often do the following.
+
+{ 0,
+  WS281XBladePtr<128, bladePin, Color8::GRB, PowerPINS<bladePowerPin2, bladePowerPin3> >(),
+  WS281XBladePtr<1, blade4Pin, Color8::GRB, PowerPINS<bladePowerPin2, bladePowerPin3> >(),
+  WS281XBladePtr<1, blade3Pin, Color8::GRB, PowerPINS<bladePowerPin2, bladePowerPin3> >(),
+}
+
+The above way allows us to assign either the same style or different styles to
+both LEDs
+
+However, if we assign the same style the LEDs will do the same thing since they
+are both seen as the first LED in a 1 LED longht blade. To avoid this, we can
+use ComboBlade, which will result in the style being run against an intermediate
+blade with a length of 2, rainbow and stripe styles to move between the two
+
+{ 0,
+  WS281XBladePtr<128, bladePin, Color8::GRB, PowerPINS<bladePowerPin2, bladePowerPin3> >(),
+  ComboBlade(WS281XBladePtr<1, blade4Pin, Color8::GRB, PowerPINS<bladePowerPin2, bladePowerPin3> >(),
+             WS281XBladePtr<1, blade3Pin, Color8::GRB, PowerPINS<bladePowerPin2, bladePowerPin3>>()),
+
+In the above example we must reduce the NUM_BLADES by 1 since the ComboBlade itself
+is the only thing that counts as an entry in the blade array.
+
+*/
+
+class ComboBladeWrapper;
+
+class ComboBladeStyleWrapper : public BladeStyle {
+public:
+  // Bladestyle implementation
+  virtual void activate() override {
+    if (real_style_)
+      real_style_->activate();
+  }
+  virtual void deactivate() override {
+    if (real_style_)
+      real_style_->deactivate();
+  }
+  virtual void run(BladeBase* blade) override;
+
+  bool IsHandled(HandledFeature effect) override {
+    if (real_style_)
+      return real_style_->IsHandled(effect);
+    return false;
+  }
+
+  bool NoOnOff() override {
+    if (real_style_)
+      return real_style_->NoOnOff();
+    return true;
+  }
+
+  void setRealStyle(BladeStyle* style){real_style_ = style;}
+
+  BladeStyle* GetRealStyle(){return real_style_;}
+
+  BladeStyle* UnSetRealStyle(){
+    BladeStyle* result = real_style_;
+    real_style_ = nullptr;
+    return result;
+  }
+
+  ComboBladeStyleWrapper(ComboBladeWrapper* comboBlade, BladeBase* primary_blade) :
+    comboBlade_(comboBlade),
+    primary_blade_(primary_blade),
+    real_style_(nullptr){
+  }
+protected:
+  ComboBladeWrapper* comboBlade_;
+  BladeBase* primary_blade_;
+  BladeStyle* real_style_;
+};
+
+class ComboBladeWrapper : public BladeWrapper {
+public:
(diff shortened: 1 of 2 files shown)


blades/combo_blade.h
```diff
@@ -0,0 +1,191 @@
...
+      return real_style_->IsHandled(effect);
+    return false;
+  }
+
+  bool NoOnOff() override {
+    if (real_style_)
+      return real_style_->NoOnOff();
+    return true;
+  }
+
```
profezzorn, line 66: This class should also implement getColor and get_max_arg.

profezzorn: And Charging()

profezzorn (commented):
Not bad, needs some cleaning up, but I don't see anything *wrong* with it. :)

blades/combo_blade.h
```diff
@@ -0,0 +1,191 @@
...
+    else
+      return nullptr;
+  }
+
+private:
+  BladeBase* blade2_;
+  ComboBladeStyleWrapper* dummy_style_;
+};
+
+void ComboBladeStyleWrapper::run(BladeBase* blade) {
```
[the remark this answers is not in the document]
KeeganMorrow: I found I had to separate this out because it is converting a ComboBladeWrapper* back to a BladeBase*

KeeganMorrow:
> Not bad, needs some cleaning up, but I don't see anything _wrong_ with it. :)

Thanks for the feedback, I made some updates based on your comments.

blades/combo_blade.h
```diff
@@ -0,0 +1,200 @@
...
+  virtual void deactivate() override {
+    real_style_->deactivate();
+  }
+  virtual void run(BladeBase* blade) override;
+
+  bool NoOnOff() override {
+    return real_style_->NoOnOff();
+  }
+
+  virtual bool Charging() { return real_style_->Charging();}
```
profezzorn, line 70: add space after ;

blades/combo_blade.h
```diff
@@ -0,0 +1,200 @@
...
+  ComboBladeWrapper(BladeBase* blade1, BladeBase* blade2):
+    blade1_num_leds_(blade1->num_leds()),
+    blade2_(blade2)
+  {
+    blade_ = blade1;
+    dummy_style_ = new ComboBladeStyleWrapper(this, blade1);
+  }
+
+  int num_leds() const override {
+    int result = blade1_num_leds_;
```
profezzorn, line 126: this would be shorter and easier to read:

return blade1_num_leds + blade2->num_leds();

blades/combo_blade.h
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  void SetStyle(BladeStyle* style) override {
+    dummy_style_->setRealStyle(style);
+    blade_->SetStyle(dummy_style_);
+    blade2_->SetStyle(dummy_style_);
+  }
+
+  BladeStyle* UnSetStyle() override {
+    if (dummy_style_)
```
profezzorn, line 172: If you call blade_ / blade2_->UnsetStyle() here then dummy_style will never have a null style (when used), and we can drop the if() statements in that class.

blades/combo_blade.h
```diff
@@ -0,0 +1,200 @@
...
+
+*/
+class ComboBladeWrapper : public BladeWrapper {
+public:
+  ComboBladeWrapper(BladeBase* blade1, BladeBase* blade2):
+    blade1_num_leds_(blade1->num_leds()),
+    blade2_(blade2)
+  {
+    blade_ = blade1;
+    dummy_style_ = new ComboBladeStyleWrapper(this, blade1);
```
profezzorn, line 122: ProffieOS code avoids using 'new' as much as possible to avoid memory fragmentation.
Just make "dummy_style" a ComboBladeStyleWrapper instead of a ComboBladeStyleWrapper*.

blades/combo_blade.h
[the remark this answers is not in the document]
profezzorn: Good point, carry on...

blades/combo_blade.h
```diff
@@ -0,0 +1,200 @@
...
+    return false;
+  }
+
+  OverDriveColor getColor(int i) override {return real_style_->getColor(i);}
+
+  int get_max_arg(int arg) override { return real_style_->get_max_arg(arg);}
+
+  void setRealStyle(BladeStyle* style){real_style_ = style;}
+
+  BladeStyle* GetRealStyle(){return real_style_;}
```
profezzorn, line 84: needs 3 more more spaces, should be:
  BladeStyle* GetRealStyle() { return real_style_; }

(same comment applies in other places as well.)

blades/combo_blade.h
KeeganMorrow: Moving these into a single object resolves this, thanks!

blades/combo_blade.h
```diff
@@ -0,0 +1,143 @@
...
+
+  void SetStyle(BladeStyle* style) override {
+    real_style_ = style;
+    blade_->SetStyle(this);
+    blade2_->SetStyle(this);
+  }
+
+  BladeStyle* UnSetStyle() override {
+    BladeStyle* result = real_style_;
+    real_style_ = nullptr;
```
profezzorn, line 124: Should also call blade_ / blade2_ -> UnSetStyle() here.

blades/combo_blade.h
```diff
@@ -0,0 +1,143 @@
...
+      real_style_->run(this);
+    }
+  }
+
+  bool NoOnOff() override { return real_style_->NoOnOff(); }
+
+  bool Charging() override { return real_style_->Charging(); }
+
+  bool IsHandled(HandledFeature effect) override {
+    if (real_style_)
```
profezzorn, line 68: once UnsetStyle is called properly below, real_style_ will never be null here.

blades/combo_blade.h
```diff
@@ -0,0 +1,143 @@
...
+
+  int get_max_arg(int arg) override { return real_style_->get_max_arg(arg); }
+
+  // BladeBase implementation
+  int num_leds() const override {
+    return blade1_num_leds_ + blade2_->num_leds();
+  }
+  void set(int led, Color16 c) override {
+    if (led < blade1_num_leds_) {
+      return blade_->set(led, c);
```
profezzorn, line 83: Now that I think about it, this doesn't really work.
In most cases, blades don't allocate space for pixel colors until right before the run() call. This means that calling set() on a blade other than the one that called run() is not allowed. Either this class would need to buffer the required pixels, or there would need to be some major rewrites to how blades work.

Sorry it took me so long to realize this. I didn't mean to make you do all the work to clean this up only to have it not checked in....

KeeganMorrow: No worries, that makes sense. I put together a version that caches the results for blade2 and it seems to work as well, though it does mean more dynamic allocation unfortunately (only alternative seemed to be statically allocating more memory than needed).

Sorry for silence on my end, lots going on the last couple weeks so haven't been able to get back to this

blades/combo_blade.h
KeeganMorrow: I'm not sure why, but it actually seems like I get  a crash due to a null pointer somewhere when I tried this, willing to explore more if you think this idea is still worth pursuing

blades/combo_blade.h
profezzorn: The allocation would only happen for those people who actually use this kind of setup, so I think that is not unreasonable.

profezzorn:
I seem to recall this having some difficult-to-resolve comment, but I can't for the life of me remember what it was now...

PR #672, opened 2024-07-15
Add HardStripes

NoSloppy:
How about this?
It's just Stripes but without the gradient.

### DIFF
--- a/styles/stripes.h
+++ b/styles/stripes.h
@@ -20,7 +20,7 @@
 
 template<class... A>
 class StripesHelper {};
-  
+
 template<>
 class StripesHelper<> {
 public:
@@ -67,20 +67,22 @@ public:
     int32_t delta_micros = now_micros - last_micros_;
     last_micros_ = now_micros;
 
-    m = MOD(m + delta_micros * speed / 333, colors_.size * 341*1024);
-    mult_ = (50000*1024 / width);
+    m = MOD(m + delta_micros * speed / 333, colors_.size * 341 * 1024);
+    mult_ = (50000 * 1024 / width);
   }
+
   SimpleColor getColor(int led) {
     // p = 0..341*len(colors)
     int p = ((m + led * mult_) >> 10) % (colors_.size * 341);
-    
+
     SimpleColor ret;
-    ret.c = Color16(0,0,0);
+    ret.c = Color16(0, 0, 0);
     colors_.get(led, p, &ret);
-    colors_.get(led, p + 341 * colors_.size, &ret);
+    colors_.get(led, p + 341 * colors_.size, &ret); // This line creates the gradient blending
     return ret;
   }
-private:
+  
+protected:
   StripesHelper<COLORS...> colors_;
   uint32_t mult_;
   uint32_t last_micros_;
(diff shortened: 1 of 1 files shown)


NoSloppy:
Should probably adhere to the same coding style . standby

NoSloppy:
Ok, nevermind it was right I thinki.

styles/stripes.h
```diff
@@ -104,4 +104,87 @@ class StripesX : public StripesBase<COLORS...> {
...
+class HardStripesHelper {};
+
+template<>
+class HardStripesHelper<> : public StripesHelper<> {
+public:
+  using StripesHelper<>::StripesHelper;
+};
+
+template<class A, class... B>
+class HardStripesHelper<A, B...> : public StripesHelper<A, B...> {
```
profezzorn, line 136: Seems like it would be easier to just add another get() method to the StripesHelper. Like maybe call it "hardGet()" or something?

styles/stripes.h
```diff
@@ -104,4 +104,87 @@ class StripesX : public StripesBase<COLORS...> {
...
+public:
+  using StripesHelper<A, B...>::StripesHelper;
+
+  void get(int led, int p, SimpleColor* ret) {
+    int segment_size = 341;
+    if (p < segment_size) {
+      OverDriveColor tmp = this->a_.getColor(led);
```
profezzorn, line 143: Even for a "hard" transition, I would recommend that there is a one-pixel transition. Basically, each pixel is a small square window, and if that window is half-way between two colors, then it would show half of one color and half of another.

NoSloppy:
this is going awry

styles/stripes.h
```diff
@@ -47,6 +55,24 @@ class StripesHelper<A, B...> {
     }
     b_.get(led, p - 341, ret);
   }
+  void hardGet(int led, int p, SimpleColor* ret) {  // Add hardGet method
+    int segment_size = 341;
+    if (p < segment_size) {
+      OverDriveColor tmp = a_.getColor(led);
+      if (p > 0 && p < 512) { // One-pixel transition logic
```
profezzorn, line 62: This seems wonky to me, pixels are not 512 units, and there shouldn't be any need for sin table stuff here.

I recommend trying something like this:

void hardGet(Range pixel, int led, int p, uint32_t mod, SimpleColor *ret)
  int segment_size = 341;
  uint32_t weight = pixel.intersect_with_stripe(Range(p, p+segment_size), mod);
  if (weight) {
    OverDriveColor tmp = a_.getColor(led);
    if (weight == pixel.size()) {
        ret->c = tmp.c;
        return;
    } else {
      weight = weight * 16384 / pixel.size();
      ret->c.r = clampi32(ret->c.r + ((tmp.c.r * weight) >> 14), 0, 65535);
      ret->c.g = clampi32(ret->c.g + ((tmp.c.g * weight) >> 14), 0, 65535);
      ret->c.b = clampi32(ret->c.b + ((tmp.c.b * weight) >> 14), 0, 65535);
  }
  b_.hardGet(pixel, led, p + segment_size, mod, ret);
}

And below, when you need to call hardGet, call it like:

  this->colors_.hardGet(Range(0, this->mult_ >> 10), led, p, this->colors_.size() * 341, &ret);

NoSloppy: what would `intersect_with_stripe` look like?

profezzorn: Looks like I forgot to "git push", now you can find it here:

https://github.com/profezzorn/ProffieOS/blob/be38b7a038827f4fef73dffc6d5828eaa89105aa/common/range.h#L37

NoSloppy: ok how about now?
Tested, looks right on blade.

styles/stripes.h
```diff
@@ -62,31 +90,47 @@ class StripesBase {
...
   StripesHelper<COLORS...> colors_;
   uint32_t mult_;
   uint32_t last_micros_;
   int32_t m;
 };
 
+template<class... COLORS>
+class HardStripesBase : public StripesBase<COLORS...> {
+public:
+  using StripesBase<COLORS...>::StripesBase;
```
profezzorn, line 121: What is this "using" statement for?

styles/stripes.h
```diff
@@ -102,6 +146,27 @@ class StripesX : public StripesBase<COLORS...> {
   PONUA SVFWrapper<SPEED> speed_;
 };
 
+template<class WIDTH, class SPEED, class... COLORS>
+class HardStripesX : public HardStripesBase<COLORS...> {
+public:
+  using HardStripesBase<COLORS...>::HardStripesBase;
```
profezzorn, line 152: again, what is this for?

styles/stripes.h
NoSloppy: i think remnants from early on getting inheritance going.
better now?

styles/stripes.h
NoSloppy: same

PR #676, opened 2024-07-18
add fast and first preset commands

NoSloppy:
Useful, especially with pesky preons.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1511,8 +1511,24 @@ public:
       next_preset();
       return true;
     }
+    if (!strcmp(cmd, "nf") || !strcmp(cmd, "next_preset_fast")) {
+      next_preset_fast();
+      return true;
+    }
     if (!strcmp(cmd, "p") || (!strcmp(cmd, "prev") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
       previous_preset();
+      return true;
+    }
+    if (!strcmp(cmd, "pf") || !strcmp(cmd, "prev_preset_fast")) {
+      previous_preset_fast();
+      return true;
+    }
+    if (!strcmp(cmd, "f") || (!strcmp(cmd, "first") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
+      first_preset();
+      return true;
+    }
+    if (!strcmp(cmd, "ff") || !strcmp(cmd, "first_preset_fast")) {
+      previous_preset_fast();
       return true;
     }
     if (!strcmp(cmd, "rotate")) {
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1528,7 +1528,7 @@ public:
       return true;
     }
     if (!strcmp(cmd, "ff") || !strcmp(cmd, "first_preset_fast")) {
-      previous_preset_fast();
+      first_preset_fast();
       return true;
     }
     if (!strcmp(cmd, "rotate")) {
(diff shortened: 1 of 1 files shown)


profezzorn:
Wait, the only reason we need these is to make next/prev skip the preon if it's already on, right?
Maybe we should just make that the default all the time?
I'm not sure why you would ever want the preon in that case.

I don't feel a need for a command to go to the first preset though. We already have "set_preset 0" after all, right?

NoSloppy:
> Maybe we should just make that the default all the time?
Meaning next_preset_fast and previous_preset_fast would go away?
That sounds fine to me. Just calling next_preset in its place seems logical.
It still announce font.wav? 






> We already have "set_preset 0" after all, right?

That's literally what it does. I was just making things uniform for readability so a call to do it in the prop (which i am adding) would just follow suit of next and prev.
Raw coded is fine too of course. The end result is the same.

NoSloppy:
Oh, this is just the console commands.
Mostly used for my testing, and again, adding made it look nice and complete / uniform.
Could throw it inside DIAGNOSTIC_COMMADS or something if you want to. Save a few bits?

profezzorn:
What I meant is that if you are changing presets when the prop is on, you probably don't want a preon. You probably just want it to go to the new preset as quickly as possible. 

That means that we change SetPreset to do FastOn if it was on before, then SetPresetFast just becomes SetPreset with announce = false.

At that point, the only difference between next_preset and next_preset_fast is playing font.wav, which I think we want, right?

profezzorn:
Also, making it complete an uniform is not a goal really. Only useful commands please.
Commands take up memory even if they are never used, unless you put an #ifdef around them.
I suppose you can always put stuff that you want in your prop if you like.

NoSloppy:
You said "if you are changing presets when the prop is on, you probably don't want a preon. You probably just want it to go to the new preset as quickly as possible."
and that is what we currently have with the code as-is. I don't think any change needs to happen at the core like SetPreset().
Wouldn't it just be moving things around sort of, and plus it would mean the existing button events in props that call prev/next_preset_fast() will need changing to plain prev/next_preset. 
Seems unfriendly to backward compatibility

Maybe leave all that but just make these commands do it? Like this
```cpp
    if (!strcmp(cmd, "n") || (!strcmp(cmd, "next") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
      If (!Saberbase::IsOn()) {
      next_preset();
      } else {
      next_preset_fast();
      }
      return true;
    }
    if (!strcmp(cmd, "p") || (!strcmp(cmd, "prev") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
      If (!Saberbase::IsOn()) {
      previous_preset();
      } else {
      previous_preset_fast();
      }
      return true;
    }
    if (!strcmp(cmd, "f") || (!strcmp(cmd, "first") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
      If (!Saberbase::IsOn()) {
      first_preset();
      } else {
      first_preset_fast();
      }
      return true;
    }
```

    

> Also, making it complete an uniform is not a goal really. Only useful commands please.
> Commands take up memory even if they are never used, unless you put an #ifdef around them.
> I suppose you can always put stuff that you want in your prop if you like.

Understood and yeah, like I said this PR shouldn't have been sent probably. it's just me making life easier for myself with the commands. That doesn't need to go global be any means.

profezzorn:
Fixed elsewhere.

PR #680, opened 2024-07-27
add easy preon check

NoSloppy:
Sometimes just checking if (!SaberBase::IsOn()) doesn't cut it.

### DIFF
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -258,6 +258,10 @@ public:
     return true;
   }
 
+  bool isDoingPreon() {
+    return state_ == STATE_WAIT_FOR_ON;
+  }
+
   void Deactivate() {
     lock_player_.Free();
     hum_player_.Free();


profezzorn:
What about just using IsOn() ?

NoSloppy:
In the use case I have where blades are turned on or off with a toggle, IsOn() is wrong in one of the 2 states. It's not a good check, so I was hoping this small addition to hybrid_font would be ok.
In addition, it makes all the "prevent during preon" checks throughout the prop more readable without needing a comment explaining why we're checking IsOn()

profezzorn:
I'm still a bit confused.

First of all: It's not clear to me if you need a way to tell the difference between "preon" and "on or preon".
Second: There is a function in prop_base called IsOn which return "on or preon", it is called IsOn() is that not what you need?
Third: If that is not what you need, there is also a variable called on_pending_ which is true during preon, why not just use that?

NoSloppy:
Ah look at that.
I hadn't realized on_pending_ would be in scope. Works perfectly, thanks. Closing this.

PR #685, opened 2024-07-31
delay say() 200ms

NoSloppy:
This allows some breathing space so when turning you don't get "ninety two...nine...ninet...nin...nine...ninety...ninety eight" for example. It lets the change settle for a beat, then speaks.  Working well in testing.

### DIFF
--- a/modes/stepped_mode.h
+++ b/modes/stepped_mode.h
@@ -43,7 +43,7 @@ struct SteppedMode : public SPEC::SelectCancelMode {
   }
   void mode_Loop() override {
     SaberBase::RequestMotion();
-    if (say_time_ && Cyclint<uint32_t>(millis()) > say_time_{
+    if (say_time_ && Cyclint<uint32_t>(millis()) > say_time_) {
       say_time_ = Cyclint<uint32_t>(0);
       say();
     }
--- a/modes/stepped_mode.h
+++ b/modes/stepped_mode.h
@@ -43,8 +43,7 @@ struct SteppedMode : public SPEC::SelectCancelMode {
   }
   void mode_Loop() override {
     SaberBase::RequestMotion();
-    if (say_time_ && Cyclint<uint32_t>(millis()) > say_time_ 
-        && Cyclint<uint32_t>(millis() - say_time_ >= 200)) {
+    if (say_time_ && Cyclint<uint32_t>(millis()) > say_time_{
       say_time_ = Cyclint<uint32_t>(0);
       say();
     }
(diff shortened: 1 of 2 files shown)


modes/stepped_mode.h
```diff
@@ -43,7 +43,8 @@ struct SteppedMode : public SPEC::SelectCancelMode {
   }
   void mode_Loop() override {
     SaberBase::RequestMotion();
-    if (say_time_ && Cyclint<uint32_t>(millis()) > say_time_) {
+    if (say_time_ && Cyclint<uint32_t>(millis()) > say_time_
```
profezzorn, line 46: This isn't the right way to implement this.

1. It would be better to do it where say_time_ is calculated instead of here.
2. say_time_ is normally calculated to be after ccchng.wav has played, which means that it's also possible to do this by adding 200ms of silence to the end of ccchng.wav. Not sure if we want that though.

One additional problem is that if there is no ccchng.wav, then a beep is played instead, but SaberBase::sound_length_ is never set, and no delay would be added.  Maybe we should set the sound_length_ to 200ms in this case but leave it alone if a wav file is found?

NoSloppy: what about like
```cpp
  virtual void update() {
    SaberBase::DoEffect(EFFECT_MENU_CHANGE, 0); // should pay a quiet "tick" every time.
    say_time_ = Cyclint<uint32_t>(millis()) + (SFX_ccchange ? (uint32_t)(SaberBase::sound_length * 1000) : 200);
    if (!say_time_) say_time_ += 1;
    fadeout(SaberBase::sound_length);
  }
  ```

profezzorn: I think it might be better to just set SaberBase::sound_length to 0.2 when we play the beep in hybrid_font.h

NoSloppy: like
```cpp
  virtual void update() {
    SaberBase::DoEffect(EFFECT_MENU_CHANGE, 0); // should pay a quiet "tick" every time.
    if (!SFX_ccchange) SaberBase::sound_length = 0.2;
    say_time_ = Cyclint<uint32_t>(millis()) + (uint32_t)(SaberBase::sound_length * 1000);
    if (!say_time_) say_time_ += 1;
    fadeout(SaberBase::sound_length);
  }
  ```
  ?

profezzorn: No, I mean that we can make the change in hybrid_font.h, not in this file at all.

NoSloppy: ah

PR #687, opened 2024-08-08
bladelength edit - compensate for -1 offset

NoSloppy:
Was showing and speaking zero to max-1.

### DIFF
--- a/modes/default_spec.h
+++ b/modes/default_spec.h
@@ -22,7 +22,7 @@ struct DefaultMenuSpec {
 
 #ifdef DYNAMIC_BLADE_LENGTH  
   typedef mode::ShowLengthStyle<SPEC> ShowLengthStyle;
-  typedef mode::ChangeBladeLengthBlade1<SPEC> ChangeBladeLengthMode;
+  typedef mode::ChangeBladeLengthMode<SPEC> ChangeBladeLengthMode;
 #endif
 #ifdef DYNAMIC_CLASH_THRESHOLD
   typedef mode::SmoothClashThresholdMode<SPEC> ChangeClashThresholdMode;
--- a/modes/settings_menues.h
+++ b/modes/settings_menues.h
@@ -73,6 +73,13 @@ public:
   int len_;
 };
 
+template<class T>
+class CustomShowColorSingleBladeTemplate : public ShowColorSingleBladeTemplate<T, T> {
+public:
+  void SetStyle(BladeBase *blade, const char* str) {
+    blade->SetStyle(new Style<T>());
+  }
+
 template<class SPEC>
 struct ChangeBladeLengthMode : public SPEC::MenuBase {
   virtual int blade() { return menu_current_blade; }
@@ -121,8 +128,7 @@ struct ChangeBladeLengthMode : public SPEC::MenuBase {
     return pos_ == -1 ? size() : pos_; 
   }
 
-
-  ShowColorSingleBladeTemplate<typename SPEC::ShowLengthStyle> showlen_;
+  CustomShowColorSingleBladeTemplate<typename SPEC::ShowLengthStyle> showlen_;
   int saved_len_;
   uint16_t pos_;
 };
--- a/modes/settings_menues.h
+++ b/modes/settings_menues.h
@@ -74,7 +74,7 @@ public:
 };
 
 template<class SPEC>
-struct ChangeBladeLengthBlade1 : public SPEC::MenuBase {
+struct ChangeBladeLengthMode : public SPEC::MenuBase {
   virtual int blade() { return menu_current_blade; }
   void mode_activate(bool onreturn) override {
     int len = prop_GetBladeLength(blade());
--- a/modes/settings_menues.h
+++ b/modes/settings_menues.h
@@ -91,7 +91,7 @@ struct ChangeBladeLengthBlade1 : public SPEC::MenuBase {
     showlen_.Stop(blade());
   }
   void say() override {
-    getSL<SPEC>()->SayWhole(this->pos_);
+    getSL<SPEC>()->SayWhole(this->pos_ + 1);
   }
   uint16_t size() override {
     return prop_GetMaxBladeLength(blade());
@@ -106,7 +106,7 @@ struct ChangeBladeLengthBlade1 : public SPEC::MenuBase {
     SPEC::MenuBase::exit();
   }
 
-  int getLength() { return this->pos_; }
+  int getLength() { return this->pos_ + 1; }
 
   ShowColorSingleBladeTemplate<typename SPEC::ShowLengthStyle> showlen_;
   int saved_len_;
--- a/modes/settings_menues.h
+++ b/modes/settings_menues.h
@@ -80,36 +80,49 @@ struct ChangeBladeLengthBlade1 : public SPEC::MenuBase {
   void mode_activate(bool onreturn) override {
     int len = prop_GetBladeLength(blade());
     int maxlen = prop_GetMaxBladeLength(blade());
-    saved_len_ = len;
     if (len == -1) len = maxlen;
-    this->pos_ = len;
+    saved_len_ = len;
+    pos_ = saved_len_;
     prop_SetBladeLength(blade(), maxlen);
     showlen_.Start(blade());
     SPEC::MenuBase::mode_activate(onreturn);
+    if (maxlen == 0) exit();
   }
   void mode_deactivate() {
     showlen_.Stop(blade());
   }
   void say() override {
-    getSL<SPEC>()->SayWhole(this->pos_ + 1);
+    getSL<SPEC>()->SayWhole(pos_);
   }
   uint16_t size() override {
     return prop_GetMaxBladeLength(blade());
   }
   void select() override {
-    prop_SetBladeLength(blade(), this->pos_);
+    prop_SetBladeLength(blade(), pos_);
     prop_SaveState();
     SPEC::MenuBase::select();
   }
   void exit() override {
+    pos_ = saved_len_;
     prop_SetBladeLength(blade(), saved_len_);
     SPEC::MenuBase::exit();
   }
 
-  int getLength() { return this->pos_ + 1; }
+  void next() override {
+    pos_ = (pos_ + 1 > size()) ? 0 : pos_ + 1;
+  }
+
+  void prev() override {
+    pos_ = (pos_ - 1 < 0) ? size() : pos_ - 1;
+  }
+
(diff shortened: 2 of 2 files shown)


modes/settings_menues.h
```diff
@@ -91,7 +91,7 @@ struct ChangeBladeLengthBlade1 : public SPEC::MenuBase {
     showlen_.Stop(blade());
   }
   void say() override {
-    getSL<SPEC>()->SayWhole(this->pos_);
+    getSL<SPEC>()->SayWhole(this->pos_ + 1);
```
profezzorn, line 94: You'll also need to change where I set pos_ in mode_activate.

NoSloppy: oh, do you mean the line `if (len == -1) len = maxlen;` should be check if zero instead?
But everything in mode_activate() and size() is all still zero-based indexing at that point so it works correctly as-is.

So if we just set getLength() to this->pos_ + 1 , that propagates through for ShowLengthStyle and all is correct to the user.
Then Saying the number as we see it makes sense too with SayWhole(this->pos_ + 1).
Blade length gets saved to the SD as zero-based index as well (as normal?)
it's working, so it seems right?

profezzorn: I was misreading your change I think.
I thought that the length was generally 1-indexed and tha pos_ is 0-indexed, but I guess that's not what is happening, instead we are changing the SayWhole() and what ShowLengthStyle is showing.

So, with that in mind, I think that this line is fine, but I have some comments for the other change...

modes/settings_menues.h
```diff
@@ -106,7 +106,7 @@ struct ChangeBladeLengthBlade1 : public SPEC::MenuBase {
     SPEC::MenuBase::exit();
   }
 
-  int getLength() { return this->pos_; }
+  int getLength() { return this->pos_ + 1; }
```
profezzorn, line 109: I think it would be better to have getLength return a zero-indexed value and fix ShowLengthStyle instead.

profezzorn: Actually, I don't think this is correct.
ShowLengthStyle is supposed to light up the last LED in the blade.
If this returns pos_ + 1, then a zero-length blade would light up the first led, which is incorrect.
A zero-length blade should not show anything with ShowLengthStyle.

We could change the menu to not allow the value zero, but then we need to change everywhere that pos_ is set or read.

NoSloppy: I thought I had sorted this and was going to update but now I have to ask a few questions first.
Are you saying that the range should allow for the blade to be set to zero?.

I was going to make it so that if the maxlen is set as zero, we can just bail and return to the previous menu.
That's easy, but if the maxlen is not zero, and we want to allow the blade length to be set to a zero value, that makes it harder because now size() would need to account for an additional position?

To confirm(?) blade length is 1-based indexing. pos_ is zero-based.
So I think if we just set this->pos_ to the blade length in mode_activate, all the +1 stuff after doesn't need to be.

Where you said "ShowLengthStyle is supposed to light up the last LED in the blade." Do you mean upon entering the mode, the maxlen should be where we start?
I was thinking we set it so that the saved length is where we start. 

Also, when would there ever be a zero-length blade? I guess whether or not that would make sense or not doesn't matter, we just need to accommodate the possibility?

modes/settings_menues.h
NoSloppy: Actually I think you're right (although pos_ was not being set in here).
If  pos_ *is* set to `len` here, then I think all else is synced. see below.

modes/settings_menues.h
NoSloppy: Well I figured out (one way) how to do this, including accomodating a zero length blade, but it required overriding next() and prev() to the point I'm not sure there's much reason to inherit from MenuBase anymore.

Replying to my own questions
> Where you said "ShowLengthStyle is supposed to light up the last LED in the blade." Do you mean upon entering the mode, the maxlen should be where we start?

Answered for myself based on the code and how it's working. This meant the last LED of the current (or saved) blade length, if none saved, it uses the maxlen.
 

> Are you saying that the range should allow for the blade to be set to zero?.
Well, it can if you want. See latest update to this PR.



> To confirm(?) blade length is 1-based indexing. pos_ is zero-based.

Yes. So to include zero as a valid blade length option, we can omit the modolo and manually calculate when to wrap instead by checking if the next() or prev() operation would result in overshooting the desired range values, and if so, setting the pos_ directly to those values.
If the user has manually set the blade length to 0 in the config, just exit as no length adjustment is possible.


> Also, when would there ever be a zero-length blade?

Maybe if there were a blade detect setup where the user decided to make a zero length for the non-inserted blade state?

modes/settings_menues.h
```diff
@@ -74,42 +74,57 @@ class ShowLengthStyle {
...
-struct ChangeBladeLengthBlade1 : public SPEC::MenuBase {
+struct ChangeBladeLengthMode : public SPEC::MenuBase {
   virtual int blade() { return menu_current_blade; }
   void mode_activate(bool onreturn) override {
     int len = prop_GetBladeLength(blade());
     int maxlen = prop_GetMaxBladeLength(blade());
-    saved_len_ = len;
     if (len == -1) len = maxlen;
-    this->pos_ = len;
+    saved_len_ = len;
```
NoSloppy, line 83: Moving it here saves it at maxlen as opposed to -1 if it was.
Does this make sense?

modes/settings_menues.h
```diff
@@ -74,42 +74,57 @@ class ShowLengthStyle {
...
     SPEC::MenuBase::select();
   }
   void exit() override {
+    pos_ = saved_len_;
     prop_SetBladeLength(blade(), saved_len_);
     SPEC::MenuBase::exit();
   }
 
+  void next() override {
+    pos_ = (pos_ + 1 > size()) ? 0 : pos_ + 1;
```
NoSloppy, line 111: These *could* all still remain as this->pos_ since inherited from MenuBase. Not sure if that's better or if this is cleaner/more appropriate since it's used pretty much exclusively here now anyway.

modes/settings_menues.h
```diff
@@ -73,43 +73,64 @@ class ShowLengthStyle {
   int len_;
 };
 
+template<class T>
```
NoSloppy, line 76: avoids SHORT_STYLE.
Blade would just light up all LEDs if maxlen was 10 or under.
You likely have a better way of doing this, but I didn't want to also edit show_color.h if that should be avoided?

modes/default_spec.h
```diff
@@ -22,7 +22,7 @@ struct DefaultMenuSpec {
 
 #ifdef DYNAMIC_BLADE_LENGTH  
   typedef mode::ShowLengthStyle<SPEC> ShowLengthStyle;
-  typedef mode::ChangeBladeLengthBlade1<SPEC> ChangeBladeLengthMode;
+  typedef mode::ChangeBladeLengthMode<SPEC> ChangeBladeLengthMode;
```
NoSloppy, line 25: more than just blade 1 now

modes/settings_menues.h
```diff
@@ -73,43 +73,62 @@ class ShowLengthStyle {
...
     int maxlen = prop_GetMaxBladeLength(blade());
-    saved_len_ = len;
     if (len == -1) len = maxlen;
-    this->pos_ = len;
+    saved_len_ = len;
+    pos_ = saved_len_;
     prop_SetBladeLength(blade(), maxlen);
     showlen_.Start(blade());
     SPEC::MenuBase::mode_activate(onreturn);
+    if (maxlen == 0) exit();
```
NoSloppy, line 95: If the user has manually set the blade length to 0 in the config, just exit as no length adjustment is possible.

modes/settings_menues.h
```diff
@@ -73,43 +73,62 @@ class ShowLengthStyle {
...
+
+  void prev() override {
+    pos_ = (pos_ - 1 < 0) ? size() : pos_ - 1;
+  }
+
   virtual int steps_per_revolution() { return 16; }
-  int getLength() { return this->pos_; }
 
-  ShowColorSingleBladeTemplate<typename SPEC::ShowLengthStyle> showlen_;
+  int getLength() { return pos_; }
```
NoSloppy, line 127: These *could* all still remain as this->pos_ since inherited from MenuBase. Not sure if that's better or if this is cleaner/more appropriate since it's used pretty much exclusively here now anyway.

modes/settings_menues.h
profezzorn: > I thought I had sorted this and was going to update but now I have to ask a few questions first.
Are you saying that the range should allow for the blade to be set to zero?.

No. I'm saying that my original assumption was that there was a 0-index vs. 1-index issue going on here, which is subtly different from allowing a zero length. My original assumption was wrong.

> I was going to make it so that if the maxlen is set as zero, we can just bail and return to the previous menu.
That's easy, but if the maxlen is not zero, and we want to allow the blade length to be set to a zero value, that makes it harder because now size() would need to account for an additional position?

maxlen should never be zero. That would be a string with zero LEDs.
And as I said, I did not intend that a length of zero should be allowed.

> To confirm(?) blade length is 1-based indexing. pos_ is zero-based.

No. blade length is not really 1-based, but if we don't want to allow the zero value, then we still need an offset between pos_ and length.  (This might seem like a minor point, but it matters when we calculate size())

> Where you said "ShowLengthStyle is supposed to light up the last LED in the blade." Do you mean upon entering the mode, the maxlen should be where we start?

No, I mean that if length is 1, then the first LED in the string should be lit up.
If length is 2, then the second LED of the string should light up.

modes/settings_menues.h
NoSloppy: As per your response above, this should never really be because it wouldn't make sense. However, it could happen since it's not guarded in the BladeConfig, and/or it could be set intentionally in a no-blade scenario.

modes/settings_menues.h
profezzorn: pos_ should always be 0 .. size() -1

modes/settings_menues.h
NoSloppy: Roger roger.

modes/settings_menues.h
NoSloppy: Got it. Maybe we should throw an error if user config tries to set 0 length.

modes/settings_menues.h
NoSloppy: got it.

profezzorn:
Checked in separately.
Your initial PR was *mostly* right I think, it just took a bad turn somewhere.

PR #689, opened 2024-08-22
Make stab bi-directional

NoSloppy:
Would this break anything or adversely affect existing use? It might not be the right way to do it, but I was needing a way to have the back-end blade register stabs (melt) when using Dual Blades.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -768,6 +768,11 @@ public:
     return false;
   }
 
+  virtual bool IsStab(const Vec3& diff) const {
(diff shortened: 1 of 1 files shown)


profezzorn:
I wonder if maybe the right way to do this is to break out the calculation into a separate function, which can then be overridden in subclasses?

NoSloppy:
> I wonder if maybe the right way to do this is to break out the calculation into a separate function, which can then be overridden in subclasses?

ok how about now?

PR #690, opened 2024-08-23
accommodate font search path for errors

NoSloppy:
When multiple directories from the font search path are scanned, there was no telling which one had a prooblem. By reordering things a bit, we can have the offending font name print with it's error.

### DIFF
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -625,21 +625,22 @@ class Effect {
 
     for (const char* dir = current_directory; dir; dir = next_current_directory(dir)) {
       ScanOneDirectory(dir);
-    }
-
-    bool warned = false;
-    for (Effect* e = all_effects; e; e = e->next_) {
-      if (e->expected_files() != (size_t)(e->num_files_)) {
-	if (!warned) {
-	  warned = true;
-	  STDOUT.println("");
-	  STDOUT.println("WARNING: This font seems to be missing some files!!");
-	  ProffieOSErrors::error_in_font_directory();
-	}
-	e->Show();
+      bool warned = false;
+      for (Effect* e = all_effects; e; e = e->next_) {
+        if (e->expected_files() != (size_t)(e->num_files_)) {
+          if (!warned) {
+            warned = true;
+            PVLOG_NORMAL << "\nWARNING: The font " << dir << " seems to be missing some files!!\n";
+            ProffieOSErrors::error_in_font_directory();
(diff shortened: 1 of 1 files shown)


NoSloppy:
oops wait. I think the reset at the end of the loop is nullifying all the effects that were found, and then the system thinks there are no files!

NoSloppy:
Hmm. Well that didn't go as planned.
Now I see it cumulatively gets ALL the files from all the fonts in the font search path and compares it to the num_files it expects?
So there's no way to "catch" the guilty font name. Not as it works now anyway it seems.

However I did notice that having an error in common, it doesn't report an error. I'm assuming voicepack files don't get checked.

profezzorn:
I'm not sure what you mean by "having an error message in common". The only "error" that gets played from SD is "lowbatt". Is that what you're talking about?

NoSloppy:
> I'm not sure what you mean by "having an error message in common". The only "error" that gets played from SD is "lowbatt". Is that what you're talking about?

Sorry. Let me explain this whole thing.
I have 3 directories in my font path
```cpp
{ "testfont2b,common,commonBU"........
```
If there's a files error in the first font, serial monitor reports "WARNING: This font seems to be missing some files!!"
and then reports the problem something like "Found hum files 1-4 + one unnumbered file in subdirs SOME FILES ARE MISSING!"
It doesn't specify which of the fonts has the problem, so I thought it would be cool to just print the `dir` that caused the error. 

So that's what started this, although I gave up when I realized that it's checking the sum of ALL of the fonts in the search path at once (I think?), so flagging the offending directory when it's encountered isn't as simple as I thought it might be.

Anyway, in further testing though, I resolved the first directory error, and then intentionally messed up common by moving clrlst01.wav out of the clrlst subdirectory up into the root. This did not generate an error_in_font_directory and no error printout. Since they are not effect files that a font is expected to contain, "I'm assuming voicepack files don't get checked" as they are not categorically typical effect sound files.

That's what I meant. I hope that makes more sense now.

profezzorn:
> { "testfont2b,common,commonBU"........

I assume you have ; not , between your font directories....

> moving clrlst01.wav out of the clrlst subdirectory up into the root.

Yeah, clrlst isn't searched and checked, instead fixed paths are used.

And yes, it makes sense now.

Also, the "effects" command prints out the directory the files are found in at the end of each line, does it not?

PR #693, opened 2024-09-01
option to report Blade ID continuously

NoSloppy:
Use a define to allow continuous reporting of Blade ID.
This would set `announce` to true each SCAN_BLADE_ID_MILLIS.


NoSloppy:
Or maybe this could just use the already-existing #ifndef DISABLE_DIAGNOSTIC_COMMANDS?

profezzorn:
Merged this accidentially.
I meant to comment and say:

How about we do this when PROFFIEOS_LOG_LEVEL >= 500

NoSloppy:
Resubmitted.

PR #697, opened 2024-09-29
Add Blade ID Scan Millis Timeout

cshannon2:
This pull request adds a timeout to the BLADE_ID_SCAN_MILLIS option. Currently if BLADE_ID_SCAN_MILLIS is enabled it will constantly check for a blade configuration even when the lightsaber is not in use. A side effect of BLADE_ID_SCAN_MILLIS is a blinking light. The blinking light can be observed long after the lightsaber has been in use. The timeout allows the lightsaber to stop searching for a blade if it hasn't been turned on in a certain amount of time. The default time is 10 minutes. The value can be changed via a config file. 

The logic: 
If the blade is off and the blade has been off for longer than the timeout, then the proffie board can stop searching for a blade. BLADE_ID_SCAN_MILLIS can be resumed by turning the lightsaber on and off. Once the lightsaber is turned off the board will restart the countdown.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -584,31 +584,27 @@ public:
 #ifndef SHARED_POWER_PINS
 #warning SHARED_POWER_PINS is recommended when using BLADE_ID_SCAN_MILLIS
 #endif
-    
 #ifdef BLADE_ID_SCAN_TIMEOUT
-    int scan_timeout = BLADE_ID_SCAN_TIMEOUT;
-#else
-    int scan_timeout = 60 * 10 * 1000;
+#define BLADE_ID_SCAN_TIMEOUT 60 * 10 * 1000
 #endif
-
-#ifdef BLADE_ID_STOP_SCAN_WHILE_IGNITED
-    bool scan_while_ignited = false;
-#else
-    bool scan_while_ignited = true;
-#endif
-
-  bool find_blade_again_pending_ = false;
-  uint32_t last_scan_id_ = 0;
+    
+    bool find_blade_again_pending_ = false;
+    uint32_t last_scan_id_ = 0;
     bool ScanBladeIdNow() {
         uint32_t now = millis();
-        bool scan;
-        if (scan_while_ignited == false && IsOn() == true){
+        
+        bool scan = true;
+        #ifdef BLADE_ID_STOP_SCAN_WHILE_IGNITED
+        if (IsOn()) {
             scan = false;
         }
-        else {
-            scan = true;
+        #endif
+        
+        if ((now - blade_id_scan_start_) < BLADE_ID_SCAN_TIMEOUT) {
+            scan = false;
         }
-        if (scan == true && (now - blade_id_scan_start_) < scan_timeout) {
+        
+        if (scan == true) {
             if (now - last_scan_id_ > BLADE_ID_SCAN_MILLIS) {
                 last_scan_id_ = now;
                 size_t best_config = FindBestConfig(PROFFIEOS_LOG_LEVEL >= 500);
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -357,6 +357,9 @@ public:
 #ifdef IDLE_OFF_TIME
     last_on_time_ = millis();
 #endif
+#ifdef BLADE_ID_SCAN_MILLIS
+    blade_id_scan_start_ = millis();
+#endif
     bool on = IsOn();
     BladeSet ret = SaberBase::OnBlades();
     if (on) {
@@ -581,22 +584,30 @@ public:
 #ifndef SHARED_POWER_PINS
 #warning SHARED_POWER_PINS is recommended when using BLADE_ID_SCAN_MILLIS
 #endif
+#ifndef BLADE_ID_SCAN_TIMEOUT
+#define BLADE_ID_SCAN_TIMEOUT 60 * 10 * 1000
+#endif
   bool find_blade_again_pending_ = false;
   uint32_t last_scan_id_ = 0;
-  bool ScanBladeIdNow() {
-    uint32_t now = millis();
-    if (now - last_scan_id_ > BLADE_ID_SCAN_MILLIS) {
-      last_scan_id_ = now;
-      size_t best_config = FindBestConfig();
-      if (current_config != blades + best_config) {
-  // We can't call FindBladeAgain right away because
-  // we're called from the blade. Wait until next loop() call.
-  find_blade_again_pending_ = true;
-      }
-      return true;
+    bool ScanBladeIdNow() {
+        uint32_t now = millis();
+        if (IsOn() == false && (now - blade_id_scan_start_) < BLADE_ID_SCAN_TIMEOUT) {
+            if (now - last_scan_id_ > BLADE_ID_SCAN_MILLIS) {
+                last_scan_id_ = now;
+                size_t best_config = FindBestConfig();
+                if (current_config != blades + best_config) {
+                    // We can't call FindBladeAgain right away because
+                    // we're called from the blade. Wait until next loop() call.
+                    find_blade_again_pending_ = true;
+                }
+                return true;
+            }
+            return false;
+        }
+        else{
+            return false;
+        }
     }
-    return false;
-  }
 
   // Must be called from loop()
   void PollScanId() {
@@ -1120,6 +1131,13 @@ public:
 
     current_mode->mode_Loop();
 
+#ifdef BLADE_ID_SCAN_MILLIS
+    if (SaberBase::IsOn() ||
+        (current_style() && current_style()->Charging())) {
+      blade_id_scan_start_ = millis();
+      }
+#endif
+      
 #ifdef IDLE_OFF_TIME
     if (SaberBase::IsOn() ||
         (current_style() && current_style()->Charging())) {
@@ -1139,6 +1157,10 @@ public:
 
 #ifdef IDLE_OFF_TIME
   uint32_t last_on_time_;
+#endif
+    
+#ifdef BLADE_ID_SCAN_MILLIS
+  uint32_t blade_id_scan_start_;
 #endif
 
 #ifdef SOUND_LIBRARY_REQUIRED
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -584,24 +584,31 @@ public:
 #ifndef SHARED_POWER_PINS
 #warning SHARED_POWER_PINS is recommended when using BLADE_ID_SCAN_MILLIS
 #endif
-#ifndef BLADE_ID_SCAN_TIMEOUT
-#define BLADE_ID_SCAN_TIMEOUT 60 * 10 * 1000
+    
+#ifdef BLADE_ID_SCAN_TIMEOUT
+    int scan_timeout = BLADE_ID_SCAN_TIMEOUT;
+#else
+    int scan_timeout = 60 * 10 * 1000;
 #endif
-#ifndef BLADE_ID_SCAN_WHILE_IGNITED
-#define BLADE_ID_SCAN_WHILE_IGNITED true
+
+#ifdef BLADE_ID_STOP_SCAN_WHILE_IGNITED
+    bool scan_while_ignited = false;
+#else
+    bool scan_while_ignited = true;
 #endif
+
   bool find_blade_again_pending_ = false;
   uint32_t last_scan_id_ = 0;
     bool ScanBladeIdNow() {
         uint32_t now = millis();
         bool scan;
-        if (BLADE_ID_SCAN_WHILE_IGNITED == false && IsOn() == true){
+        if (scan_while_ignited == false && IsOn() == true){
             scan = false;
         }
         else {
             scan = true;
         }
-        if (scan == true && (now - blade_id_scan_start_) < BLADE_ID_SCAN_TIMEOUT) {
+        if (scan == true && (now - blade_id_scan_start_) < scan_timeout) {
             if (now - last_scan_id_ > BLADE_ID_SCAN_MILLIS) {
                 last_scan_id_ = now;
                 size_t best_config = FindBestConfig(PROFFIEOS_LOG_LEVEL >= 500);
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -584,40 +584,38 @@ public:
 #ifndef SHARED_POWER_PINS
 #warning SHARED_POWER_PINS is recommended when using BLADE_ID_SCAN_MILLIS
 #endif
-#ifdef BLADE_ID_SCAN_TIMEOUT
-#define BLADE_ID_SCAN_TIMEOUT 60 * 10 * 1000
-#endif
     
     bool find_blade_again_pending_ = false;
     uint32_t last_scan_id_ = 0;
     bool ScanBladeIdNow() {
         uint32_t now = millis();
         
-        bool scan = true;
-        #ifdef BLADE_ID_STOP_SCAN_WHILE_IGNITED
+        bool scan = (now - last_scan_id_) > BLADE_ID_SCAN_MILLIS;
+        
+#ifdef BLADE_ID_STOP_SCAN_WHILE_IGNITED
         if (IsOn()) {
             scan = false;
         }
-        #endif
+#endif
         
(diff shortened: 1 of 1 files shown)


profezzorn:
Interesting, but maybe we should just be using IDLE_OFF_TIME for this?

cshannon2:
Great Observation. I actually started this using IDLE_OFF_TIME. I was using the variable that IDLE_OFF_TIME creates which is called: last_on_time_. During testing I noticed that I was running into an issue with the BLADE_ID_SCAN_MILLIS unexpectedly restarting after a timeout had occurred. I found that it was because there is another function that modifies last_on_time_. 

`    if (millis() - last_on_time_ > IDLE_OFF_TIME) {
      SaberBase::DoOff(OFF_IDLE, 0);
      last_on_time_ = millis();
    }`

The above code appeared to be resetting the timeout which resulted in restarting the BLADE_ID_SCAN_MILLIS without the lightsaber being powered on. 

I wasn't sure why last_on_time_ was being reset in that function. If we can remove the `last_on_time_ = millis();` from that function then I believe that I could change this to use the  last_on_time_ variable.

NoSloppy:
>  blinking light side effect 

By this do you mean like a single "weak" LED such as on an emitter blade PCB?

cshannon2:
@NoSloppy the short answer is yes. I have seen this side effect discussed in the forums but now I am having a hard time finding the thread. So I will just talk about my experience until I can find the thread. The emitter on my lightsaber has around 9 small leds. When there isn't a blade in the lightsaber and the lightsaber is powered on these leds provide lighting effects. In my experience when BLADE_ID_SCAN_MILLIS is enabled the 9 small leds in the emitter begin to  flash / blink. This flash/blink does not occur everytime a blade_id_scan occurs and the blinking/flashing can vary in its brightness. Based off of what I have read in the forums this can be expected due to how BLADE_ID_SCAN_MILLIS functions. By setting a timeout we can prevent the blinking from occuring after the lightsaber is no longer in use. I am not sure if this change has any impact on battery life but I feel like it could potentially be beneficial for battery life.

props/prop_base.h
```diff
@@ -580,23 +583,31 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
-      size_t best_config = FindBestConfig();
-      if (current_config != blades + best_config) {
-  // We can't call FindBladeAgain right away because
-  // we're called from the blade. Wait until next loop() call.
-  find_blade_again_pending_ = true;
-      }
-      return true;
+    bool ScanBladeIdNow() {
+        uint32_t now = millis();
+        if (IsOn() == false && (now - blade_id_scan_start_) < BLADE_ID_SCAN_TIMEOUT) {
```
profezzorn, line 594: Why is IsOn() == false here?
That will make blade scanning while on not work properly.

cshannon2: The assumption was that a blade config file wouldn't change will the blade was ignited, as a result only search for a blade if the lightsaber is off and the timeout hasn't been reached. If you would like I can just remove the IsOn() check. The resulting if statement would be `if ((now - blade_id_scan_start_) < BLADE_ID_SCAN_TIMEOUT)`

Sabersense-CC: FWIW I think shutting off BLADE_ID_SCAN_MILLIS when the blade is lit is a good idea. In my limited experiments with it, I've found that scan values can change quite significantly when the blade is lit - particularly if using SnapshotID - resulting the array constantly trying to change when you don't want it to. Indeed I would go further and add a small delay between shutting off the blade and restarting BLADE_ID_SCAN_MILLIS to ensure that unusual scan values don't cause further unwanted array switches.

cshannon2: @profezzorn I just wanted to bump this PR. Do you have any additional thoughts on this change?

profezzorn: > FWIW I think shutting off BLADE_ID_SCAN_MILLIS when the blade is lit is a good idea.

Maybe, but we'd need a define for it then.

>  @profezzorn I just wanted to bump this PR. Do you have any additional thoughts on this change?

It is now conflicting, and cannot be merged.

cshannon2: @profezzorn I have resolved the conflict. I have also added a new define BLADE_ID_SCAN_WHILE_IGNITED with the default value of true. This new define will allow the user to turn off scanning when the blade is ignited. By default it will always scan the blade. The only case it will not scan for a blade is if the blade is on and BLADE_ID_SCAN_WHILE_IGNITED is defined in the user's config to equal false. 

Let me know if you have any other concerns.

props/prop_base.h
```diff
@@ -580,23 +583,41 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
 #ifdef BLADE_ID_SCAN_MILLIS
 #ifndef SHARED_POWER_PINS
 #warning SHARED_POWER_PINS is recommended when using BLADE_ID_SCAN_MILLIS
+#endif
+#ifndef BLADE_ID_SCAN_TIMEOUT
+#define BLADE_ID_SCAN_TIMEOUT 60 * 10 * 1000
+#endif
+#ifndef BLADE_ID_SCAN_WHILE_IGNITED
+#define BLADE_ID_SCAN_WHILE_IGNITED true
```
profezzorn, line 591: This not usually how feature defines work in ProffieOS.
(They don't usually have any values, and we use #ifdef / #ifndef / defined() to test them.)

cshannon2: @profezzorn I refactored the way the defines where written. I believe they are identical to the other defines. Let me know if this matches the preferred style.

props/prop_base.h
```diff
@@ -584,24 +584,31 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
 #ifndef SHARED_POWER_PINS
 #warning SHARED_POWER_PINS is recommended when using BLADE_ID_SCAN_MILLIS
 #endif
-#ifndef BLADE_ID_SCAN_TIMEOUT
-#define BLADE_ID_SCAN_TIMEOUT 60 * 10 * 1000
+    
+#ifdef BLADE_ID_SCAN_TIMEOUT
```
profezzorn, line 588: BLADE_ID_SCAN_TIMEOUT was fine (and better) before.
Please change it back.

props/prop_base.h
```diff
@@ -584,24 +584,31 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
+    int scan_timeout = BLADE_ID_SCAN_TIMEOUT;
+#else
+    int scan_timeout = 60 * 10 * 1000;
 #endif
-#ifndef BLADE_ID_SCAN_WHILE_IGNITED
-#define BLADE_ID_SCAN_WHILE_IGNITED true
+
+#ifdef BLADE_ID_STOP_SCAN_WHILE_IGNITED
+    bool scan_while_ignited = false;
```
profezzorn, line 595: I don't want to add variables outside of the function for this.

props/prop_base.h
```diff
@@ -584,24 +584,31 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
         bool scan;
-        if (BLADE_ID_SCAN_WHILE_IGNITED == false && IsOn() == true){
+        if (scan_while_ignited == false && IsOn() == true){
             scan = false;
         }
         else {
             scan = true;
         }
-        if (scan == true && (now - blade_id_scan_start_) < BLADE_ID_SCAN_TIMEOUT) {
+        if (scan == true && (now - blade_id_scan_start_) < scan_timeout) {
```
profezzorn, line 611: scan == true -> scan

props/prop_base.h
```diff
@@ -584,24 +584,31 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
+    bool scan_while_ignited = true;
 #endif
+
   bool find_blade_again_pending_ = false;
   uint32_t last_scan_id_ = 0;
     bool ScanBladeIdNow() {
         uint32_t now = millis();
         bool scan;
-        if (BLADE_ID_SCAN_WHILE_IGNITED == false && IsOn() == true){
+        if (scan_while_ignited == false && IsOn() == true){
```
profezzorn, line 605: I would recommend:
bool scan = true;
#ifdef BLADE_ID_STOP_SCAN_WHILE_IGNITED
   if (IsOn()) scan=false;
#endif

props/prop_base.h
cshannon2: This change has been reverted back.

props/prop_base.h
cshannon2: Simplified the logic. This line should now read if (scan == true) then perform a scan. The logic to determine set the value of scan was refactored in another line of code.

props/prop_base.h
cshannon2: This code has been implemented.

props/prop_base.h
cshannon2: I believe this was addressed when refactoring the scan variable.

props/prop_base.h
```diff
@@ -581,22 +584,43 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
+#define BLADE_ID_SCAN_TIMEOUT 60 * 10 * 1000
+#endif
+    
+    bool find_blade_again_pending_ = false;
+    uint32_t last_scan_id_ = 0;
+    bool ScanBladeIdNow() {
+        uint32_t now = millis();
+        
+        bool scan = true;
+        #ifdef BLADE_ID_STOP_SCAN_WHILE_IGNITED
```
profezzorn, line 597: Don't indent preprocessor statements, makes them harder to see.

props/prop_base.h
```diff
@@ -581,22 +584,43 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
+        if (IsOn()) {
+            scan = false;
+        }
+        #endif
+        
+        if ((now - blade_id_scan_start_) < BLADE_ID_SCAN_TIMEOUT) {
```
profezzorn, line 603: Actually, now that I look at it, this should probably be inside #ifdef BLADE_ID_SCAN_TIMEOUT

props/prop_base.h
```diff
@@ -581,22 +584,43 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
+            scan = false;
+        }
+        
+        if (scan == true) {
+            if (now - last_scan_id_ > BLADE_ID_SCAN_MILLIS) {
```
profezzorn, line 608: maybe move this up to line 596:

bool scan = (now - last_scan_id_) > BLADE_ID_SCAN_MILLIS;

Then line 607 and 608 can just be: if (scan) {

props/prop_base.h
```diff
@@ -581,22 +584,43 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
+                size_t best_config = FindBestConfig(PROFFIEOS_LOG_LEVEL >= 500);
+                if (current_config != blades + best_config) {
+                    // We can't call FindBladeAgain right away because
+                    // we're called from the blade. Wait until next loop() call.
+                    find_blade_again_pending_ = true;
+                }
+                return true;
+            }
+            return false;
+        }
```
profezzorn, line 619: } else {

props/prop_base.h
```diff
@@ -1120,6 +1144,13 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
 
     current_mode->mode_Loop();
 
+#ifdef BLADE_ID_SCAN_MILLIS
+    if (SaberBase::IsOn() ||
+        (current_style() && current_style()->Charging())) {
+      blade_id_scan_start_ = millis();
```
profezzorn, line 1150: This is wrong.
blade_id_scan_start_ should be set when we start scanning (or maybe when we are *not* scanning)
This sets it when the saber is on, which is wrong.

props/prop_base.h
profezzorn: Actually this #Ifdef should probably go away now. (See comments further down.)

props/prop_base.h
cshannon2: This was mirrored from IDLE_OFF_TIME. The goal was to have this feature function similarly to IDLE_OFF_TIME.

profezzorn: Hmm..
Well if that's the case, why not just use the same variable?

props/prop_base.h
cshannon2: Fixed

props/prop_base.h
cshannon2: Implemented

props/prop_base.h
cshannon2: Implemented the suggestion in the other comment.

props/prop_base.h
cshannon2: Implemented

props/prop_base.h
cshannon2: Fixed

props/prop_base.h
```diff
@@ -581,22 +584,41 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
+        bool scan = (now - last_scan_id_) > BLADE_ID_SCAN_MILLIS;
+        
+#ifdef BLADE_ID_STOP_SCAN_WHILE_IGNITED
+        if (IsOn()) {
+            scan = false;
+        }
+#endif
+        
+#ifdef BLADE_ID_SCAN_TIMEOUT
+#define BLADE_ID_SCAN_TIMEOUT 60 * 10 * 1000
```
profezzorn, line 602: Remove this line.

props/prop_base.h
cshannon2: I actually talked a little about this in the earlier conversation. Basically IDLE_OFF_TIME creates a variable called last_on_time_. The last_on_time_ gets modified by another function which causes the counter to for blade scanning to be reset unexpectedly.

props/prop_base.h
cshannon2: GitHub highlighted 4 different lines. Which one are you saying should be removed? Also I am not sure why any of the 4 lines highlighted would be removed.

profezzorn: This line:
#define BLADE_ID_SCAN_TIMEOUT 60 * 10 * 1000

It will produce a warning message that BLADE_ID_SCAN_TIMEOUT is already defined.

cshannon2: This has been completed. Now with the re-write / refactoring of the feature the user should be able to define a timeout if they want blade id to timeout but it won't effect users who don't specify a timeout.

props/prop_base.h
```diff
@@ -356,6 +356,9 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
   BladeSet BladeOff() {
 #ifdef IDLE_OFF_TIME
     last_on_time_ = millis();
+#endif
+#ifdef BLADE_ID_SCAN_MILLIS
```
profezzorn, line 360: shouldn't this be #ifdef BLADE_ID_SCAN_TIMEOUT ?

props/prop_base.h
```diff
@@ -1120,6 +1140,13 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
 
     current_mode->mode_Loop();
 
+#ifdef BLADE_ID_SCAN_MILLIS
```
profezzorn, line 1143: same here (#ifdef BLADE_ID_SCAN_TIMEOUT)

props/prop_base.h
```diff
@@ -1140,6 +1167,10 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
 #ifdef IDLE_OFF_TIME
   uint32_t last_on_time_;
 #endif
+    
+#ifdef BLADE_ID_SCAN_MILLIS
```
profezzorn, line 1171: and here (#ifdef BLADE_ID_SCAN_TIMEOUT)

props/prop_base.h
cshannon2: done

props/prop_base.h
cshannon2: done

props/prop_base.h
cshannon2: done

PR #699, opened 2024-09-29
report blade ID often when log level 500

NoSloppy:


props/prop_base.h
```diff
@@ -587,11 +587,15 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
     uint32_t now = millis();
     if (now - last_scan_id_ > BLADE_ID_SCAN_MILLIS) {
       last_scan_id_ = now;
+#if PROFFIEOS_LOG_LEVEL >= 500
+      size_t best_config = FindBestConfig(true);
```
profezzorn, line 591: no need for if/else/endif just change `true` to `PROFFIEOS_LOG_LEVEL >= 500`

NoSloppy: So good. I have to start thinking this way more.

PR #703, opened 2024-11-26
BCprop Cleanup-Move current_twist_angle_ reset to make more sense.

NoSloppy:


profezzorn:
The twist code is still a bit... twisted, but this seems like an improvement.
(It does change behavior a bit, hope you made sure it works the way you want...)

PR #704, opened 2024-11-27
BCprop - De-wonkify how saved_twist works.

NoSloppy:
Making the time out 300ms seems to be the sweet spot for things to do what I expect.
This was a challenge, lots of mental tracking but it should be "right" and it's working as intended.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -835,6 +835,7 @@ public:
     TWIST_CLOSE,
     TWIST_LEFT,
     TWIST_RIGHT,
+    UNKNOWN_GESTURE,
 
     SHAKE_CLOSE,
     SHAKE_FWD,
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -937,7 +937,7 @@ public:
   }
 
   // The prop should call this from Loop() if it wants to detect twists.
-  void DetectTwist() {
+  virtual void DetectTwist() {
     Vec3 gyro = fusor.gyro();
     bool process = false;
     if (fabsf(gyro.x) > 200.0 &&
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -905,6 +905,8 @@ public:
       case TWIST_LEFT:
       case TWIST_RIGHT:
         return TWIST_CLOSE;
+      case UNKNOWN_GESTURE:
+        return UNKNOWN_GESTURE;
       case SHAKE_CLOSE:
       case SHAKE_FWD:
       case SHAKE_REW:
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -959,36 +959,46 @@ public:
   }
 
   // The prop should call this from Loop() if it wants to detect twists.
-  virtual void DetectTwist() {
+  void DetectTwist() {
+    bool process = DetectTwistStrokes();
+    if (process) {
+      ProcessTwistEvents();
+    }
+  }
+
+  bool DetectTwistStrokes() {
     Vec3 gyro = fusor.gyro();
-    bool process = false;
     if (fabsf(gyro.x) > 200.0 &&
         fabsf(gyro.x) > 3.0f * abs(gyro.y) &&
         fabsf(gyro.x) > 3.0f * abs(gyro.z)) {
-      process = DoGesture(gyro.x > 0 ? TWIST_RIGHT : TWIST_LEFT);
+      return DoGesture(gyro.x > 0 ? TWIST_RIGHT : TWIST_LEFT);
     } else {
-      process = DoGesture(TWIST_CLOSE);
+      return DoGesture(TWIST_CLOSE);
     }
-    if (process) {
-      if ((strokes[NELEM(strokes)-1].type == TWIST_LEFT &&
-           strokes[NELEM(strokes)-2].type == TWIST_RIGHT) ||
-          (strokes[NELEM(strokes)-1].type == TWIST_RIGHT &&
-           strokes[NELEM(strokes)-2].type == TWIST_LEFT)) {
-        if (strokes[NELEM(strokes) -1].length() > 90UL &&
-            strokes[NELEM(strokes) -1].length() < 300UL &&
-            strokes[NELEM(strokes) -2].length() > 90UL &&
-            strokes[NELEM(strokes) -2].length() < 300UL) {
-          uint32_t separation =
+  }
+
+  // Process normal twists)
+  bool ProcessTwistEvents() {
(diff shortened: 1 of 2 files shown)


props/saber_BC_buttons.h
```diff
@@ -1587,39 +1579,69 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
-      CheckSavedTwist(EVENT_TWIST_LEFT);
-      current_twist_angle_ = fusor.angle2();
-    } else if (a > M_PI / 3) {
-      CheckSavedTwist(EVENT_TWIST_RIGHT);
-      current_twist_angle_ = fusor.angle2();
+  // Track the last twist type and time
+  StrokeType saved_twist = UNKNOWN_GESTURE;
+  uint32_t saved_twist_start_millis = 0;
+
+  // Call this from Loop() to detect twists.
```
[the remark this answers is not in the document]
NoSloppy: That was just copied over from base. 
Updated comment.

props/prop_base.h
```diff
@@ -855,6 +855,7 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
     TWIST_CLOSE,
     TWIST_LEFT,
     TWIST_RIGHT,
+    UNKNOWN_GESTURE,
```
profezzorn, line 858: If you add a new value here, you should add it to GetStrokeGroup

props/prop_base.h
```diff
@@ -956,7 +957,7 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
   }
 
   // The prop should call this from Loop() if it wants to detect twists.
-  void DetectTwist() {
+  virtual void DetectTwist() {
```
profezzorn, line 960: Since prop_base never calls this function, there is no need to make it virtual.

props/saber_BC_buttons.h
```diff
@@ -1587,39 +1579,70 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
-    } else if (a > M_PI / 3) {
-      CheckSavedTwist(EVENT_TWIST_RIGHT);
-      current_twist_angle_ = fusor.angle2();
+  // Track the last twist type and time
+  StrokeType saved_twist = UNKNOWN_GESTURE;
+  uint32_t saved_twist_start_millis = 0;
+
+  // Adds a timer to TWIST_LEFT/RIGHT to avoid interference
+  // with normal, 2 direction twists
+  void DetectTwist() override {
```
profezzorn, line 1588: This function is ~75% the same as PropBase::DetectTwist, right?
How about we just break PropBase::DetectTwist into three functions, where the third one would just look like this:

void DetectTwist() {
   bool process = DetectTwistStrokes();
   if (process) {
       ProcessTwistEvents();
   }
}

Then your DetectTwist() can re-use ProcessTwistEvents() and DetectTwistStrokes().
This reduces the amount of duplication, and it also makes it really obvious what the new part is in your function.

props/prop_base.h
NoSloppy: done

props/prop_base.h
NoSloppy: done.

props/saber_BC_buttons.h
NoSloppy: Ok refactored. How about now?

props/saber_BC_buttons.h
```diff
@@ -1587,39 +1579,45 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
+  void DetectTwist() {
+    bool process = DetectTwistStrokes();
+    if (process) {
+      // Start timer if no saved_twist
+      if (saved_twist == UNKNOWN_GESTURE && 
+          (strokes[NELEM(strokes)-1].type == TWIST_LEFT || 
+           strokes[NELEM(strokes)-1].type == TWIST_RIGHT)) {
+        saved_twist = strokes[NELEM(strokes)-1].type;
+        saved_twist_start_millis = millis();
+        return;
```
profezzorn, line 1597: This return feels a bit odd, and it seems tricky to know if it's doing the right thing or not.

I guess the question is: Is it possible that this "return" would block ProcessTwistEvents() at the time that it would match a regular twist event?

Do we even need saved_twist ? Isn't the same information available in the strokes[] array?

NoSloppy: > Is it possible that this "return" would block ProcessTwistEvents() at the time that it would match a regular twist event?

The return inside the if condition should be fine because once a normal twist is confirmed, we don’t need to check for a single twist.

> Do we even need saved_twist ? Isn't the same information available in the strokes[] array?

I needed saved_twist because strokes[] doesn’t handle timing for delaying single twists.

I had this all working in a proprietary way, and then converted over at your suggestion.
So this is the result of that conversion.
If there's a better way, by all means that could be ... better.
But I do know that this current code works as intended.

The key was to allow an action such as Hold POW + Twist to not confuse Hold POW + Twist Left/Right.
I have discrete features on all 3 varieties there.

profezzorn: Don't get me wrong, I think this code is fantastic compared to the old code.
However, I think we can make it even simpler.

I think we can just do this in DoSavedTwist():

Stroke* stroke=strokes + NELEM(strokes)-1;
switch (stroke->type) {
  case TWIST_LEFT:
  case TWIST_RIGHT:
     if (stroke->end_millis == 0) return;
     if (stroke->length() <100) return;
     if (millis()->end_millis < 300) return;
    // May want to add another condition that checks separation between this stroke and the previous one
 }
Event(BUTTON_NONE, stroke->type == TWIST_LEFT ? EVENT_TWIST_LEFT : EVENT_TWIST_RIGHT);
stroke->type = UNKNOWN_GESTURE; // avoid triggering it again

Doing it this way means that we don't need "saved_twist" anymore.
We'll still need to set the stroke type to unknown when ProcessTwistEvents() return true though.


This also solves a problem I think, because process only becomes true when a stroke is *done*, but I think we want the saved stroke to only trigger if no new stroke is initiated, right? (as it is, the saved stroke can still trigger even while you're in the process of doing another stroke.)

NoSloppy: Yeah I was thinking that the way it needs to work is like you say.
A single twist one direction and resting there (for 300ms in this code) should trigger a LEFT/RIGHT twist.
If a 2nd, opposite direction is detected during that time, it cancels the single twist, assuming a regular 2-direction twist is happening. At that point, yes, the saved single twist can be cleared.
If we are saying the same thing, then, I will look into what you said above.

PR #706, opened 2024-11-27
Add idle.wav option

NoSloppy:
This is a proposal for playing an idle.wav, mirroring idle.bmp image behavior.
I’m not sure this is the best way to do this, but I gave this a shot, and it seems to work well.
The feature is enabled with `#define ENABLE_IDLE_SOUND` and plays a sound named idle.wav whenever the blade is off.

Currently, SB_Off() sets idling_ to true, and after retraction (and postoff if it exists) sounds are finished playing, it starts playing idle.wav by triggering it from Loop().

Then idling_  gets set to false at ignition and when IDLE_OFF_TIME expires (when OFF_IDLE) and it stops playing.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -408,7 +408,8 @@ public:
     hybrid_font.hum_player_.Free();
 #ifdef ENABLE_IDLE_SOUND
     hybrid_font.idling_ = true;
-#endif    FreeBladeStyles();
+#endif
+    FreeBladeStyles();
     current_preset_.SetPreset(preset_num);
     AllocateBladeStyles();
     chdir(current_preset_.font.get());
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -405,7 +405,7 @@ public:
     SaveColorChangeIfNeeded();
     // First free all styles, then allocate new ones to avoid memory
     // fragmentation.
-    hybrid_font.idle_player_.Free();
+    hybrid_font.hum_player_.Free();
     hybrid_font.idling_ = true;
     FreeBladeStyles();
     current_preset_.SetPreset(preset_num);
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -405,6 +405,8 @@ public:
     SaveColorChangeIfNeeded();
     // First free all styles, then allocate new ones to avoid memory
     // fragmentation.
+    hybrid_font.idle_player_.Free();
+    hybrid_font.idling_ = true;
     FreeBladeStyles();
     current_preset_.SetPreset(preset_num);
     AllocateBladeStyles();
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -405,10 +405,6 @@ public:
     SaveColorChangeIfNeeded();
     // First free all styles, then allocate new ones to avoid memory
     // fragmentation.
-    hybrid_font.hum_player_.Free();
-#ifdef ENABLE_IDLE_SOUND
-    hybrid_font.idling_ = true;
-#endif
     FreeBladeStyles();
     current_preset_.SetPreset(preset_num);
     AllocateBladeStyles();
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -406,8 +406,9 @@ public:
     // First free all styles, then allocate new ones to avoid memory
     // fragmentation.
     hybrid_font.hum_player_.Free();
+#ifdef ENABLE_IDLE_SOUND
     hybrid_font.idling_ = true;
-    FreeBladeStyles();
+#endif    FreeBladeStyles();
     current_preset_.SetPreset(preset_num);
     AllocateBladeStyles();
     chdir(current_preset_.font.get());
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -740,6 +740,7 @@ EFFECT2(lock, lock);
 EFFECT(swng);
 EFFECT(slsh);
 EFFECT(quote);
+EFFECT2(idle, idle);  // Plays when blade is off
 
 // Looped swing fonts. (SmoothSwing V1/V2)
 EFFECT2(swingl, swingl);  // Looped swing, LOW
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -740,8 +740,9 @@ EFFECT2(lock, lock);
 EFFECT(swng);
 EFFECT(slsh);
 EFFECT(quote);
+#ifdef ENABLE_IDLE_SOUND
 EFFECT2(idle, idle);  // Plays when blade is off
-
+#endif
 // Looped swing fonts. (SmoothSwing V1/V2)
 EFFECT2(swingl, swingl);  // Looped swing, LOW
 EFFECT2(swingh, swingh);  // Looped swing, HIGH
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -708,7 +708,14 @@ EFFECT(preon);
 EFFECT(pstoff);
 
 // Monophonic fonts
-EFFECT(boot);     // also polyphonic
+#ifdef ENABLE_IDLE_SOUND
+EFFECT2(idle, idle);
+EFFECT2(boot, idle);
+EFFECT2(font, idle);
+#else
+EFFECT(boot);
+EFFECT(font);     // also polyphonic
+#endif
 EFFECT(bladein);  // also polyphonic
 EFFECT(bladeout);  // also polyphonic
 EFFECT2(hum, hum);
@@ -726,7 +733,6 @@ EFFECT(spin);     // also polyphonic
 EFFECT(blaster);
 EFFECT2(lockup, lockup);
 EFFECT(poweronf); // force poweron
-EFFECT(font);     // also polyphonic
 EFFECT(bgnlock);  // monophonic and polyphonic begin lock
 EFFECT(endlock);  // Plecter endlock support, used for polyphonic name too
 
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -746,9 +746,7 @@ EFFECT2(lock, lock);
 EFFECT(swng);
 EFFECT(slsh);
 EFFECT(quote);
-#ifdef ENABLE_IDLE_SOUND
-EFFECT2(idle, idle);  // Plays when blade is off
-#endif
+
 // Looped swing fonts. (SmoothSwing V1/V2)
 EFFECT2(swingl, swingl);  // Looped swing, LOW
 EFFECT2(swingh, swingh);  // Looped swing, HIGH
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -815,7 +815,16 @@ public:
     }
 #endif
   }
-  
+      void CheckStopIdle() {
+#if defined(ENABLE_IDLE_SOUND) && !defined(DISABLE_DIAGNOSTIC_COMMANDS)
+       // Since idle sound re-uses hum_player,this will prevent
+      // a swing command from triggering monophonic hum while SaberBase is OFF.
+     if (!SaberBase::IsOn() && !GetWavPlayerPlaying(&SFX_idle)) {
+        StopIdleSound();
+      } 
+#endif
+    }
+
   void StopIdleSound() {
   #ifdef ENABLE_IDLE_SOUND
     idling_ = false;
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -232,12 +232,8 @@ public:
     if (loop) hum_player_->PlayLoop(loop);
    }
 
-  void PlayMonophonic(Effect* f, Effect* loop, float xfade = 0.003f) {
-      Effect::FileID id = f->RandomFile();
-      if (!id && f->GetFollowing()) {
-          id = f->GetFollowing()->RandomFile();
-      }
-      PlayMonophonic(id, loop, xfade);
+  void PlayMonophonic(Effect* f, Effect* loop, float xfade = 0.003f)  {
+    PlayMonophonic(f->RandomFile(), loop, xfade);
   }
   
   // Use after changing alternative.
@@ -266,11 +262,8 @@ public:
     return player;
   }
 
-  RefPtr<BufferedWavPlayer> PlayPolyphonic(Effect* f) {
-    RefPtr<BufferedWavPlayer> p = PlayPolyphonic(f->RandomFile());
-    if (!p && f->GetFollowing())
-        p = PlayPolyphonic(f->GetFollowing()->RandomFile());
-    return p;
+  RefPtr<BufferedWavPlayer> PlayPolyphonic(Effect* f)  {
+    return PlayPolyphonic(f->RandomFile());
   }
 
   void Play(Effect* monophonic, Effect* polyphonic) {
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -28,25 +28,25 @@ public:
       char name[32];
       strcpy(name, "ProffieOS.");
       switch (e->GetFileType()) {
-    case Effect::FileType::SOUND:
-      strcat(name, "SFX.");
-      break;
-    case Effect::FileType::IMAGE:
-      strcat(name, "IMG.");
-      break;
-    default:
-      continue;
+        case Effect::FileType::SOUND:
+          strcat(name, "SFX.");
+          break;
+        case Effect::FileType::IMAGE:
+          strcat(name, "IMG.");
+          break;
+        default:
+          continue;
       }
       strcat(name, e->GetName());
       strcat(name, ".");
       char* x = name + strlen(name);
 
       struct PairedVariable : public VariableBase {
-    Effect* e_;
-    PairedVariable(Effect* e) : e_(e) {}
-    void set(float value) override { e_->SetPaired(value > 0.5); }
-    float get() override { return e_->GetPaired(); }
-    void setDefault() override { e_->SetPaired(false);  }
+        Effect* e_;
+        PairedVariable(Effect* e) : e_(e) {}
+        void set(float value) override { e_->SetPaired(value > 0.5); }
+        float get() override { return e_->GetPaired(); }
+        void setDefault() override { e_->SetPaired(false);  }
       };
       
       strcpy(x, "paired");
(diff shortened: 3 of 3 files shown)


profezzorn:
Apart from how it's implemented, what is it for? who needs it? And I wonder if it would be better to use a prop that is always on, but has one alt mode for "on" and another one for "off" would be better?

NoSloppy:
> Apart from how it's implemented, what is it for? who needs it? And I wonder if it would be better to use a prop that is always on, but has one alt mode for "on" and another one for "off" would be better?

Most suggestions I got _were_ regarding other props besides sabers. Plasma packs, Tron Discs, shelf display with ambiience, etc...
I don't disagree an "always on" prop could do something like this, I just thought that since there's an idle image that plays that this could be small addition to enable with a define and having the file exist.
This could go to a back burner for sure.

profezzorn:
> Most suggestions I got were regarding other props besides sabers. Plasma packs, Tron Discs, shelf display with ambiience, etc...

So is anybody actually making any of these things?

NoSloppy:
> So is anybody actually making any of these things?

There's a ton of Tron Discs, a bunch of Proton packs, Tricorders, Wizard Wands, Kylo Ren Helmets, Canes/staffs, Jetpacks, Thor Hammers, Chest control things, and of course blasters. I'm sure there's more props but yeah... I'd say yes.

profezzorn:
... with proffieboards in them?

NoSloppy:
> ... with proffieboards in them?

Yes? I know, you don't Facebook much lol. 
I could research and find links to projects if really required, but I'll throw out James @Mountainsabers for one.
And of course https://crucible.hubbe.net/c/props/9

profezzorn:
Come to think of it, that is probably the wrong question.
Basically I don't want to add this based on hypotheticals.
I'd like to know of a *specific* use case where this makes sense, as I think that is the only way to test if this is better or worse than other solutions.

mountainsabers101:
Brian wanted me to chime in here as the idea was brought to my attention by a client and i wanted to see if we can use it. The answer to the question if we can use this idea is yes. There are a lot of other projects besides sabers that this could be used for. One of my clients was inquiring about an idle sound while power off instead of using a track with track player....which i had to create a 45 minute sound to accomodate this. My client wanted it just to play on it's own without using anything else to support it. So if i can be done I say it would have many uses..chamber idles sound when off....Energy bindings, motor sounds or internal parts moving while off. Many, many uses. I hope it can be created for the future. I for one will personally use this for my projects if available. Thanks.

sound/effect.h
```diff
@@ -740,6 +740,7 @@ EFFECT2(lock, lock);
 EFFECT(swng);
 EFFECT(slsh);
 EFFECT(quote);
+EFFECT2(idle, idle);  // Plays when blade is off
```
profezzorn, line 743: #ifdef ENABLE_IDLE_SOUND

sound/hybrid_font.h
```diff
@@ -205,6 +205,7 @@ class HybridFont : public SaberBase, public Looper {
   RefPtr<BufferedWavPlayer> next_hum_player_;
   RefPtr<BufferedWavPlayer> swing_player_;
   RefPtr<BufferedWavPlayer> lock_player_;
+  RefPtr<BufferedWavPlayer> idle_player_;
```
profezzorn, line 208: Can't we just re-use the hum player for this?

sound/hybrid_font.h
```diff
@@ -799,8 +803,44 @@ class HybridFont : public SaberBase, public Looper {
...
+        StartIdleSound();
+      }
+    }
+#endif
+  }
+  
+  void StopIdleSound() {
+  #ifdef ENABLE_IDLE_SOUND
+    idling_ = false;
+    if (idle_player_ && idle_player_->isPlaying()) {
```
profezzorn, line 822: Should we maybe play some other effect here?
(a sort of idle "out" sound?)

sound/hybrid_font.h
```diff
@@ -799,8 +803,44 @@ class HybridFont : public SaberBase, public Looper {
...
       }
     }
+    // Wait for retraction sounds to finish, then play idle.wav
+#ifdef ENABLE_IDLE_SOUND
+    if (idling_) {
+      if (GetWavPlayerPlaying(&SFX_in) || GetWavPlayerPlaying(&SFX_pstoff)) {
+        return;
+      } else {
+        idling_ = false;
+        StartIdleSound();
```
profezzorn, line 813: Maybe we should fade the idle sound in?
Or play it gapless after the retraction sound?
Or start it before "in" is done to allow some overlap?

profezzorn (commented):
Still not entirely sure this is a good idea, but if it's behind a define, I guess that's fine...

sound/hybrid_font.h
NoSloppy: Sure can! done.

sound/effect.h
NoSloppy: done.

sound/hybrid_font.h
NoSloppy: My feeling is it's a single event sound, just a background ambience, sfx, whooshing, humming, beeping, whatever. No actual "HERE'S THE IDLE SOUND!" announcement. Although if you want to add something here we could, or leave it for now and add it later if the "need" arises?

sound/hybrid_font.h
NoSloppy: I love the idea.
Fade in works great for me.
Updated.

sound/hybrid_font.h
NoSloppy: If the sounds were built similar to lockups, I could see a bgnidle.wav maybe being an option, yeah.

props/prop_base.h
```diff
@@ -405,6 +405,8 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
     SaveColorChangeIfNeeded();
     // First free all styles, then allocate new ones to avoid memory
     // fragmentation.
+    hybrid_font.hum_player_.Free();
+    hybrid_font.idling_ = true;
```
profezzorn, line 409: #ifdef ENABLE_IDLE_SOUND

sound/hybrid_font.h
```diff
@@ -205,6 +205,7 @@ class HybridFont : public SaberBase, public Looper {
   RefPtr<BufferedWavPlayer> next_hum_player_;
   RefPtr<BufferedWavPlayer> swing_player_;
   RefPtr<BufferedWavPlayer> lock_player_;
+  RefPtr<BufferedWavPlayer> hum_player_;
```
profezzorn, line 208: Does this work? Because there is already a hum_player_ on line 204...

props/prop_base.h
NoSloppy: done

sound/hybrid_font.h
NoSloppy: oops. that's leftover from when there was still an idle_player sorry. 
Don't know if it works, my local copy has 208 blank.
Removed.

profezzorn: > Don't know if it works, my local copy has 208 blank.
So does it work now?
I don't want to review code that doesn't work.

NoSloppy: Oh, sorry that was unclear.
You asked "Does this work" and my reply was supposed to convey that
my local copy did not have it when successfully testing. So I can't say if it works with a redundant hum_player, as i hadn't run the code with it present.
Removed that line and tested. Works well.

sound/hybrid_font.h
```diff
@@ -799,8 +802,58 @@ class HybridFont : public SaberBase, public Looper {
...
+        idling_ = false;
+        StartIdleSound();
+      }
+    }
+#endif
   }
-  bool swinging_ = false;
+  void CheckStopIdle() {
+#ifdef ENABLE_IDLE_SOUND
+    // Since idle sound re-uses hum_player,this will prevent
```
profezzorn, line 819: This comment doesn't make sense to me.
Swings will trigger a hum regardless of whether we use the hum_player or not in a monophonic font.

sound/hybrid_font.h
```diff
@@ -799,8 +802,58 @@ class HybridFont : public SaberBase, public Looper {
...
+    // a swing command from triggering monophonic hum while SaberBase is OFF.
+    if (!SaberBase::IsOn() && !GetWavPlayerPlaying(&SFX_idle)) {
```
profezzorn, line 821: Since it's always going to be the hum_player playing SFX_idle, we can do hum_player_ && hum_player->isPlaying() && hum_player->current_file_id().GetEffect() == &SFX_idle instead.  (longer, but also much faster.)

sound/hybrid_font.h
NoSloppy: This came about after changing a dedicated idle_player_ to hum_player_ instead.
HOWEVER. I just tested again now and everything works fine without CheckStopIdle() even existing...?

Unless I'm crazy because hum was indeed playing unexpectedly during development for some reason and I thought this was needed.
A swing command would gaplessly transition to the hum sound once the swing/slash/spin sound was finished..
The comment was to explain the guarding condition, which was really only specifically for the off chance that a serial command is sent while the blade is off anyway.

Actually, if CheckStopIdle() does exist, it stops idle after the swing command (obviously), which I guess could be considered a negative behavior anyway.
So... I've just removed it.
Oh well.

sound/hybrid_font.h
NoSloppy: This works great, and I see it's more efficient than running through all the wavplayers hunting for idle. 
But, it seems it's not even needed. See above.

profezzorn:
I feel like we should link postoff to idle (if postoff exists)
If postoff doesn't exist, we should link out directly to idle.
Then the idle sound will basically start itself when we turn the saber off.
(Similar to how preon->in or preon->poweron->hum are linked.)

NoSloppy:
> I feel like we should link postoff to idle (if postoff exists) If postoff doesn't exist, we should link out directly to idle. Then the idle sound will basically start itself when we turn the saber off. (Similar to how preon->in or preon->poweron->hum are linked.)

Makes perfect sense. I'll need to try to reverse engineer how pairing actually works under the hood, determine what special conditions are also included for cases of preon and postoff so as to recognize/ignore code that only pertains to those, then figure out how to make idle behave as expected based on how they work.

Question: Would we need to enforce the requirement to have the same number of sounds of both out and/or postoff?
I feel as though idle will typically be more like a hum sound, in that it is sort of a singular, thematic sound for the font.
Now of course, the user could ensure this matching by duplicating the same sound as idle01, idle02 etc..., but I think that seems awkward. 
What if idle is used as more of an entire saber-themed sound, and were to be used in a 'common' folder? The chances of same amount of paired sounds decreases significantly.

Opening a can of worms here, maybe the matching number of files is somehow modifiable, where across the board that restriction could be lifted? Be ing unfamiliar with how it really works under-the-hood at the moment, I assume it is how it is currently for good reason, but since we "can do anything" with code, i figured it doesn't hurt to ask/consider?

Else, maybe idle can be a special case sound that doesn't require it, or... we don't use pairing for it.

NoSloppy:
Well I've spent a good amount of time on this. And while setting idle with GetFollowing does indeed start the sound as expected, there are cons:
1. It doesn't stop lol.
2. It doesn't play after boot.
3. I tried moving the logic from StartIdleSound() into SetPreset after the preset is activated, which *does* work when changing to the preset, however it plays over font.wav, and again... see #1.
4/ i tried moving the StartIdleSound logic into SB_On() right at the beginning. It never executes.

So, not great luck so far.
Other than setting idle to follow in or postoff, all the other points are broken, and it makes me wonder if it's worth it?
I'm all for making it work smarter and cleaner, but it of course needs to behave as expected 100%.
I might need some assistance as to what I'm doing wrong.

profezzorn:
1. The way to fix that would be to call GetWavPlayerPlaying(SFX_idle) in StopIdle().
2&3. We should probably make sure font and boot are followed by idle as well. (You know we can do that with EFFECT2, right?) However, if there is no font or boot, then we have to start it explicitly.
4. In SB_On, we should kill the idle sound, not start it....

NoSloppy:
Well how are we going to delay font and boot so they are not overlapped by idle starting to play? See to me that sounds like a whole nother delay function thing.

profezzorn:
I mean this EFFECT2:
https://github.com/profezzorn/ProffieOS/blob/43bdfbdea9338f8f0937fcbea4fe65b1ca8de938/sound/effect.h#L736
It allows you to set the "following" effect when declaring the effect, so:

EFFECT2(font, idle);
EFFECT2(boot, idle);

I just realized something....
We should fix PlayCommon/PlayPoly/PlayMono to automatically play the following sound if a sound is not available.
So if you do PlayCommon(&SFX_boot); but there is no boot sound, it would automatically play the idle sound (if one is available)
I don't think it already does that....
That should simplify some of our existing code as well.

NoSloppy:
> It allows you to set the "following" effect when declaring the effect, so:
> EFFECT2(font, idle); EFFECT2(boot, idle);
> 

oh *THAT'S* where it get's "looped". I see. 
I was wondering why idle was working as repeating when I had the call as PlayOnce().
Ok. Then what's going on here, where there's both a playOnce *and* a PlayLoop right after?
https://github.com/profezzorn/ProffieOS/blob/935c36e78dfe9b05c9aa04feb4a895631a45530f/sound/hybrid_font.h#L739


> I just realized something.... We should fix PlayCommon/PlayPoly/PlayMono to automatically play the following sound if a sound is not available. So if you do PlayCommon(&SFX_boot); but there is no boot sound, it would automatically play the idle sound (if one is available) I don't think it already does that.... That should simplify some of our existing code as well.

Such as how preon is set to play or not based on existence?

I'm still foggy on a bunch of this audio unfortunately. It's seems straight forward yet pretty complex at the same time. I feel I have the capacity to really get it, I'm just not there yet.
EFFECT2 is an "ah ha!" moment. 
I'll keep digging.

profezzorn:
>  the difference of using GetFollowing and/or just using EFFECT2

There is no difference.
I just use SetFollowing when the "following" depends on some condition. If it's always the same, then I just use EFFECT2.

NoSloppy:
Ok I think I have it working right.
It seems like a lot more code changes than the way I had it originally, but maybe that's what you get when you do it right?
Also it meant I needed to change some lines in my rop, like idle shouldn't play when in Scroll Presets mode, but because EFFECT2 has idle follow font, it would play. So I had to set the SFX_font to nullptr manually, *plus* put it inside an #ifdef guard for each.... however maybe I got it close to correct.

I also added fallback to the following sound if the primary doesn't exist in PlayPolyphonic and PlayMonophonic.

So... how does it look now?
Unfortunately, a spaces vs tab war happened, so it appears as if there's a bunch more diffs than there actually are. 
Not sure of the best way to fix that.

Do you think some comments would help?

sound/hybrid_font.h
```diff
@@ -232,18 +232,22 @@ class HybridFont : public SaberBase, public Looper {
     if (loop) hum_player_->PlayLoop(loop);
    }
 
-  void PlayMonophonic(Effect* f, Effect* loop, float xfade = 0.003f)  {
-    PlayMonophonic(f->RandomFile(), loop, xfade);
+  void PlayMonophonic(Effect* f, Effect* loop, float xfade = 0.003f) {
+      Effect::FileID id = f->RandomFile();
```
profezzorn, line 236: I just checked in a change that does the getfollowing thing in RandomFile, which applies to all the play functions in one go.

sound/hybrid_font.h
```diff
@@ -787,25 +811,38 @@ class HybridFont : public SaberBase, public Looper {
...
+        SaberBase::DoEffect(EFFECT_POSTOFF, saved_location_);
       }
     }
   }
-  bool swinging_ = false;
+
+  void StopIdleSound() {
+#ifdef ENABLE_IDLE_SOUND
+    RefPtr<BufferedWavPlayer> idlePlayer = GetWavPlayerPlaying(&SFX_idle);
+    if (idlePlayer && idlePlayer->isPlaying()) {
```
profezzorn, line 831: you don't need to check isPlaying(), GetWavPlayerPlaying already does that.

profezzorn:
> . So I had to set the SFX_font to nullptr manually, plus put it inside an #ifdef guard for each.... however maybe I got it close to correct.

Might be easier to set the effect volume to something very low temporarily.

NoSloppy:
> Might be easier to set the effect volume to something very low temporarily.

That could be one way, but I feel it's appropriate to not have the "live" feature active while "offline" in a settings menu.
It's not too bad I think. Maybe someday ENABLE_IDLE_SOUND won't be behind a define and the code all cleans up a bit.

sound/hybrid_font.h
NoSloppy: oh nice. Ok let me adjust all this then, because I didn't know you were doing that.
Might be a tomorrow thing.

sound/hybrid_font.h
NoSloppy: I thought so. ok done.

sound/hybrid_font.h
```diff
@@ -472,15 +473,17 @@ class HybridFont : public SaberBase, public Looper {
   }
 
   void SB_Off(OffType off_type, EffectLocation location) override {
+    idling_ = true;
```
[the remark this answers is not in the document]
NoSloppy: Outdated, this is gone. But, yes, both of those. 
It meant "in the state of idling". So SaberBase is OFF, and idle sound should be playing if it is enabled and exists.

NoSloppy:
> > Unfortunately, a spaces vs tab war happened, so it appears as if there's a bunch more diffs than there actually are.
> > Not sure of the best way to fix that.
> 
> Convert all tabs in the whole file to spaces. There should be no tabs.
> 
> Also, github says there are conflicts now.

Well that's the thing. The file had what I assume were tabs that were converted  to spaces when I reposted the file, but the indentation was different.
Like in saber_base.h currently, I see these single dashes (tabs?) that should equate to 6 spaces (the dots), so the indentation is off.
Screenshot 2025-02-14 at 2 05 34 AM

sound/hybrid_font.h
NoSloppy: > I just checked in a change that does the getfollowing thing in RandomFile, which applies to all the play functions in one go.

ProffieOS/modes/settings_menues.h:219:48: error: 'OFF_IDLE' was not declared in this scope;

NoSloppy: Nope, that was easy. Just reverted. Done.

PR #710, opened 2024-12-06
clarify SD mounting in serial mon

NoSloppy:
THis just showed 
```cpp
1
```
Now it's more helpful?
```cpp
SD card mounted to PC 1
```

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1578,7 +1578,7 @@ public:
 #ifdef MOUNT_SD_SETTING
     if (!strcmp(cmd, "sd")) {
       if (arg) LSFS::SetAllowMount(atoi(arg) > 0);
-      STDOUT.println(LSFS::GetAllowMount());
+      STDOUT << "SD card mounted to PC " << LSFS::GetAllowMount() << "\n";
       return true;
     }
 #endif
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1578,7 +1578,9 @@ public:
 #ifdef MOUNT_SD_SETTING
     if (!strcmp(cmd, "sd")) {
       if (arg) LSFS::SetAllowMount(atoi(arg) > 0);
-      STDOUT << "SD card mounted to PC " << LSFS::GetAllowMount() << "\n";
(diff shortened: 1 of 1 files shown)


profezzorn:
That's also not very clear.
Maybe "on"/"off" ?

Also, just because it *can* be mounted by the PC doesn't mean that it is.
This command is really meant to be used by the workbench, which is why I didn't bother making it user friendly...

NoSloppy:
oh. Well from my use, I see it in serial monitor, and only when it gets mounted. 
When disconnecting from the PC, it doesn't say anything. I've never entered `0` to disallow, since there's no way to mount it "accidentally".
So maybe "SD allowed to mount to PC" ? Or just forget it?

profezzorn:
> since there's no way to mount it "accidentally".

Disconnect - reconnect - mounted
I generally do:

sd 1
mount, change files, unmount
sd 0

NoSloppy:
> > since there's no way to mount it "accidentally".
> 
> Disconnect - reconnect - mounted I generally do:
> 
> sd 1 mount, change files, unmount sd 0

Ah. I never found a reason to to `sd 0`
sd 1
it mounts, do edits, unmount via computer
Then just use the saber normally.
Just sd 1 again to re-mount.

So anyway, how about "SD Access ON" and "SD Access OFF" since the voice prompt says "SD Access" ?

props/prop_base.h
```diff
@@ -1578,7 +1578,9 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
 #ifdef MOUNT_SD_SETTING
     if (!strcmp(cmd, "sd")) {
       if (arg) LSFS::SetAllowMount(atoi(arg) > 0);
-      STDOUT << "SD card mounted to PC " << LSFS::GetAllowMount() << "\n";
+      STDOUT << "\nSD Access " 
+             << (LSFS::GetAllowMount() ? "ON" : "OFF") 
+             << "\n\n";
```
profezzorn, line 1583: Single newline is enough.

NoSloppy: ok

PR #716, opened 2024-12-10
Add swng, spin, slsh to serial commands

NoSloppy:
This not only adds good functionality for testing sounds and such, but facilitates access to trigger these sounds via remote control apps.

### DIFF
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -39,6 +39,8 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(POWERSAVE)                    \
     DEFINE_EFFECT(BATTERY_LEVEL)                \
     DEFINE_EFFECT(VOLUME_LEVEL)                 \
+    DEFINE_EFFECT(ACCENT_SWING)                 \
+    DEFINE_EFFECT(ACCENT_SLASH)                 \
     /* Allows style to turn blade ON for interactive effects if prop/style support, FAST_ON skips PREON. */          \
     DEFINE_EFFECT(ON)                           \
     DEFINE_EFFECT(FAST_ON)                      \
--- a/common/saber_base.h
+++ b/common/saber_base.h
@@ -41,7 +41,8 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(VOLUME_LEVEL)                 \
     DEFINE_EFFECT(ACCENT_SWING)                 \
     DEFINE_EFFECT(ACCENT_SLASH)                 \
-    /* Allows style to turn blade ON for interactive effects if prop/style support, FAST_ON skips PREON. */          \
+    DEFINE_EFFECT(SPIN)                         \
+  /* Allows style to turn blade ON for interactive effects if prop/style support, FAST_ON skips PREON. */          \
     DEFINE_EFFECT(ON)                           \
     DEFINE_EFFECT(FAST_ON)                      \
     DEFINE_EFFECT(QUOTE)			\
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1379,26 +1379,18 @@ public:
     if (!strcmp(cmd, "swing")) {
       SaberBase::DoEffect(EFFECT_ACCENT_SWING, 0);
       Event(BUTTON_NONE, EVENT_SWING);
-      // Since idle sound re-uses hum_player,this will prevent
-      // a swing command from triggering monophonic hum while SaberBase is OFF.
-      if (!SaberBase::IsOn() && !GetWavPlayerPlaying(&SFX_idle)) {
-        hybrid_font.StopIdleSound();
-      }
+      hybrid_font.CheckStopIdle();
       return true;
     }
     if (!strcmp(cmd, "slash")) {
       SaberBase::DoEffect(EFFECT_ACCENT_SLASH, 0);
-      if (!SaberBase::IsOn() && !GetWavPlayerPlaying(&SFX_idle)) {
-        hybrid_font.StopIdleSound();
-      }
+      hybrid_font.CheckStopIdle();
       return true;
     }
 #ifdef ENABLE_SPINS
     if (!strcmp(cmd, "spin")) {
       SaberBase::DoEffect(EFFECT_SPIN, 0);
-      if (!SaberBase::IsOn() && !GetWavPlayerPlaying(&SFX_idle)) {
-        hybrid_font.StopIdleSound();
-      } 
+      hybrid_font.CheckStopIdle();
      return true;
     }
 #endif
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1378,16 +1378,27 @@ public:
     }
     if (!strcmp(cmd, "swing")) {
       SaberBase::DoEffect(EFFECT_ACCENT_SWING, 0);
+      // Since idle sound re-uses hum_player,this will prevent
+      // a swing command from triggering monophonic hum while SaberBase is OFF.
+      if (!SaberBase::IsOn() && !GetWavPlayerPlaying(&SFX_idle)) {
+        hybrid_font.StopIdleSound();
+      }
       return true;
     }
     if (!strcmp(cmd, "slash")) {
       SaberBase::DoEffect(EFFECT_ACCENT_SLASH, 0);
+      if (!SaberBase::IsOn() && !GetWavPlayerPlaying(&SFX_idle)) {
+        hybrid_font.StopIdleSound();
+      }
       return true;
     }
 #ifdef ENABLE_SPINS
     if (!strcmp(cmd, "spin")) {
-      hybrid_font.PlayPolyphonic(&SFX_spin);
-      return true;
+      SaberBase::DoEffect(EFFECT_SPIN, 0);
+      if (!SaberBase::IsOn() && !GetWavPlayerPlaying(&SFX_idle)) {
+        hybrid_font.StopIdleSound();
+      } 
+     return true;
     }
 #endif
 
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1379,18 +1379,15 @@ public:
     if (!strcmp(cmd, "swing")) {
       SaberBase::DoEffect(EFFECT_ACCENT_SWING, 0);
       Event(BUTTON_NONE, EVENT_SWING);
-      hybrid_font.CheckStopIdle();
       return true;
     }
     if (!strcmp(cmd, "slash")) {
       SaberBase::DoEffect(EFFECT_ACCENT_SLASH, 0);
-      hybrid_font.CheckStopIdle();
       return true;
     }
 #ifdef ENABLE_SPINS
     if (!strcmp(cmd, "spin")) {
       SaberBase::DoEffect(EFFECT_SPIN, 0);
-      hybrid_font.CheckStopIdle();
      return true;
     }
 #endif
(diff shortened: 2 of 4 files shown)


props/prop_base.h
```diff
@@ -1376,6 +1376,18 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
     }
+    if (!strcmp(cmd, "swing")) {
+      hybrid_font.PlayPolyphonic(&SFX_swng);
+      return true;
+    }
+    if (!strcmp(cmd, "slash")) {
+      hybrid_font.PlayPolyphonic(&SFX_slsh);
+      return true;
+    }
+    if (!strcmp(cmd, "spin")) {
```
profezzorn, line 1387: This needs to be in #ifdef ENABLE_SPINS

props/prop_base.h
```diff
@@ -1376,6 +1376,18 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
       }
       return true;
     }
+    if (!strcmp(cmd, "swing")) {
+      hybrid_font.PlayPolyphonic(&SFX_swng);
```
profezzorn, line 1380: swing is a lot more complicated than this, as it can be monophonic or polyphonic, swng.wav or swing.wav
Perhaps we should even consider breaking the function in hybrid font into a swing-detection function and a swing playing function, and then we could just call the playing function from here.

Or even better, have the detection function emit an event, and then do the playing in SB_Event2...

props/prop_base.h
NoSloppy: whoops. indeed.

props/prop_base.h
NoSloppy: Hmm. Does this small "perk" feature warrant all that?
Would it suffice to just check if swng OR swing exist and then play one that does with PlayCommon()?

NoSloppy: I've updated things to a rough start that works. Monophonic stuff needs to be considered next, but how does it look so far now?

profezzorn (commented):
Overall I like it, needs some work, but I think it's going in the right direction.

sound/hybrid_font.h
```diff
@@ -320,9 +320,15 @@ class HybridFont : public SaberBase, public Looper {
               float s = (rss - font_config.ProffieOSMinSwingAcceleration) / font_config.ProffieOSMaxSwingAcceleration;
 	      effect->SelectFloat(s);
             }
-            swing_player_ = PlayPolyphonic(effect);
+            current_swing_effect_ = effect;
```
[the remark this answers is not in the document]
NoSloppy: Hmmm. Are you thinking a singleton something instead?
Maybe another class needs to be created, or somehow add a parameter to SB_Effect? 
 I am not sure what to do instead.
 Is there a problem with the use of the global variable?

sound/hybrid_font.h
```diff
@@ -320,9 +320,15 @@ class HybridFont : public SaberBase, public Looper {
               float s = (rss - font_config.ProffieOSMinSwingAcceleration) / font_config.ProffieOSMaxSwingAcceleration;
 	      effect->SelectFloat(s);
```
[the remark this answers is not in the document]
NoSloppy: well wouldn't that mean it would need to be repeated for both EFFECT_ACCENT_SWING and EFFECT_ACCENT SLASH cases? Is it not neater to keep it in just one occurrence where it is?
Or again, is this lending itself to all of these changes somehow in a new class or function?
This is starting to grow out of my ability to brain this.

sound/hybrid_font.h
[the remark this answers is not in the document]
profezzorn: Well, the ideal would probably be if we had different EFFECT_* for each case, so when we get to SB_Effect() we don't have to check "current_swing_effect_", we can just look at the "effect" argument and we'll know what to do.

The problem with global variables is that they may get out of sync. Like what happens if something else calls DoEffect(EFFECT_ACCENT_SLASH) ?

common/saber_base.h
```diff
@@ -39,6 +39,8 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(POWERSAVE)                    \
     DEFINE_EFFECT(BATTERY_LEVEL)                \
     DEFINE_EFFECT(VOLUME_LEVEL)                 \
+    DEFINE_EFFECT(ACCENT_SWING)                 \
+    DEFINE_EFFECT(ACCENT_SLASH)                 \
```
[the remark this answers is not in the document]
NoSloppy: > Probably want to add one for spins too.

done

sound/hybrid_font.h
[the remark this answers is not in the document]
NoSloppy: ok how about now?

sound/hybrid_font.h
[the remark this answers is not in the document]
NoSloppy: ok how about now?

props/prop_base.h
```diff
@@ -1376,6 +1376,31 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
       }
       return true;
     }
+    if (!strcmp(cmd, "swing")) {
+      SaberBase::DoEffect(EFFECT_ACCENT_SWING, 0);
```
NoSloppy, line 1380: or... would it be better to just do `prop.Event(BUTTON_NONE, EVENT_SWING);` here, and below, and then SaberBase will handle what EFFECT to do, then SB_Effect will pick it up?
I am not sure that is how it would happen.

profezzorn: It's an interesting thought.
I'm going to have to think about that for a bit, because I'm not sure which is better.
A lot of the props already have some sort of swing detection in them for "swing on" and similar things, so I guess the question is if we can (and should) combine some of this stuff so that swings move from motion->prop->prop base->sound/blade in a similar way that buttons and clashes do.

NoSloppy: Well the way it is at the moment, having both lines behaves as intended.
If OFF, "swing" will do swingon. if on, does EFFECT_ACCENT_SWINGThe "drawback" is it also triggers any EFFECT_ACCENT_SWING in the blade style.

sound/hybrid_font.h
```diff
@@ -553,6 +559,9 @@ class HybridFont : public SaberBase, public Looper {
 	return;
       case EFFECT_PREON: SB_Preon(location); return;
       case EFFECT_POSTOFF: SB_Postoff(); return;
+      case EFFECT_ACCENT_SWING: Play(&SFX_swing, &SFX_swng); return;
+      case EFFECT_ACCENT_SLASH: PlayPolyphonic(&SFX_slsh);  return;
+      case EFFECT_SPIN: Play(&SFX_spin, &SFX_spin); return;
```
profezzorn, line 564: #ifdef ENABLE_SPIN

props/prop_base.h
```diff
@@ -1376,6 +1376,32 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
...
+      Event(BUTTON_NONE, EVENT_SWING);
+      // Since idle sound re-uses hum_player,this will prevent
+      // a swing command from triggering monophonic hum while SaberBase is OFF.
+      if (!SaberBase::IsOn() && !GetWavPlayerPlaying(&SFX_idle)) {
```
profezzorn, line 1384: This seems like it should be fixed in hybrid_font or something (and be guarded by an #ifdef)
Basically, instead of transitioning to the hum sound we should just transition to the idle if the saber is off.
(Assuming we care about supporting idle in monophonic fonts that is...)

sound/hybrid_font.h
NoSloppy: done

props/prop_base.h
NoSloppy: I'm all for not caring, as i personally haven't come across a monophonic font for a good few years now.
That said, it's nice to be robust.
In the off chance someone uses idle.wav in a common folder (hi!) then choosing a mono font with it is likely.


> This seems like it should be fixed in hybrid_font or something (and be guarded by an #ifdef)

I'll look into that. 
#ifdef ... what's the condition? You mean for SPINS again? I try not to forget.

NoSloppy: Moved to a function in hybrid_font called CheckStopIdle().

sound/hybrid_font.h
```diff
@@ -553,6 +559,11 @@ class HybridFont : public SaberBase, public Looper {
 	return;
       case EFFECT_PREON: SB_Preon(location); return;
       case EFFECT_POSTOFF: SB_Postoff(); return;
+      case EFFECT_ACCENT_SWING: Play(&SFX_swing, &SFX_swng); return;
```
profezzorn, line 562: I think this (and also the next one) should set swing_player_
(At least that's what the code used to do.)

NoSloppy: Sorry, I don't follow.
EVENT_SWING is generated, and all the code for a swing runs, including swing_player_ processes, no?
Not sure what explicitly would go here that isn't handled already elsewhere.

profezzorn: Unless I'm missing something, swing_player_ isn't set anywhere anymore.
There is code on line 303 and elsewhere which prevents overlap and changes the volume of the swing sound, but none of that works unless swing_player_ is set correctly.

NoSloppy: > none of that works unless swing_player_ is set correctly.

Ok, I think the Mono sound was what was confusing me, as that doesn't need overlap prevention? So I think the code just needs to specifically handle each sound (mono vs poly) in each effect's case, so that the poly version assigns swing_player_.
Changes made. Let me know if this is accomplishing things correctly?

props/saber_BC_buttons.h
```diff
@@ -1968,6 +1968,24 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
 
 #endif  // BC_DUAL_BLADES
 
+  bool Parse(const char *cmd, const char* arg) override {
```
profezzorn, line 1971: Don't think this belongs in this PR.

sound/hybrid_font.h
```diff
@@ -553,6 +559,25 @@ class HybridFont : public SaberBase, public Looper {
 	return;
       case EFFECT_PREON: SB_Preon(location); return;
       case EFFECT_POSTOFF: SB_Postoff(); return;
+      case EFFECT_ACCENT_SWING:
+        if (guess_monophonic_) {
```
profezzorn, line 563: I think this misses the CFX case, where SFX_humm is being used, and SFX_swing is being played, but in polyphonic mode.

Maybe this?

if (SFX_swng) {
   swing_player_ = PlayPolyphonic(&SFX_swng);
} else if (!guess_monophonic_) {
   swing_player_ = PlayPolyphonic(&SFX_swing);
} else {
   PlayMonophonic(&SFX_swing, &SFX_hum);
}

This prefers SFX_swng if available, which isn't exactly how it worked before, but it seems reasonable to me.

profezzorn:
.... and we have a conflict...

sound/hybrid_font.h
NoSloppy: ok I made it that.
Monophonic fonts are going to be the death of me lol.

NoSloppy:
> .... and we have a conflict...

Ok should be resolved.

props/saber_BC_buttons.h
NoSloppy: I commented about this I swear. Don't know where it went.
THis was in a mode initially, then moved to the body of the BCSaberProp class.

I figured it applied here as it's "adding serial commands"

If you want I can make another standalone PR for it.... ?

profezzorn: Please make separate PR

PR #717, opened 2024-12-22
New saber_sabersense_buttons.h Prop File Request

Sabersense-CC:
INTRODUCING THE UPDATED SABERSENSE™
LIGHTSABER SWITCH CONTROL SYSTEM

OVERVIEW The Sabersense Button System has been engineered with simplicity in mind. It is designed to be very easy to use, bomb-proof and consistent in terms of ‘rules’. It doesn’t require a ‘knack’ to access certain features, and it avoids tricky combinations like holding a button while simultaneously twisting or making other active gestures which can be difficult to execute.

HARMONIZED ONE-BUTTON AND TWO-BUTTON CONTROLS By default, basic features like normal ignition, track playing, force effects, mute ignition, quotes etc. all have the same button controls on both one-button and two-button setups. This means users with large hilt collections which include one-button and two-button sabers don’t need to remember different sets of controls for these core features.

CONSISTENT ‘RULES’ This again means the user has less to remember. For example, one click of POWER with blade off always lights the blade - short click for normal, long click for muted; double clicking POWER always plays a sound effect - track, character quote or force effect depending hilt status (see next paragraph). This is all intended to make it easier to remember the various controls.

COMPREHENSIVE SOUND FONT NAVIGATION One-button and two-button setups both allow users to move forwards or backwards one font at a time, or to skip to first font, last font or middle font. Two-button setups include additional navigation features which allow the user to skip forwards or backwards five and ten fonts at a time.

SEPARATE FORCE EFFECT AND CHARACTER QUOTE PLAYER This allows users to access separate force effects, character quotes and music tracks depending on hilt orientation and status. Tracks can be pIayed with the blade OFF, hilt pointing down, Force effects can be played with blade ON, hilt pointing down, and Character Quotes can be played in both blade ON and OFF states with hilt pointing up. If desired, these orientations can be reversed using #define SABERSENSE_FLIP_AUDIO_PLAYERS. Character Quotes are always played sequentially while Force Effects are played randomly.  

ON DEMAND BLADE ID AND MANUAL ARRAY SELECTOR This allows users to run Blade ID on demand with a button press, eliminating potential blade ID scan errors that might occur under constant automatic scanning when blades are lit. Alternatively the BladeID engine can be reconfigured with a simple define allowing you to cycle through all blade and preset arrays manually, regardless of the actual BladeID status. This is useful if you have a number of different blades, for instance, that all return the same BladeID values and therefore cannot be differentiated by the system. Alternatively it can be used to divide sound fonts into separate arrays; for instance all light side fonts could be on array 1,  all dark side fonts on array 2 and all diagnostic and engineering fonts on array 3.

RESTORE 'FACTORY' DEFAULTS
If you've ever found yourself stuck with settings changes you don't like, you will appreciate this feature. Using a simple, but unlikely, button press (to prevent accidental triggering) the system will delete all user-modified settings such as volume, blade colour and font selection and revert to the settings that were uploaded onto your saber.

OPTIONAL BUTTON ‘CLICKER’ DEFINE For hilts that have buttons with poor tactile feedback (KR’s Scavenger for instance, with the emitter wheel button setup), simply adding a press.wav and release.wav of a clicking sound to the appropriate font folder, and including the #define SABERSENSE_BUTTON_CLICKER in your config will play those effects with their respective button actions to aid in feature navigation. Suitable wav files are included to download from the Sabersense website.

### DIFF
--- a/props/saber_sabersense_buttons.h
+++ b/props/saber_sabersense_buttons.h
@@ -1,0 +1,1273 @@
+/*
+============================================================
+================== SABERSENSE™ PROP FILE ===================
+============================================================
+
+Built on SA22C programming by Matthew McGeary.
+Modified by Chris Carter with substantial support and 
+contributions from Fredrik Hubinette and Brian Conner.
+
+This prop file references certain custom sound files 
+to aid in saber function navigation. These sound files 
+are optional and  are available as a free download from:
+https://sabersense.square.site/downloads
+https://sabersense.square.site/downloads
+
+
+============================================================
+========= BUTTON SYSTEM PRINCIPLES AND LOGIC NOTES ========= 
+
+The Sabersense™ button control system has been engineered 
+for simplicity and usability. Complex gesture controls and 
+features like Battle Mode are NOT included.
+
+By default, the Sabersense™ button prop features harmonized 
+controls between one-button and two-button operation without 
+compromising the greater usability of two-button systems. 
+
+Where practicable, the same controls apply to both one-button 
+and two-button systems, with two-button operation benefitting 
+from some extra features.
+
+Where possible without causing conflicts, some one-button 
+controls also appear in two-button mode, despite two-button 
+having its own controls for the same feature.
+
+The overall intention is that by default, users need only 
+remember a minimum set of control principles in order to 
+access all functions. As such, the logic is that the 
+same button presses do the same thing within the various 
+states, subject to inevitable and obvious variants.
+
+Hence:
+  ONE AND TWO BUTTON:
+      Single click MAIN always lights the blade...
+      Short click lights blade with sound
+      Long click lights blade mute
+
+      Double click MAIN always plays a sound file...
+        Character Quote or Music track with blade OFF
+        Character Quote or Force Effect with blade ON
+        Battery level announcement if held down on second click
+
+      Holding down MAIN (1-button) or AUX (2-button) 
+      and waiting with blade OFF always skips to specific preset...
+        Hilt pointing up - first preset
+        Hilt horizontal - middle preset
+        Hilt pointing down - last preset
+
+      Triple-clicking MAIN is always a saber management feature
+        Colour change with blade ON
+        BladeID/Array Switch with blade OFF
+
+  TWO BUTTON ONLY:
+      Short clicking AUX with blade OFF always moves to a different preset, 
+      forward with hilt pointing up, backwards with hilt pointing down...
+        Single short click - one preset
+        Double short click - five presets
+        Triple short click - ten presets
+
+      Holding MAIN and short clicking AUX always enters a control menu...
+        Colour change with blade ON
+        Volume menu with blade OFF
+
+
+==========================================================
+=================== 1 BUTTON CONTROLS ==================== 
+
+MAIN FUNCTIONS
+  Activate blade            Short click while OFF. *
+  Activate blade mute       Long click while OFF, hilt horizontal.
+                              (Hold for one second then release).
+  Deactivate blade          Hold and wait until blade is off while ON.
+
+FUNCTIONS WITH BLADE OFF
+  Next preset               Long click while OFF, hilt pointing up.
+                              (Hold for one second then release).
+  Previous preset           Long click while OFF, hilt pointing down.
+                              (Hold for one second then release).
+  Skip to first preset      Press and hold until it switches, hilt pointing upwards.
+  Skip to middle preset     Press and hold until it switches, hilt horizontal.
+  Skip to last preset       Press and hold until it switches, hilt pointing downwards.
+  Play Character Quote      Fast double-click, hilt pointing up, plays sequentially. **
+  Play Music Track          Fast double-click, hilt pointing down. **
+  Speak battery voltage     Fast double-click-and-hold while OFF.
+  Run BladeID/Array Select  Fast triple-click while OFF. (Applicable installs only).
+  Restore Factory Defaults	Fast four-clicks while OFF, hold on last click. (If enabled).
+                              Release once announcement starts.
+  Enter/Exit VOLUME MENU    Hold and clash while OFF.
+    Volume up               Click while in VOLUME MENU, hilt pointing up.
+    Volume down             Click while in VOLUME MENU, hilt pointing down.
+                              Volume adjusts in increments per click.
+                              You must exit VOLUME MENU to resume using lightsaber normally.
+
+FUNCTIONS WITH BLADE ON
+  Blade lockup              Hold and hit clash while ON.
+  Blade tip drag            Hold and hit clash while ON pointing the blade tip down.
+  Play Character Quote      Fast double-click, hilt pointing up, plays sequentially. **
+  Force Effect              Fast double-click, hilt pointing down, plays randomly. **
+  Lightning block           Double click and hold while ON.
+  Melt                      Hold and thrust forward clash while ON. (Selected fonts only).
+  Blaster blocks            Short click/double click/triple click while ON.
+  Enter multi-blast mode    Hold while swinging for one second and release.
+                              To trigger blaster block, swing saber while in multi-blast mode.
+                              To exit, hold while swinging for one second and release.
+
+COLOUR CHANGE FUNCTIONS WITH BLADE ON
+  Enter COLOUR MENU         Fast triple-click while ON.
+                              Announcement confirms you are in the COLOUR MENU. 
+  Cycle to next colour      Rotate hilt whilst in COLOUR MENU until desired colour is reached.
+  Exit COLOUR MENU          Short click saves new colour and exits.
+                            Fast-double-click exits and reverts to original colour.
+                              Announcement confirms you are exiting COLOUR MENU.
+                              You must exit COLOUR MENU to resume using lightsaber normally.
+                              
+  *   = Gesture ignitions also available via defines.
+  **  = Audio player orientations can be reversed using SABERSENSE_FLIP_AUDIO_PLAYERS define.
+
+
+============================================================
+===================== 2 BUTTON CONTROLS ====================
+
+MAIN FUNCTIONS
+  Activate blade            Short click MAIN. *
+  Activate blade mute       Long click MAIN (hold for one second then release). 
+  Deactivate blade          Press and hold MAIN and wait until blade is off.
+
+FUNCTIONS WITH BLADE OFF
+  Next preset               Short click AUX, hilt pointing upwards.
+  Previous preset           Short click AUX, hilt pointing downwards.
+  Previous preset           Hold AUX and short click MAIN.
+                              (Duplicate legacy command).
+  Skip to first preset      Press and hold any button until it switches, hilt upwards.
+  Skip to middle preset     Press and hold any button  until it switches, hilt horizontal.
+  Skip to last preset       Press and hold any button  until it switches, hilt downwards.
+  Skip forward 5 presets    Fast double-click AUX, hilt pointing upwards.
+  Skip back 5 presets       Fast double-click AUX, hilt pointing downwards.
+  Skip forward 10 presets   Fast triple-click AUX, hilt pointing upwards.
+  Skip back 10 presets      Fast triple-click AUX, hilt pointing downwards.
+  Play Character Quote      Fast double-click MAIN, hilt pointing up, plays sequentially. **
+  Play Music Track          Fast double-click MAIN, pointing down. **
+  Speak battery voltage     Fast double-click-and-hold MAIN, or hold AUX for one second.
+  Run BladeID/Array Select  Fast triple-click. (Applicable installs only).
+  Restore Factory Defaults	Fast four-clicks MAIN, hold on last click. (If enabled).
+                              Release once announcement starts.
+  Enter/Exit VOLUME MENU    Hold MAIN then quickly click AUX and release both simultaneously.
+    Volume up               Click MAIN while in VOLUME MENU, hilt pointing up.
+    Volume down             Click MAIN while in VOLUME MENU, hilt pointing down, OR click 
+                              AUX while in VOLUME MENU.
+                              Volume adjusts in increments per click.
+                              You must exit VOLUME MENU to resume using saber normally.
+
+FUNCTIONS WITH BLADE ON
+  Blade lockup              Press and hold AUX.
+  Blade tip drag            Press and hold AUX while blade is pointing down.
+  Play Character Quote      Fast double-click MAIN, hilt pointing up, plays sequentially. **
+  Force Effect              Fast double-click MAIN, hilt pointing down, plays randomly. **
+  Lightning block           Double-click MAIN and hold.
+  Melt                      Hold MAIN and push blade tip against wall (stab).
+  Blaster blocks            Short click AUX. (Add Short click MAIN using define).
+  Enter multi-blast mode    Hold MAIN while swinging for one second and release. 
+                              Saber will play two quick blasts confirming mode.
+                              Swing blade to trigger blaster block.
+                              To exit multi-blast mode, fast single click AUX.
+
+COLOUR CHANGE FUNCTIONS WITH BLADE ON
+  Enter COLOUR MENU         Hold MAIN then quickly click AUX and release both
+                            buttons simultaneously. Or fast triple-click MAIN.
+                              Announcement confirms you are in the COLOUR MENU.
+  Cycle to next colour      Rotate hilt whilst in COLOUR MENU until desired colour is reached.
+                              Most Sabersense presets have 12 colour options.
+  Exit COLOUR MENU          Short click saves new colour and exits.
+                            Fast-double-click exits and reverts to original colour.
+                              Announcement confirms you are exiting COLOUR MENU.
+                              You must exit COLOUR MENU to resume using lightsaber normally.
+                              
+  *  = Gesture ignitions also available via defines.
+  ** = Audio player orientations can be reversed using SABERSENSE_FLIP_AUDIO_PLAYERS define.
+
+
+===========================================================
+=================== SABERSENSE™ DEFINES ===================
+
+#define SABERSENSE_BLADE_ID
+  Replaces regular BladeID with on-demand BladeID scanning. 
+  Plays array-specific bladeidX.wav files when switching.
+  
+#define SABERSENSE_ARRAY_SELECTOR
+  Replaces regular BladeID and allows cycling between 
+  different blade/preset arrays manually, regardless 
+  of actual BladeID status. Plays array-specific 
+  arrayX.wav files when switching.
+  
+#define SABERSENSE_ENABLE_ARRAY_FONT_IDENT
+  Plays font ident after array ident when switching arrays.
+  Use with SABERSENSE_BLADE_ID and SABERSENSE_ARRAY_SELECTOR.
+
+#define SABERSENSE_FLIP_AUDIO_PLAYERS
+  Reverses UP/DOWN orientation for playing FORCE, QUOTE and 
+  MUSIC TRACK audio files. Default (no define) is UP for 
+  sequential QUOTE with blade ON and OFF, and DOWN for random 
+  FORCE effect (ON) and music TRACK (OFF). Define acts on 
+  both ON and OFF states for consistency.
+  
+#define SABERSENSE_BLAST_MAIN_AND_AUX
+  Adds blaster block button to MAIN as well as AUX in 
+  2-button mode. Improves 1 and 2-button harmonization, 
+  but makes accidental blasts more likely when double-clicking 
+  for Quotes or Force Effect.
+  
+#define SABERSENSE_BUTTON_CLICKER
+  Button Clicker to play press/release wav files when 
+  buttons are pressed. Intended for Scavenger hilt 
+  where wheel makes tactile feel difficult.
+  Requires press.wav and release.wav files to work.
+  
+#define SABERSENSE_ENABLE_RESET
+  Enables system to be completely reset to 'factory defaults' 
+  i.e. original config, using button press to delete 
+  all save files.
+  
+#define SABERSENSE_OS7_LEGACY_SUPPORT
+  Required when using this prop file with ProffieOS 7.x. 
+  Not needed (will not work) with ProffieOS 8.x or later.
+  
+#define SABERSENSE_NO_COLOR_CHANGE
+  Use instead of DISABLE_COLOR_CHANGE.
+  
+#define SABERSENSE_NO_LOCKUP_HOLD
+  Applicable to two-button mode only, reverts to lockup being 
+  triggered by clash while holding aux.
+
+GESTURE CONTROLS
+  There are four gesture types: Twist, Stab, Swing and Thrust.
+  Gesture controls bypass all preon effects.
+  #define SABERSENSE_TWIST_ON
+  #define SABERSENSE_TWIST_OFF
+  #define SABERSENSE_STAB_ON
+  #define SABERSENSE_SWING_ON
+  #define SABERSENSE_THRUST_ON
+
+============================================================
+============================================================
+*/
+#ifndef PROPS_SABER_SABERSENSE_BUTTONS_H
+#define PROPS_SABER_SABERSENSE_BUTTONS_H
+ 
+ 
+#include "prop_base.h"
+#include "../sound/hybrid_font.h"
+
+#undef PROP_TYPE
+#define PROP_TYPE SabersenseButtons
+ 
+#ifndef MOTION_TIMEOUT
+#define MOTION_TIMEOUT 60 * 15 * 1000
+#endif
+ 
+#ifndef SABERSENSE_SWING_ON_SPEED
+#define SABERSENSE_SWING_ON_SPEED 250
+#endif
+ 
+#ifndef SABERSENSE_LOCKUP_DELAY
+#define SABERSENSE_LOCKUP_DELAY 200
+#endif
+ 
+#ifndef SABERSENSE_FORCE_PUSH_LENGTH
+#define SABERSENSE_FORCE_PUSH_LENGTH 5
+#endif
+ 
+#ifndef BUTTON_DOUBLE_CLICK_TIMEOUT
+#define BUTTON_DOUBLE_CLICK_TIMEOUT 300
+#endif
+ 
+#ifndef BUTTON_SHORT_CLICK_TIMEOUT
+#define BUTTON_SHORT_CLICK_TIMEOUT 300
+#endif
+ 
+#ifndef BUTTON_HELD_TIMEOUT
+#define BUTTON_HELD_TIMEOUT 300
+#endif
+ 
+#ifndef BUTTON_HELD_MEDIUM_TIMEOUT
+#define BUTTON_HELD_MEDIUM_TIMEOUT 1000
+#endif
+ 
+#ifndef BUTTON_HELD_LONG_TIMEOUT
+#define BUTTON_HELD_LONG_TIMEOUT 2000
+#endif
+ 
+#ifdef SABERSENSE_SWING_ON
+#define SABERSENSE_SWING_GESTURE
+#endif
+ 
+#ifdef SABERSENSE_STAB_ON
+#define SABERSENSE_STAB_GESTURE
+#endif
+ 
+#ifdef SABERSENSE_TWIST_ON
+#define SABERSENSE_TWIST_GESTURE
+#endif
+ 
+#ifdef SABERSENSE_THRUST_ON
+#define SABERSENSE_THRUST_GESTURE
+#endif
+ 
+EFFECT(dim);      // for EFFECT_POWERSAVE
+EFFECT(battery);  // for EFFECT_BATTERY_LEVEL
+EFFECT(vmbegin);  // for Begin Volume Menu
+EFFECT(vmend);    // for End Volume Menu
+EFFECT(volup);    // for increse volume
+EFFECT(voldown);  // for decrease volume
+EFFECT(volmin);   // for minimum volume reached
+EFFECT(volmax);   // for maximum volume reached
+EFFECT(faston);   // for EFFECT_FAST_ON
+EFFECT(blstbgn);  // for Begin Multi-Blast
+EFFECT(blstend);  // for End Multi-Blast
+EFFECT(array);    // for playing array idents
+EFFECT(bladeid);  // for playing bladeid idents
+EFFECT(reset);    // for playing factory default reset completed
+#ifdef SABERSENSE_OS7_LEGACY_SUPPORT
+EFFECT(quote);    // for playing quotes. Required for ProffieOS 7.x.
+#endif
+
+
+  // The Saber class implements the basic states and actions
+  // for the saber.
+class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
+public:
+  SabersenseButtons() : PropBase() {}
+  const char* name() override { return "SabersenseButtons"; }
+
+
+  // MAIN LOOP
+  void Loop() override {
+    PropBase::Loop();
+    DetectTwist();
+    Vec3 mss = fusor.mss();
+    if (SaberBase::IsOn()) {
+      DetectSwing();
+      if (auto_lockup_on_ &&
+        !swinging_ &&
+        fusor.swing_speed() > 120 &&
+        millis() - clash_impact_millis_ > SABERSENSE_LOCKUP_DELAY &&
+          SaberBase::Lockup()) {
+          SaberBase::DoEndLockup();
+          SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+        auto_lockup_on_ = false;
+      }
+      if (auto_melt_on_ &&
+        !swinging_ &&
+        fusor.swing_speed() > 60 &&
+        millis() - clash_impact_millis_ > SABERSENSE_LOCKUP_DELAY &&
+          SaberBase::Lockup()) {
+          SaberBase::DoEndLockup();
+          SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+        auto_melt_on_ = false;
+      }
+
+      // EVENT_PUSH
+      if (fabs(mss.x) < 3.0 &&
+        mss.y * mss.y + mss.z * mss.z > 70 &&
+        fusor.swing_speed() < 30 &&
+        fabs(fusor.gyro().x) < 10) {
+          if (millis() - push_begin_millis_ > SABERSENSE_FORCE_PUSH_LENGTH) {
+            Event(BUTTON_NONE, EVENT_PUSH);
+            push_begin_millis_ = millis();
+            }
+          } else {
+            push_begin_millis_ = millis();
+          }
+      } else {
+        // EVENT_SWING - Swing On gesture control to allow fine tuning of speed needed to ignite
+        if (millis() - saber_off_time_ < MOTION_TIMEOUT) {
+          SaberBase::RequestMotion();
+            if (swinging_ && fusor.swing_speed() < 90) {
+              swinging_ = false;
+            }
+            if (!swinging_ && fusor.swing_speed() > SABERSENSE_SWING_ON_SPEED) {
+              swinging_ = true;
+            Event(BUTTON_NONE, EVENT_SWING);
+            }
+        }
+      // EVENT_THRUST
+        if (mss.y * mss.y + mss.z * mss.z < 16.0 &&
+          mss.x > 14  &&
+          fusor.swing_speed() < 150) {
+          if (millis() - thrust_begin_millis_ > 15) {
+            Event(BUTTON_NONE, EVENT_THRUST);
+            thrust_begin_millis_ = millis();
+          }
+          } else {
+          thrust_begin_millis_ = millis();
+        }
+      }
+      //  Enables sequential sounds and processes on array handling.
+#ifdef SABERSENSE_BLADE_ID
+        if (do_font_after_sound_ && !IsBladeidSoundPlaying()) {
+        SaberBase::DoNewFont();
+        do_font_after_sound_ = !do_font_after_sound_;
+        }
+#endif
+
+#ifdef SABERSENSE_ARRAY_SELECTOR
+        if (do_font_after_sound_ && !IsArraySoundPlaying()) {
+        SaberBase::DoNewFont();
+        do_font_after_sound_ = !do_font_after_sound_;
+        }
+#endif
+    }
+
+
+  // VOLUME MENU
+  void VolumeUp() {
+    STDOUT.println("Volume up");
+    if (dynamic_mixer.get_volume() < VOLUME) {
+      dynamic_mixer.set_volume(std::min<int>(VOLUME + VOLUME * 0.1,
+        dynamic_mixer.get_volume() + VOLUME * 0.10));
+      if (SFX_volup) {
+        hybrid_font.PlayCommon(&SFX_volup);
+      } else {
+        beeper.Beep(0.5, 2000);
+      }
+      STDOUT.print("Volume Up - Current Volume: ");
+      STDOUT.println(dynamic_mixer.get_volume());
+    } else {
+      // Cycle through ends of Volume Menu option
+      #ifdef VOLUME_MENU_CYCLE
+        if (!max_vol_reached_) {
+          if (SFX_volmax) {
+            hybrid_font.PlayCommon(&SFX_volmax);
+          } else {
+            beeper.Beep(0.5, 3000);
+          }
+          STDOUT.print("Maximum Volume: ");
+          max_vol_reached_ = true;
+        } else {
+          dynamic_mixer.set_volume(std::max<int>(VOLUME * 0.1,
+          dynamic_mixer.get_volume() - VOLUME * 0.90));
+          if (SFX_volmin) {
+            hybrid_font.PlayCommon(&SFX_volmin);
+          } else {
+            beeper.Beep(0.5, 1000);
+          }
+          STDOUT.print("Minimum Volume: ");
+          max_vol_reached_ = false;
+        }
+      #else
+        if (SFX_volmax) {
+          hybrid_font.PlayCommon(&SFX_volmax);
+        } else {
+          beeper.Beep(0.5, 3000);
+        }
+        STDOUT.print("Maximum Volume: ");
+      #endif
+    }
+  }
+
+  void VolumeDown() {
+    STDOUT.println("Volume Down");
+    if (dynamic_mixer.get_volume() > (0.10 * VOLUME)) {
+      dynamic_mixer.set_volume(std::max<int>(VOLUME * 0.1,
+        dynamic_mixer.get_volume() - VOLUME * 0.10));
+      if (SFX_voldown) {
+        hybrid_font.PlayCommon(&SFX_voldown);
+      } else {
+        beeper.Beep(0.5, 2000);
+      }
+      STDOUT.print("Volume Down - Current Volume: ");
+      STDOUT.println(dynamic_mixer.get_volume());
+    } else {
+      #ifdef VOLUME_MENU_CYCLE
+        if (!min_vol_reached_) {
+          if (SFX_volmin) {
+            hybrid_font.PlayCommon(&SFX_volmin);
+          } else {
+            beeper.Beep(0.5, 1000);
+          }
+          STDOUT.print("Minimum Volume: ");
+          min_vol_reached_ = true;
+        } else {
+          dynamic_mixer.set_volume(VOLUME);
+          if (SFX_volmax) {
+            hybrid_font.PlayCommon(&SFX_volmax);
+          } else {
+            beeper.Beep(0.5, 3000);
+          }
+          STDOUT.print("Maximum Volume: ");
+          min_vol_reached_ = false;
+        }
+      #else
+        if (SFX_volmin) {
+          hybrid_font.PlayCommon(&SFX_volmin);
+        } else {
+          beeper.Beep(0.5, 1000);
+        }
+        STDOUT.print("Minimum Volume: ");
+      #endif
+ 
+    }
+  }
+
+
+  //  BLADE ID OPTIONS AND ARRAY MANAGEMENT
+  //  True BladeID, runs scan on-demand with unique 'bladeid' ident sound.
+#ifdef SABERSENSE_BLADE_ID
+  #ifndef ENABLE_POWER_FOR_ID    
+    #error "SABERSENSE_BLADE_ID requires ENABLE_POWER_FOR_ID to be defined."
+  #endif
+
+    bool IsBladeidSoundPlaying() {
+      return !!GetWavPlayerPlaying(&SFX_bladeid);
+    }
+#ifdef SABERSENSE_ENABLE_ARRAY_FONT_IDENT  //  Plays 'bladeid' sound AND 'font' sound.
+    void TriggerBladeID() {
+      FindBladeAgain();
+        SFX_bladeid.Select(current_config - blades);
+        hybrid_font.PlayCommon(&SFX_bladeid);
+          //  Calls Loop function to handle waiting for effect before running DoNewFont.
+          do_font_after_sound_ = !do_font_after_sound_;
+        }
+#else
+    //  Plays 'bladeid' sound only, or 'font' sound if no 'bladeid' sound available.
+    void TriggerBladeID() {
+      FindBladeAgain();
+        if (SFX_bladeid) {
+          SFX_bladeid.Select(current_config - blades);
+          hybrid_font.PlayCommon(&SFX_bladeid);  //  Play 'bladeid' sound file if available.
+        } else {
+        SaberBase::DoNewFont();  //  Play font ident if 'bladeid' sound file missing.
+        }
+#endif
+#endif
+
+
+    //  Manual Array Selector, switches on-demand to next array, plays 'array' ident sound. 
+#ifdef SABERSENSE_ARRAY_SELECTOR
+  #ifdef SABERSENSE_BLADE_ID    //  Only one Sabersense BladeID standard permitted.
+    #error "SABERSENSE_ARRAY_SELECTOR and SABERSENSE_BLADE_ID cannot be defined at the same time."
+  #endif
+    bool IsArraySoundPlaying() {
+     return !!GetWavPlayerPlaying(&SFX_array);
+    } 
+#ifdef SABERSENSE_ENABLE_ARRAY_FONT_IDENT  //  Plays 'array' sound AND 'font' sound.
+    void NextBladeArray() {
+      FakeFindBladeAgain();
+        SFX_array.Select(current_config - blades);
+        hybrid_font.PlayCommon(&SFX_array);
+          //  Calls Loop function to handle waiting for effect before running DoNewFont.
+          do_font_after_sound_ = !do_font_after_sound_;
+        }
+#else
+    //  Plays 'array' sound only, or 'font' sound if no 'array' sound available.
+    void NextBladeArray() {
+      FakeFindBladeAgain();
+        if (SFX_array) {
+          SFX_array.Select(current_config - blades);
+          hybrid_font.PlayCommon(&SFX_array);  //  Play 'array' sound file if available.
+        } else {
+        SaberBase::DoNewFont();  //  Play font ident if 'array' sound file missing.
+        }
+    }
+#endif
+
+  // Manual Blade Array Selection version of FindBladeAgain()
+#ifdef SABERSENSE_OS7_LEGACY_SUPPORT
+#undef ACTIVATE
+#define ACTIVATE(N) do {                        \
+    if (!current_config->blade##N) {            \
+      goto bad_blade;                           \
+    }                                           \
+    current_config->blade##N->Activate();       \
+  } while(0);
+#else
+#undef ACTIVATE
+#define ACTIVATE(N) do {                        \
+    if (!current_config->blade##N) {            \
+      goto bad_blade;                           \
+    }                                           \
+    current_config->blade##N->Activate(N);      \
+  } while(0);
+#endif
+
+  void FakeFindBladeAgain() {
+    // Reverse everything FindBladeAgain does, except for recalculating best_config.
+    ONCEPERBLADE(UNSET_BLADE_STYLE)
+#undef DEACTIVATE
+#define DEACTIVATE(N) do {                      \
+    if (current_config->blade##N)               \
+      current_config->blade##N->Deactivate();   \
+  } while(0);
+    ONCEPERBLADE(DEACTIVATE);
+    SaveVolumeIfNeeded();
+      if (fusor.angle1() > 0) {
+        //  Cycle forwards (next array) if hilt pointing up...
+        current_config = blades + (current_config - blades + 1) % NELEM(blades);
+      } else {
+        //  Cycle backwards (previous array) if hilt pointing down.
+        current_config = blades + (current_config - blades - 1) % NELEM(blades);
+      }   
+
+    ONCEPERBLADE(ACTIVATE);
+    RestoreGlobalState();
+#ifdef SAVE_PRESET
+    ResumePreset();
+#else
+    SetPreset(0, false);
+#endif // SAVE_PRESET
+//    PVLOG_NORMAL << "** FakeFindBladeAgain() Completed\n";
+    return;    
+#if NUM_BLADES != 0
+  bad_blade:
+    ProffieOSErrors::error_in_blade_array();
+#endif
+  }
+#endif 
+
+ 
+  //  RESET FACTORY DEFAULTS (Delete Save Files).
+  //  Script to determine if sound effect has finished.
+#ifdef SABERSENSE_ENABLE_RESET
+    bool IsResetSoundPlaying() {
+     return !!GetWavPlayerPlaying(&SFX_reset);
+    }
+#endif
+
+
+    //  CLICK PLAYER FOR BUTTON PRESSES (optional).
+    void PlaySound(const char* sound) {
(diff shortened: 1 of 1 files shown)


props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+      if (SaberBase::Lockup()) {
+        SaberBase::DoEndLockup();
+        SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+        return true;
+        }
+      }
+      return false;
+    }
+ 
+#ifdef SABERSENSE_OS7_LEGACY_SUPPORT
```
profezzorn, line 1225: Please write this as:

#ifdef  SABERSENSE_OS7_LEGACY_SUPPORT
  void SB_Effect(.....)
#else
  void SB_Effect(.........)
#endif
{
  switch (effect) {
#ifdef SABERSENSE_OS7_LEGACY_SUPPORT
      case EFFECT_QUOTE: ..... break;
#endif


By doing it this way, we don't duplicate the { in the if/else branches, which tends to confuse editors (and readers)

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+      FindBladeAgain();
+      SaberBase::DoBladeDetect(false);
+    return true;
+#endif
+ 
+ 
+    // Events that needs to be handled regardless of what other buttons are pressed.
+    case EVENTID(BUTTON_AUX2, EVENT_RELEASED, MODE_ANY_BUTTON | MODE_ON):
+      if (SaberBase::Lockup()) {
+        SaberBase::DoEndLockup();
```
profezzorn, line 1217: fix indentation

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+      SaberBase::DoBladeDetect(true);
+    return true;
+ 
+    case EVENTID(BUTTON_BLADE_DETECT, EVENT_LATCH_OFF, MODE_ANY_BUTTON | MODE_ON):
+    case EVENTID(BUTTON_BLADE_DETECT, EVENT_LATCH_OFF, MODE_ANY_BUTTON | MODE_OFF):
+      //  Might need to do something cleaner, but let's try this for now.
+      blade_detected_ = false;
+      FindBladeAgain();
+      SaberBase::DoBladeDetect(false);
+    return true;
```
profezzorn, line 1210: fix indentation

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+                }
+              ++dirIterator;  // Move to the next entry in the root directory.
+            }
+      // Unlock SD card after deletion is complete.
+      LOCK_SD(false);
+
+      if (SFX_reset) {  //  Optional confirmation sound file 'reset'.
+        hybrid_font.PlayCommon(&SFX_reset);
+          while(IsResetSoundPlaying());  //  Lock system while sound finishes...
+        STM32.reset(); //  ...then reboot saber.
```
profezzorn, line 1175: STM.reset() is the same in both if and else branches, so it can be moved out (and the else branch can then be removed)

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+        for (const char* targetFile : filesToDelete) {
+          if (LSFS::Exists(targetFile)) {
+            LSFS::Remove(targetFile);
+              Serial.print("Deleted from root: ");
+              Serial.println(targetFile);
+              }
+           }
+        // Find all immediate subdirectories of the root and delete files.
+        LSFS::Iterator dirIterator("/");  // Iterator for the root directory.
+          while (dirIterator) {
```
profezzorn, line 1147: All save directories are listed in the blades[] array, so it would be better to iterate over that instead of iterating over all directories on the SD card.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+#endif
+      if (accel_.x < -0.15) {
+        SaberBase::SetLockup(SaberBase::LOCKUP_DRAG);
+      } else {
+        SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
+      }
+        swing_blast_ = false;
+        SaberBase::DoBeginLockup();
+        return true;
+      break;
```
profezzorn, line 1072: no need for break after return

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+        SaberBase::DoBlast();
+      }
+      return true;
+
+
+    //  LOCKUP
+#if NUM_BUTTONS == 1
+    //  1 button lockup
+    case EVENTID(BUTTON_NONE, EVENT_CLASH, MODE_ON | BUTTON_POWER):
+#else
```
profezzorn, line 1056: use #elif defined(SANERSEMSE_NO_LOCKUP_HOLD)

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+#ifdef SABERSENSE_BLAST_MAIN_AND_AUX
+      //  For harmonized 'Exit Colour Menu' control in OS-7.x. Ignored in OS-8.
+      if (SaberBase::GetColorChangeMode() != SaberBase::COLOR_CHANGE_MODE_NONE) {
+        ToggleColorChangeMode();          
+      } else {  
+        //  Fires blast in all OSs.
+        swing_blast_ = false;
+        SaberBase::DoBlast();
+      }
+    return true;
```
profezzorn, line 1016: return should be on the same indentation level as previous line

profezzorn (commented):
first pass...

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+#ifndef BUTTON_HELD_MEDIUM_TIMEOUT
+#define BUTTON_HELD_MEDIUM_TIMEOUT 1000
+#endif
+ 
+#ifndef BUTTON_HELD_LONG_TIMEOUT
+#define BUTTON_HELD_LONG_TIMEOUT 2000
+#endif
+ 
+#ifdef SABERSENSE_SWING_ON
+#define SABERSENSE_SWING_GESTURE
```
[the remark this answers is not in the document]
Sabersense-CC: Ermm, I'm not sure. That bit was essentially copied from the SA22C prop. But I'll run some tests without that section to see if it affects anything - it sounds like it won't - in which case I'll delete it.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+EFFECT(voldown);  // for decrease volume
+EFFECT(volmin);   // for minimum volume reached
+EFFECT(volmax);   // for maximum volume reached
+EFFECT(faston);   // for EFFECT_FAST_ON
+EFFECT(blstbgn);  // for Begin Multi-Blast
+EFFECT(blstend);  // for End Multi-Blast
+EFFECT(array);    // for playing array idents
+EFFECT(bladeid);  // for playing bladeid idents
+EFFECT(reset);    // for playing factory default reset completed
+#ifdef SABERSENSE_OS7_LEGACY_SUPPORT
```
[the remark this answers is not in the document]
Sabersense-CC: Apologies, I'm a little confused on that. Do you mean EFFECT_QUOTE needs to be added to the sound/effect.h file? If so, I thought OS-8/Github Master already had that, and the only reason I need to mention it in my prop is to make my prop work with OS-7.x (as EFFECT_QUOTE isn't included in OS-7's sound/effect.h file). What am I missing here?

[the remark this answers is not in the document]
profezzorn: I mean, we could add

#define HAVE_EFFECT_QUOTE to effect.h

Then you don't have to have a special define that the user has to add to the their config file for it to work, it would just work.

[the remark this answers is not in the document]
Sabersense-CC: But wouldn't that mean that anyone trying to use this prop file with OS-7 would need to edit their own sound/effect.h file for it to work (which obviously we can't do for them as OS-7 is already 'out there')? My understanding is that "quote" is already included in OS-8, and so doesn't need listing in the prop file like this - it only needs listing in the prop when using quotes with this prop under OS-7, doesn't it?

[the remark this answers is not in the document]
profezzorn: No.

effect.h has #define HAVE_EFFECT_QUOTE in it (OS8 only)
Then here, we could just do:

#ifndef HAVE_EFFECT_QUOTE
EFFECT(quote);
#endif

This means that if HAVE_EFFECT_QUOTE *isn't* there, then it gets added here.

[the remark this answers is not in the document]
Sabersense-CC: Aaahhhhh - sorry - the penny has finally dropped! LOL! I'm with you. Will get that sorted. :)

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Now deleted to no ill effect as far as I can tell. Not sure what they were for under SA22C, but gone now.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Sorry that took a while for me to grasp. Now sorted.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+#endif
+
+
+  //  The Saber class implements the basic states and actions
+  //  for the saber.
+class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
+public:
+  SabersenseButtons() : PropBase() {}
+  const char* name() override { return "SabersenseButtons"; }
+
```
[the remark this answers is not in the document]
Sabersense-CC: All instances removed.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+          SaberBase::RequestMotion();
+            if (swinging_ && fusor.swing_speed() < 90) {
+              swinging_ = false;
+            }
+            if (!swinging_ && fusor.swing_speed() > SABERSENSE_SWING_ON_SPEED) {
+              swinging_ = true;
+            Event(BUTTON_NONE, EVENT_SWING);
+            }
+        }
+      //  EVENT_THRUST
```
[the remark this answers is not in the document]
Sabersense-CC: Still getting my head round the finer points of indenting best practice, but hopefully it's all correct now.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+        if (mss.y * mss.y + mss.z * mss.z < 16.0 &&
+          mss.x > 14  &&
+          fusor.swing_speed() < 150) {
+          if (millis() - thrust_begin_millis_ > 15) {
+            Event(BUTTON_NONE, EVENT_THRUST);
+            thrust_begin_millis_ = millis();
+          }
+          } else {
```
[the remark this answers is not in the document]
Sabersense-CC: Hopefully correct now.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+          thrust_begin_millis_ = millis();
+        }
+      }
+      //  Enables sequential sounds and processes on array handling.
+#ifdef SABERSENSE_BLADE_ID
+        if (do_font_after_sound_ && !IsBladeidSoundPlaying()) {
```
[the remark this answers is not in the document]
Sabersense-CC: Hopefully correct now.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1275 @@
...
+          dynamic_mixer.get_volume() - VOLUME * 0.90));
+          if (SFX_volmin) {
+            hybrid_font.PlayCommon(&SFX_volmin);
+          } else {
+            beeper.Beep(0.5, 1000);
+          }
+          STDOUT.print("Minimum Volume: ");
+          max_vol_reached_ = false;
+        }
+      #else
```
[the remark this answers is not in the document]
Sabersense-CC: Fixed.

props/saber_sabersense_buttons.h
Sabersense-CC: Done.

props/saber_sabersense_buttons.h
Sabersense-CC: Hopefully correct now.

props/saber_sabersense_buttons.h
Sabersense-CC: Hopefully correct now.

props/saber_sabersense_buttons.h
Sabersense-CC: In the course of fixing this I found that things were'nt very user-friendly without the reset.wav file. So I've added a confirmation beep where no reset file is available which should hopefully make things work better.

props/saber_sabersense_buttons.h
Sabersense-CC: Done.

props/saber_sabersense_buttons.h
Sabersense-CC: Three instances deleted.

props/saber_sabersense_buttons.h
Sabersense-CC: Not used #elif before so I think I've done this right.

props/saber_sabersense_buttons.h
Sabersense-CC: Fixed.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Doh! I need to look at this again. It works under OS-7 but throws an error under OS-8. Also I can't see HAVE_EFFECT_QUOTE anywhere in sound/effect.h in GitHub Master. If we're proposing to add it, that feels like more work than just using the SABERSENSE_OS7_LEGACY_SUPPORT define in my prop?

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1235 @@
...
+This prop file references certain custom sound files 
+to aid in saber function navigation. These sound files 
+are optional and are available as a free download from:
+https://sabersense.square.site/downloads
+
+
+============================================================
+========= BUTTON SYSTEM PRINCIPLES AND LOGIC NOTES ========= 
+
+The Sabersense™ button control system has been engineered
```
profezzorn, line 19: C++ files are generally supposed to be plain ASCII.
To avoid problems either remove the TM character, or replace it with "(TM)"

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1235 @@
...
+  #define SABERSENSE_STAB_ON
+  #define SABERSENSE_SWING_ON
+  #define SABERSENSE_THRUST_ON
+
+============================================================
+============================================================
+*/
+#ifndef PROPS_SABER_SABERSENSE_BUTTONS_H
+#define PROPS_SABER_SABERSENSE_BUTTONS_H
+
```
profezzorn, line 257: extra empty line (there is never a need for three newlines in a row)

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
profezzorn: Adding HAVE_EFFECT_QUOTE to sound/effect.h is no harder than adding a line to this file.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1235 @@
...
+      //  EVENT_PUSH
+      if (fabs(mss.x) < 3.0 &&
+        mss.y * mss.y + mss.z * mss.z > 70 &&
+        fusor.swing_speed() < 30 &&
+        fabs(fusor.gyro().x) < 10) {
+          if (millis() - push_begin_millis_ > SABERSENSE_FORCE_PUSH_LENGTH) {
+            Event(BUTTON_NONE, EVENT_PUSH);
+            push_begin_millis_ = millis();
+            }
```
profezzorn, line 359: indentation doesn't like up here

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1235 @@
...
+      } else {
+        //  EVENT_SWING - Swing On gesture control to allow fine tuning of speed needed to ignite
+        if (millis() - saber_off_time_ < MOTION_TIMEOUT) {
+          SaberBase::RequestMotion();
+            if (swinging_ && fusor.swing_speed() < 90) {
+              swinging_ = false;
+            }
+            if (!swinging_ && fusor.swing_speed() > SABERSENSE_SWING_ON_SPEED) {
+              swinging_ = true;
+            Event(BUTTON_NONE, EVENT_SWING);
```
profezzorn, line 372: indentation

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1235 @@
...
+            }
+        }
+        //  EVENT_THRUST
+        if (mss.y * mss.y + mss.z * mss.z < 16.0 &&
+          mss.x > 14  &&
+          fusor.swing_speed() < 150) {
+          if (millis() - thrust_begin_millis_ > 15) {
+            Event(BUTTON_NONE, EVENT_THRUST);
+            thrust_begin_millis_ = millis();
+            }
```
profezzorn, line 382: indentation

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
profezzorn: hard to tell, because there are earlier problems with indentation

profezzorn (commented):
There are lots and lots and lots of indentation issues, which makes the code hard to read.
I suggest finding something that can format the whole code for you, set it to two spaces and re-format the whole code.

props/saber_sabersense_buttons.h
Sabersense-CC: I fear the indenting thing may be the thing that defeats this endeavour. I've tried running the code through a couple of different online formatters, but they all seem to do something slightly different with it, and I can't tell which is correct. They all seem also to do things that look wrong to me, but given that my formatting is wrong, maybe they are all correct, even with their respective differences. I've tried to read up on it and tried comparing with other ProffieOS prop files, but I just can't seem to find definitive correct methods. Anyway I've posted a version that, according to the reformatter, is correct. I've also done the bits I CAN do, like losing the TM and sorting double line spaces. I still need to sort out HAVE_EFFECT_QUOTE as I was kind of under the impression that the only file I could change was the prop file so that I wouldn't mess with anything else in the OS. But I'll have a crack at making that work. Although of course if the indenting does indeed prove to be the final impasse, HAVE_EFFECT_QUOTE becomes redundant.

profezzorn: Unfortunately github took away the spaces...

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Have messaged you in more detail, but for now I've removed the entire OS-7 legacy thing, as I couldn't get it to work the way you suggested. I was also, in truth, slightly nervous about messing with other files in ProffieOS.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+  const char* name() override { return "SabersenseButtons"; }
+
+  void Loop() override {
+    PropBase::Loop();
+    DetectTwist();
+    Vec3 mss = fusor.mss();
+    if (SaberBase::IsOn()) {
+      DetectSwing();
+      if (auto_lockup_on_ &&
+        !swinging_ &&
```
profezzorn, line 328: This line should be indented 4 steps from the if, because it's a continued statement.
As is, it forms a straight line all the way to 334, which makes it difficult to see that lines 328-331 make up the IF and lines 332-334 is the code that will be executed if the IF is true.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+    }
+#endif
+  }
+
+  // VOLUME MENU
+  void VolumeUp() {
+    STDOUT.println("Volume up");
+    if (dynamic_mixer.get_volume() < VOLUME) {
+      dynamic_mixer.set_volume(std::min<int>(VOLUME + VOLUME * 0.1,
+        dynamic_mixer.get_volume() + VOLUME * 0.10));
```
profezzorn, line 403: continued statemene, +4 spaces (not 2)

profezzorn (commented):
Took a quick look, just focusing on indentation.
It look *much* better now.
There are a few cases that should still be fixed, but indentation-wise it's pretty much there.
I will take another look tomorrow to review the actual content, which is much easier now that the indentation makes sense.

props/saber_sabersense_buttons.h
Sabersense-CC: Fixed.

props/saber_sabersense_buttons.h
Sabersense-CC: Understood. Now fixed.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+
+  *   = Gesture ignitions also available via defines.
+  **  = Audio player orientations can be reversed using SABERSENSE_FLIP_AUDIO_PLAYERS define.
+  *** = Feature must be enabled in config file using SABERSENSE_ENABLE_RESET define.
+
+===========================================================
+=================== SABERSENSE DEFINES ====================
+
+#define SABERSENSE_BLADE_ID
+  Replaces regular BladeID with on-demand BladeID scanning.
```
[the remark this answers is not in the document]
Sabersense-CC: It's intended so the system doesn't need to have constant bladeID scanning running. The idea is you fit your blade, hit the button (in this case three fast clicks) and the system then runs the scan, identifies the blade and switches to its associated array. It puts the user in the driving seat and also removes the risk of spurious ID readings switching blade when you don't want it to, especially if using SnapshotID.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+    // Enables sequential sounds and processes on array handling.
+#ifdef SABERSENSE_BLADE_ID
+    if (do_font_after_sound_ && !IsBladeidSoundPlaying()) {
+      SaberBase::DoNewFont();
+      do_font_after_sound_ = !do_font_after_sound_;
+    }
+#endif
+
+#ifdef SABERSENSE_ARRAY_SELECTOR
+    if (do_font_after_sound_ && !IsArraySoundPlaying()) {
```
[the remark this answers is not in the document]
Sabersense-CC: Do you mean what happens if you have SABERSENSE_ARRAY_SELECTOR and SABERSENSE_BLADE_ID defined at the same time? The prop won't allow it, and gives an error at compile if you try. The error message is on line 523.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+      if (SFX_volup) {
+        hybrid_font.PlayCommon(&SFX_volup);
+      } else {
+        beeper.Beep(0.5, 2000);
+      }
+      STDOUT.print("Volume Up - Current Volume: ");
+      STDOUT.println(dynamic_mixer.get_volume());
+    } else {
+      // Cycle through ends of Volume Menu option
+#ifdef VOLUME_MENU_CYCLE
```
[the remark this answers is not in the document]
Sabersense-CC: Will fix.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+#ifdef SABERSENSE_BLADE_ID
+#ifndef ENABLE_POWER_FOR_ID
+#error "SABERSENSE_BLADE_ID requires ENABLE_POWER_FOR_ID to be defined."
+#endif
+
+  bool IsBladeidSoundPlaying() {
+    return !!GetWavPlayerPlaying(&SFX_bladeid);
+  }
+
+#ifdef SABERSENSE_ENABLE_ARRAY_FONT_IDENT  // Plays 'bladeid' sound AND 'font' sound.
```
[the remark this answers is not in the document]
Sabersense-CC: Will fix.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+  void TriggerBladeID() {
+    FindBladeAgain();
+    SFX_bladeid.Select(current_config - blades);
+    hybrid_font.PlayCommon(&SFX_bladeid);
+    // Calls Loop function to handle waiting for effect before running DoNewFont.
+    do_font_after_sound_ = !do_font_after_sound_;
```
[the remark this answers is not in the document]
Sabersense-CC: I had a bit of help with that bit so I'll need to look into it in case there was a legit reason that I can't remember. Will report back...

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
profezzorn: Don't tell me, tell the person reading this file.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+  // Manual Array Selector, switches on-demand to next array, plays 'array' ident sound.
+#ifdef SABERSENSE_ARRAY_SELECTOR
+#ifdef SABERSENSE_BLADE_ID    // Only one Sabersense BladeID standard permitted.
+#error "SABERSENSE_ARRAY_SELECTOR and SABERSENSE_BLADE_ID cannot be defined at the same time."
+#endif
+  bool IsArraySoundPlaying() {
+    return !!GetWavPlayerPlaying(&SFX_array);
+  }
+
+#ifdef SABERSENSE_ENABLE_ARRAY_FONT_IDENT  // Plays 'array' sound AND 'font' sound.
```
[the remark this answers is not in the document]
Sabersense-CC: Will fix.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+  void NextBladeArray() {
+    FakeFindBladeAgain();
+    SFX_array.Select(current_config - blades);
+    hybrid_font.PlayCommon(&SFX_array);
+    // Calls Loop function to handle waiting for effect before running DoNewFont.
+    do_font_after_sound_ = !do_font_after_sound_;
```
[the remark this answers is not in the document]
Sabersense-CC: Will fix.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+    current_config->blade##N->Deactivate();   \
+} while(0);
+  ONCEPERBLADE(DEACTIVATE);
+  SaveVolumeIfNeeded();
+  if (fusor.angle1() > 0) {
+    // Cycle forwards (next array) if hilt pointing up...
+    current_config = blades + (current_config - blades + 1) % NELEM(blades);
+  } else {
+    // Cycle backwards (previous array) if hilt pointing down.
+    current_config = blades + (current_config - blades - 1) % NELEM(blades);
```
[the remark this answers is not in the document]
Sabersense-CC: I think I'm going remove that reverse completely. I added it to harmonize the principle with individual font selection, in which you can go forwards or backwards, but I've found that in actual use it's not that great and isn't really needed as people will likely only ever have a small number of arrays.
Consider it dumped.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+bool IsResetSoundPlaying() {
+  return !!GetWavPlayerPlaying(&SFX_reset);
+}
+
+bool IsBootSoundPlaying() {
+  return !!GetWavPlayerPlaying(&SFX_boot);
+}
+#endif
+
+// CLICK PLAYER FOR BUTTON PRESSES (optional).
```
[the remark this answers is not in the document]
Sabersense-CC: I was using caps as a kind of chapter heading for separate sections. But the purpose of this one is made clearer on line 622 which reads:       // Intended for Scavenger hilt where wheel makes tactile feel difficult. But I agree, it's not clear. I'll rejuggle the comments...

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+bad_blade:
+  ProffieOSErrors::error_in_blade_array();
+#endif
+}
+#endif
+
+// RESET FACTORY DEFAULTS (Delete Save Files).
+// Script to determine if sound effects have finished.
+#ifdef SABERSENSE_ENABLE_RESET
+bool IsResetSoundPlaying() {
```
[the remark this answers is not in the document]
Sabersense-CC: Errmm... I'll need to go away and do some research on how to do that. I'm guessing you mean we build it such that it checks if ANY sound is still playing, not just the ones explicitly specified in these bools (i.e. reset and boot)? If we do that, do we not run the run the risk of messing up other sounds that are supposed to overlap?

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+        }
+      } else {
+        thrust_begin_millis_ = millis();
+      }
+    }
+    // Enables sequential sounds and processes on array handling.
+#ifdef SABERSENSE_BLADE_ID
+    if (do_font_after_sound_ && !IsBladeidSoundPlaying()) {
+      SaberBase::DoNewFont();
+      do_font_after_sound_ = !do_font_after_sound_;
```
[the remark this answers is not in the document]
Sabersense-CC: Willdo.

Sabersense-CC:
OK, quite a few bits for me to look at there, most of which are fine, but a couple I'll need to research a little. Also rejigging some of the IFs means I want to test thoroughly before I submit an update. Will get it sorted as quick as I can - likely the next couple of days as I'm aware the OS-8 clock is ticking...

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Looking at this now, and I'm happy to change it to whatever you prefer, but It just occurred to me, are we comparing like-for-like here? CIRCULAR_VOLUME_MENU suggests to me that the volume goes up or down with hilt rotation, like colour change, not with button clicks. But that isn't the case here. Mine just does it on button clicks - albeit on one-button systems point up and click for up, point down and click for down. Or is that sufficiently circular for this title?

props/saber_sabersense_buttons.h
Sabersense-CC: Fixed.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+      STDOUT.print("Maximum Volume: ");
+#endif
+    }
+  }
+
+  void VolumeDown() {
+    STDOUT.println("Volume Down");
+    if (dynamic_mixer.get_volume() > (0.10 * VOLUME)) {
+      dynamic_mixer.set_volume(std::max<int>(VOLUME * 0.1,
+        dynamic_mixer.get_volume() - VOLUME * 0.10));
```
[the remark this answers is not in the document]
Sabersense-CC: Done.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+        if (SFX_volmax) {
+          hybrid_font.PlayCommon(&SFX_volmax);
+        } else {
+          beeper.Beep(0.5, 3000);
+        }
+        STDOUT.print("Maximum Volume: ");
+        max_vol_reached_ = true;
+      } else {
+        dynamic_mixer.set_volume(std::max<int>(VOLUME * 0.1,
+        dynamic_mixer.get_volume() - VOLUME * 0.90));
```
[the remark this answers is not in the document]
Sabersense-CC: Now done.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Have changed it to SABERSENSE_VOLUME_MENU for now.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Done.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Toggles now removed and replaced with hard true or false as applicable.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Done.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Done.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Now removed - much cleaner.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+#undef ACTIVATE
+#define ACTIVATE(N) do {                      \
+  if (!current_config->blade##N) {            \
+    goto bad_blade;                           \
+  }                                           \
+  current_config->blade##N->Activate(N);      \
+} while(0);
+
+void FakeFindBladeAgain() {
+  // Reverse everything FindBladeAgain does, except for recalculating best_config.
```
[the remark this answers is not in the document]
Sabersense-CC: Removed as it it didn't really make sense.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Comments rejigged - hopefully now clearer.

props/saber_sabersense_buttons.h
profezzorn: it doesn't look fixed

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Edit: line 530 now, due to tweaking lines above.

props/saber_sabersense_buttons.h
```diff
@@ -526,18 +533,17 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
...
-#ifdef SABERSENSE_ENABLE_ARRAY_FONT_IDENT  // Plays 'array' sound AND 'font' sound.
   void NextBladeArray() {
     FakeFindBladeAgain();
+#ifdef SABERSENSE_ENABLE_ARRAY_FONT_IDENT  // Plays 'array' sound AND 'font' sound.
     SFX_array.Select(current_config - blades);
     hybrid_font.PlayCommon(&SFX_array);
     // Calls Loop function to handle waiting for effect before running DoNewFont.
-    do_font_after_sound_ = !do_font_after_sound_;
+    do_font_after_sound_ = true;
+    do_font_after_sound_ = true;
```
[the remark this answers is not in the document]
Sabersense-CC: Doh! How d'I miss that? Thanks Oli! Now fixed.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: I've had another look at this to see if I can lose the repetition of the Saberbase and do_font lines, but unfortunately I can't see a way to do it that works. I can make it work if you have either of the stated defines, but if you have NEITHER of those defines in your config, you end up with those two lines being somewhat orphaned.
As such, unless there's a slicker method that I don't know about (very likely, I admit), then I think as it is is the clearest and simplest way to do it. I've added an extra comment to hopefully help with clarity.

props/saber_sabersense_buttons.h
Sabersense-CC: Sorry - now done.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1184 @@
...
+                              You must exit COLOUR MENU to resume using lightsaber normally.
+
+  *   = Gesture ignitions also available via defines.
+  **  = Audio player orientations can be reversed using SABERSENSE_FLIP_AUDIO_PLAYERS define.
+  
+===========================================================
+=================== SABERSENSE DEFINES ====================
+
+#define SABERSENSE_BLADE_ID
+  Replaces regular BladeID (which relies on continuous ID
```
profezzorn, line 200: This is a bit confusing, because regular BladeID can be done with and without continuous ID scanning.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1184 @@
...
+  scanning) with on-demand ID scanning.
+  Triple-click MAIN will run the scan, identify the blade
```
profezzorn, line 202: MAIN? is that the power button?

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1194 @@
...
+        millis() - clash_impact_millis_ > SABERSENSE_LOCKUP_DELAY &&
+        SaberBase::Lockup()) {
+        SaberBase::DoEndLockup();
+        SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+        auto_melt_on_ = false;
+      }
+
+      // EVENT_PUSH
+      if (fabs(mss.x) < 3.0 &&
+        mss.y * mss.y + mss.z * mss.z > 70 &&
```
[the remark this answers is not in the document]
profezzorn: line 358 is still only 2 steps indented from line 357, should be 4.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1184 @@
...
+  }
+
+  // BLADE ID OPTIONS AND ARRAY MANAGEMENT
+  // True BladeID, runs scan on-demand with unique 'bladeid' ident sound.
+#ifdef SABERSENSE_BLADE_ID
+#ifndef ENABLE_POWER_FOR_ID
+#error "SABERSENSE_BLADE_ID requires ENABLE_POWER_FOR_ID to be defined."
+#endif
+
+  bool IsBladeidSoundPlaying() {
```
profezzorn, line 505: This function is only called in one place, seems like it would be easier to just inline the code.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
profezzorn: I think I would just do this:

#if defined(SABERSENSE_BLADE_ID) || defined(SABERSENSE_ARRAY_SELECTOR)
   if (do_font_after_sound_
#ifdef SABERSENSE_BLADE_ID
       && !GetWavPlayerPlaying(&SFX_bladeid)
#endif       
#ifdef SABERSENSE_ARRAY_SELECTOR
       && !GetWavPlayerPlaying(&SFX_array)
#endif
    ) {
          SaberBase::DoNewFont();
      do_font_after_sound_ = false; 
   }
#endif

Note, the indentation is probably way off...

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1184 @@
...
+          dynamic_mixer.get_volume() - VOLUME * 0.10));
+      if (SFX_voldown) {
+        hybrid_font.PlayCommon(&SFX_voldown);
+      } else {
+        beeper.Beep(0.5, 2000);
+      }
+      STDOUT.print("Volume Down - Current Volume: ");
+      STDOUT.println(dynamic_mixer.get_volume());
+    } else {
+#ifdef SABERSENSE_VOLUME_MENU
```
profezzorn, line 468: I can't find a description for this define, what does it do?

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1184 @@
...
+  }
+#endif
+#endif
+
+  // Manual Array Selector, switches on-demand to next array, plays 'array' ident sound.
+#ifdef SABERSENSE_ARRAY_SELECTOR
+#ifdef SABERSENSE_BLADE_ID    // Only one Sabersense BladeID standard permitted.
+#error "SABERSENSE_ARRAY_SELECTOR and SABERSENSE_BLADE_ID cannot be defined at the same time."
+#endif
+  bool IsArraySoundPlaying() {
```
profezzorn, line 534: Short and only called in one place -> inline

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1184 @@
...
+  }
+#else
+  // Plays 'array' sound only, or 'font' sound if no 'array' sound available.
+    if (SFX_array) {
+      SFX_array.Select(current_config - blades);
+      hybrid_font.PlayCommon(&SFX_array);  // Play 'array' sound file if available.
+    } else {
+      SaberBase::DoNewFont();  // Play font ident if 'array' sound file missing.
+    }
+  }
```
profezzorn, line 554: move this to after the endif (and delete line 545)

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1184 @@
...
+    // Calls Loop function to handle waiting for effect before running DoNewFont.
+    do_font_after_sound_ = true;
+  }
+#else
+  // Plays 'array' sound only, or 'font' sound if no 'array' sound available.
+    if (SFX_array) {
+      SFX_array.Select(current_config - blades);
+      hybrid_font.PlayCommon(&SFX_array);  // Play 'array' sound file if available.
+    } else {
+      SaberBase::DoNewFont();  // Play font ident if 'array' sound file missing.
```
profezzorn, line 552: This won't work. You *have* to call DoNewFont() when selecting a new font, or several things (like displays) won't get initialized properly.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1184 @@
...
+
+          // Iterate over the files to delete in this save directory.
+          for (const char* targetFile : filesToDelete) {
+            PathHelper filePath(saveDirPath);
+            filePath.Append(targetFile);
+
+            // If the file exists in this directory, delete it.
+            if (LSFS::Exists(filePath)) {
+              LSFS::Remove(filePath);
+              Serial.print("Deleted from ");
```
profezzorn, line 1081: STDOUT << "Deleted from " << saveDirPath << ": " < targetFile <<"\n";

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1184 @@
...
+// Manual Blade Array Selection version of FindBladeAgain()
+#undef ACTIVATE
+#define ACTIVATE(N) do {                      \
+  if (!current_config->blade##N) {            \
+    goto bad_blade;                           \
+  }                                           \
+  current_config->blade##N->Activate(N);      \
+} while(0);
+
+void FakeFindBladeAgain() {
```
profezzorn, line 566: Please add a comment that explains what this function is for.
(and how it's different from FindBladeAgain())

props/saber_sabersense_buttons.h
Sabersense-CC: Ahh - that's tricky.
So is there a way to DoNewFont without playing the font ident? I'm guessing not.
One of the intended uses of array switching is to have mutliple blade arrays, each with a different blade length, all looking at the same preset array, so people could simply switch to a different blade length on the buttons. The system would play a confirmation wav telling them which array they've switched to, but since, in this instance, no save files would be specified, they wouldn't actually be switching to a new font, only a new array.
But of course that's only one usage - other usages would have multiple save files and as such most likely would be switching to a new font whenever the array is switched.
I'm wondering if there is perhaps a way to code it so the system knows if it's changing font as well as array - if it only changes array it doesn't run DoNewFont, but if chaging array makes it land on a new font, that will prompt it to run DoNewFont to ensure the various initializations are done.
I'll need to go away and figure that one out.
Work in progress....

profezzorn: This might work:

int saved_volume = SFX_font.GetVolume();
SFX_font.SetVolume(0);
SaberBase::DoNewFont();
SFX_font.SetVolume(saved_volume);

It will still play it of course, but at volume zero....

Sabersense-CC: Neat workaround! Just tested and work a treat. Thanks so much Prof. And I'm guessing it "plays" the mute font ident at the same time as the regular array ident, so everything gets done as quick as possible.
Thanks again Prof. Am working my way through the other bits - should have an update posted soon. :)

props/saber_sabersense_buttons.h
Sabersense-CC: If I've understood it correctly, our latest tweak to run DoNewFont with the font ident muted means we need to keep that endif as is because the optional define to force font idents to play after array switches. I think...?

props/saber_sabersense_buttons.h
Sabersense-CC: Will explain that better in the comments.

props/saber_sabersense_buttons.h
Sabersense-CC: I just realised it's superfluous to what I need. Will delete that whole section.

props/saber_sabersense_buttons.h
Sabersense-CC: Now explained in more detail.

props/saber_sabersense_buttons.h
Sabersense-CC: Sorry, that's my own terminology. All instances of MAIN now changed to POWER.

props/saber_sabersense_buttons.h
Sabersense-CC: I made the change you recommended and tested it, so this line has now gone. Sincere thanks for the solution.

props/saber_sabersense_buttons.h
Sabersense-CC: Now gone.

props/saber_sabersense_buttons.h
Sabersense-CC: Was part of the change you recommended. Now gone.

props/saber_sabersense_buttons.h
Sabersense-CC: Done, albeit over three lines, as it was a long line that started quite a long way out due to indenting.

props/saber_sabersense_buttons.h
Sabersense-CC: Done. Hope it's clear.

props/saber_sabersense_buttons.h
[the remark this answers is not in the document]
Sabersense-CC: Should be done now.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1145 @@
...
+#endif
+
+  void TriggerBladeID() {
+    FindBladeAgain();
+#ifdef SABERSENSE_ENABLE_ARRAY_FONT_IDENT  // Plays 'bladeid' sound AND 'font' sound.
+    SFX_bladeid.Select(current_config - blades);
+    hybrid_font.PlayCommon(&SFX_bladeid);
+    // Calls Loop function to handle waiting for effect before running DoNewFont.
+    do_font_after_sound_ = true;
+  }
```
profezzorn, line 471: this "}" (and the one on line 484) can be combined and moved after the #endif

props/saber_sabersense_buttons.h
profezzorn: The last line of  the #if and #else branches are the same, which means that they can be combined and moved after the #endif.  (I did not mean to suggest removing the #endif.)

props/saber_sabersense_buttons.h
profezzorn: It would be easier to override the blade ID than to replicate (almost) all of FindBladeAgain() I think.
You could basically just do:

   FakeBladeID::SetBladeIDOnce(next_entry.id);
   FindBladeAgain();

Overriding blade ID would need to be done before including prop_base.h though.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1145 @@
...
+#endif
+
+    // SOUND EFFECT PLAYERS.
+    // With Blade ON - UP for Character Quote, plays sequentially.
+    // With Blade ON - DOWN for Force Effect, plays randomly.
+    case EVENTID(BUTTON_POWER, EVENT_SECOND_SAVED_CLICK_SHORT, MODE_ON):
+#ifndef SABERSENSE_FLIP_AUDIO_PLAYERS
+      //  Define reverses UP/DOWN options for all QUOTE/FORCE/TRACK audio player.
+      //  Quote player points upwards.
+      if (SFX_quote) {
```
profezzorn, line 821: code starts the same in both #ifndef and #else, so some of this can be moved up.
same at the bottom, same can be moved down.

basically, instead of:

#if something
if (A) {
  FOO;
}
#else
if (A) {
  BAR
}
#endif

you can just write:

if (A) {
#if something
   FOO;
#else
   BAR;
#endif
}

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1145 @@
...
+
+          // Iterate over the files to delete in this save directory.
+          for (const char* targetFile : filesToDelete) {
+            PathHelper filePath(saveDirPath);
+            filePath.Append(targetFile);
+
+            // If the file exists in this directory, delete it.
+            if (LSFS::Exists(filePath)) {
+              LSFS::Remove(filePath);
+              STDOUT << "Deleted from "   \
```
profezzorn, line 1043: No need for backslashes at the end of the line here. (and below)

Sabersense-CC: Fixed.

props/saber_sabersense_buttons.h
Sabersense-CC: Fixed.

Sabersense-CC:
Regarding your comments about it being easier to override the blade ID than to replicate (almost) all of FindBladeAgain(), I get where you're coming from, but I'm hazy on the details of how to achieve it. However I notice that NoSloppy's 'Add Array Switching' pull request seems to have stalled on the same thing (I basically used his code for that bit) so I think he and I perhaps need to compare notes try to figure this bit out. Work in progress....

profezzorn:
> Regarding your comments about it being easier to override the blade ID than to replicate (almost) all of FindBladeAgain(), I get where you're coming from, but I'm hazy on the details of how to achieve it. However I notice that NoSloppy's 'Add Array Switching' pull request seems to have stalled on the same thing (I basically used his code for that bit) so I think he and I perhaps need to compare notes try to figure this bit out. Work in progress....

I'm confused, didn't we have this conversation already:
https://crucible.hubbe.net/t/manually-switch-from-one-preset-array-to-another/5959/31

Sabersense-CC:
That's true - we (well, you) got it working well just toggling between two arrays, then Olivier picked up the baton with Brian and I and we tried to get multiple array switching working using essentially the same principle as the toggle switching.
Unfortunately we ran aground somewhat and couldn't quite make it work in the same way, so we had to change direction a bit. The solution I reached meant another user define in the config specifying the number of arrays. My way of playing arrayx.wav files also meant specifiying each wav in the prop, which imposed a limit on the number of arrays you could have. But Brian found a better way to do it all which didn't have any of those limitations, so I jumped aboard his ship so to speak, and that's what we've got currently.
But I concur that somewhere in the original plan there is a solution waiting to be found, which will be a bit neater than duplicating that big chunk of prop_base.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1105 @@
...
+
+  void TriggerBladeID() {
+    FindBladeAgain();
+#ifdef SABERSENSE_ENABLE_ARRAY_FONT_IDENT  // Plays 'bladeid' sound AND 'font' sound.
+    SFX_bladeid.Select(current_config - blades);
+    hybrid_font.PlayCommon(&SFX_bladeid);
+    // Calls Loop function to handle waiting for effect before running DoNewFont.
+    do_font_after_sound_ = true;
+#else
+  // Plays 'bladeid' sound only, or 'font' sound if no 'bladeid' sound available.
```
profezzorn, line 489: indentation: two more spaces

props/saber_sabersense_buttons.h
profezzorn: No, I mean that this line (529) is identical to the line before the #else (516)
Both lines are just a bracket: }
Removing both of those lines, and then adding an identical line after the #endif will produce the same result, it will be easier to read, and is generally just better.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1105 @@
...
+      NextBladeArray();
+      return true;
+#endif
+
+    // SOUND EFFECT PLAYERS.
+    // With Blade ON - UP for Character Quote, plays sequentially.
+    // With Blade ON - DOWN for Force Effect, plays randomly.
+    case EVENTID(BUTTON_POWER, EVENT_SECOND_SAVED_CLICK_SHORT, MODE_ON):
+      if (SFX_quote) {
+#ifndef SABERSENSE_FLIP_AUDIO_PLAYERS
```
profezzorn, line 799: This #ifndef causes unbalanced { for editors that don't understand #ifdefs.

Instead of:

```
#ifndef BLABLABLA
    if(X) {
#else
    if (Y) {
#endif
```
it would be better to do:

```
#if BLABLABLA
     if (X)
#else
    if (Y)
#endif
    {
```

This makes the { and the } balanced, with or without preprocessing.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1105 @@
...
+  #define SABERSENSE_THRUST_ON
+
+============================================================
+============================================================
+*/
+#ifndef PROPS_SABER_SABERSENSE_BUTTONS_H
+#define PROPS_SABER_SABERSENSE_BUTTONS_H
+
+#ifdef SABERSENSE_ARRAY_SELECTOR
+  struct SabersenseArraySelector {
```
profezzorn, line 270: This whole struct should be shifted left two spaces.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1105 @@
...
+    static int return_value; // Tracks the current array index.
+    float id() { 
+      return return_value; 
+    }
+    static void cycle() {
+      return_value = (return_value + 1) % NELEM(blades);
```
profezzorn, line 276: This approach does work, but requires the user to use id 0, 1, 2, 3, etc. in the blades[] array.
If that's how you think it should work, then make sure to document that at the beginning of the file.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1124 @@
...
+  #define SABERSENSE_THRUST_ON
+
+============================================================
+============================================================
+*/
+#ifndef PROPS_SABER_SABERSENSE_BUTTONS_H
+#define PROPS_SABER_SABERSENSE_BUTTONS_H
+
+#include "prop_base.h"
+#include "../sound/hybrid_font.h"
```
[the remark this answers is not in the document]
Sabersense-CC: I think it may be a leftover from when I was aiming for OS7 Backwards compatibility for quote effects. Just tested without it and all works fine, so now deleted.

props/saber_sabersense_buttons.h
```diff
@@ -0,0 +1,1124 @@
...
+      do_font_after_sound_ = false;
+    }
+#endif
+  }
+
+  // VOLUME MENU
+  void VolumeUp() {
+    STDOUT.println("Volume up");
+    if (dynamic_mixer.get_volume() < VOLUME) {
+      dynamic_mixer.set_volume(std::min<int>(VOLUME + VOLUME * 0.1,
```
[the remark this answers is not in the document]
Sabersense-CC: I have to admit I don't know. All the volume handling is from the SA22C prop that I used the basis for this one. But now changed for something hopefully a little more logical.

props/saber_sabersense_buttons.h
Sabersense-CC: Soz. Fixed.

props/saber_sabersense_buttons.h
Sabersense-CC: Done.

props/saber_sabersense_buttons.h
Sabersense-CC: Yeh that suddenly occurred to me late last night. And to be honest, it isn't ideal as it messes up using Array Selector with Blade Detect. I've just tried it, and with BD showing a blade in, it gets to the end array and stops there - it won't cycle back to the first 'no_blade' array. And with no blade fitted, it won't cycle arrays at all.
Unfortunately I couldn't figure out how to make your recommended approach work - I kept getting errors along the lines of "override doesn't actually override".
All of that said, I wonder if numbered arrays will help simplify my aim to have the system save the selected array? If so, it might be worth the BD sacrifice.
For now I've added an explainer, and I've also added an error if people try to use BD at the same time. But I will most likely revisit it if and when I manage to get saving sorted.

profezzorn: when I manage to get saving sorted.
So does that mean you're fine with it working this way for now?

Sabersense-CC: Well, yes and no.
If the choice is between no Sabersense Prop and this Sabersense Prop, then yes, I'm happy to go with it as is until the next PR. But it's unfortunate that users have that extra hoop to jump through of numbering arrays sequentially, which it would be nice to avoid. 
I also don't like the idea of removing more functionality than is necessary to get new functionality to work, and this obviously removes Blade Detect. That said, I think integrating Array Selector with Blade Detect is potentially slightly confusing for the end user, so maybe it's not that big a deal after all. 
As I write this, I've actually just been trying to get your recommended approach to work, but I just can't seem to crack it. I also have a hunch that having those array index numbers will make array saving easier when I get that far, though of course I have nothing to back that hunch up.

So yes, let's call it and stick with this system until the next PR as it does work well.

profezzorn:
Time to merge this.
Doesn't mean it's done, it just means that we can focus on improvements rather than getting the first version done. :)

PR #719, opened 2025-01-15
BCprop - speakBladeId add say "blade"

NoSloppy:
Is this a backward compatibility issue? (if non-V2 Voicepack)

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -1524,6 +1524,7 @@ public:
 #ifdef SPEAK_BLADE_ID
   void SpeakBladeID(float id) override {
     if (&SFX_mnum) {
+      sound_library_v2.SayBlade();
       sound_library_.SayNumber(id, SAY_WHOLE);
     } else {
       PVLOG_NORMAL << "** No mnum.wav number prompts found.\n";


profezzorn:
No idea. You are calling init() on a V1 sound library and then using a V2 sound.
Currently that works, but isn't guaranteed to play anything. However, I would actually prefer if that *didn't* work, so I might try to change that in the future...

NoSloppy:
> You are calling init() on a V1 sound library and then using a V2 sound.

I don't follow. I thought that's the proper way to call for a V2 sound?
For example, I use `sound_library_v2.SaySequential()`.

profezzorn:
Hmm, I had that wrong. You call init() on the v2 library, but only if MENUS_SPEC_TEMPLATE is undefined....

So when MENU_SPEC_TEMPLATE is undefined, then an error message will be heard if you do not have a V2 sound library, so I'm not sure why you're worrying about backwards compatibility...

(When MENU_SPEC_TEMPLATE *is* defined there doesn't seem to be an init call, so no error messages, and all bets are off.)

NoSloppy:
> (When MENU_SPEC_TEMPLATE _is_ defined there doesn't seem to be an init call, so no error messages, and all bets are off.)

Ok well this is handled in prop_base:
```cpp
class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterface {
public:
  PropBase() : CommandParser() {
    current_mode = this;
#ifdef MENU_SPEC_TEMPLATE
    MKSPEC<MENU_SPEC_TEMPLATE>::SoundLibrary::init();
#endif
```
 so that's covered.
 
And then in the BC prop it says "This prop version requires a V2 ProffieOS Voicepack for menus to work right." ... so covered there as well.

PR #727, opened 2025-01-26
Add twist to commands

NoSloppy:
Works as a modifier too like `pow twist` or `pow tleft`

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1409,6 +1409,19 @@ public:
       }
       return true;
     }
+    if (!strcmp(cmd, "twist")) {
+        Event(BUTTON_NONE, EVENT_TWIST);
+        return true;
+    }
+    if (!strcmp(cmd, "twistleft") || !strcmp(cmd, "tleft")) {
+        Event(BUTTON_NONE, EVENT_TWIST_LEFT);
+        return true;
+    }
+
+    if (!strcmp(cmd, "twistright") || !strcmp(cmd, "tright")) {
+        Event(BUTTON_NONE, EVENT_TWIST_RIGHT);
+        return true;
+    }
 
 #ifdef ENABLE_AUDIO
 
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1409,19 +1409,6 @@ public:
       }
       return true;
     }
-    if (!strcmp(cmd, "twist")) {
-        Event(BUTTON_NONE, EVENT_TWIST);
-        return true;
-    }
-    if (!strcmp(cmd, "twistleft") || !strcmp(cmd, "tleft")) {
-        Event(BUTTON_NONE, EVENT_TWIST_LEFT);
-        return true;
-    }
-
-    if (!strcmp(cmd, "twistright") || !strcmp(cmd, "tright")) {
-        Event(BUTTON_NONE, EVENT_TWIST_RIGHT);
-        return true;
-    }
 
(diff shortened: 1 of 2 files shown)


props/prop_base.h
```diff
@@ -1409,6 +1409,19 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
       }
       return true;
     }
+    if (!strcmp(cmd, "twist")) {
```
profezzorn, line 1412: Not sure these should go in prop_base since lots of props don't have/use twists.

NoSloppy: ok sure. Moved to user prop instead.

NoSloppy:
> Maybe it should just be "left" and "right" ? prior art:
> 
> https://github.com/profezzorn/ProffieOS/blob/b88b82da47b2c5b02691bc216c150edf2613a53b/modes/stepped_mode.h#L82
> 
> 
> https://github.com/profezzorn/ProffieOS/blob/b88b82da47b2c5b02691bc216c150edf2613a53b/props/saber_fett263_buttons.h#L4588

I had *just* opened that file to start seeing about getting it to cooperate in a mode.
Thank you. yes standby.

NoSloppy:
https://github.com/profezzorn/ProffieOS/blob/b88b82da47b2c5b02691bc216c150edf2613a53b/modes/stepped_mode.h#L82
> 

This is unresponsive when in the mode. still responds to the prop Parse(). working on it.

NoSloppy:
Do you think it has to do with my custom modes? BCVolume Mode inherits from SteppedMode, but doesn't override mode_Parse(), so I don't get why it wouldn't work. 
Clearly I am missing something.
 
If I run the default saber.h prop, left and right are executed from stepped_mode mode_Parse().
But using my prop, it continues listening to button_base's Parse() and overridden versions.
hmm...

NoSloppy:
Sigh. Ok progress.
So I learned that if I have Parse() overridden in my prop SaberBCButtons class, apparently it nullifies the one in modes.
I removed it and modes work fine.
So I don't know yet how to have commands available for both non-mode and mode states.

NoSloppy:
> Maybe it should just be "left" and "right" ?

So the answer to this apparently is "it can't be the same".

profezzorn:
Sounds like you forgot to call PropBase::Parse from your Parse function.
Like;

if (PropBase::Parse(...)) return true;

NoSloppy:
uh huh. I did it wrong. 
Need to check at each command.

profezzorn:
> uh huh. I did it wrong.
> Need to check at each command.
> Working well now, thanks.

No, just do it at the beginning of the function.

NoSloppy:
ahh of course.
ok done.

props/saber_BC_buttons.h
```diff
@@ -1245,6 +1245,23 @@ struct BCVolumeMode : public SPEC::SteppedMode {
...
+    }
+    if (!strcmp(cmd, "left")) {
+        Event(BUTTON_NONE, EVENT_TWIST_LEFT);
+        return true;
+    }
+    if (!strcmp(cmd, "right")) {
+        Event(BUTTON_NONE, EVENT_TWIST_RIGHT);
+        return true;
+    }
+  }
```
profezzorn, line 1263: missing "return false" at end

NoSloppy: oops. ok fixed.

PR #729, opened 2025-01-28
BCprop - just exit mode after setting blade length

NoSloppy:
I find that 99% of the time, only one (main) blade is being edited. 
This saves some time and clicks.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -1319,14 +1319,18 @@ struct BCSelectBladeMode : public SPEC::MenuBase {
   uint16_t size() override { return NUM_BLADES; }
 
   void mode_activate(bool onreturn) override {
-    mode::getSL<SPEC>()->SaySelectBlade();
-    SPEC::SteppedMode::mode_activate(onreturn);
-    highlighted_blade_.Start(current_blade());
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -1319,29 +1319,33 @@ struct BCSelectBladeMode : public SPEC::MenuBase {
   uint16_t size() override { return NUM_BLADES; }
 
   void mode_activate(bool onreturn) override {
-    mode::getSL<SPEC>()->SaySelectBlade();
-    SPEC::SteppedMode::mode_activate(onreturn);
-    highlighted_blade_.Start(current_blade());
-    PVLOG_NORMAL << "** Highlighting blade: " << current_blade() << "\n";
+    // Just exit after setting length
+    if (onreturn) {
+      popMode();
```
profezzorn, line 1324: I think this is fine, but there is an alternative that may potentially be simpler. Basically, you could do popMode() right before pushing BladeLengthMode on the mode stack. That way this mode won't be in the mode stack anymore, and we won't need to do anything special when we return from BladeLengthMode.

NoSloppy: ah. Works great, thanks.

PR #731, opened 2025-01-29
Add tick option to smooth color change 

NoSloppy:
Apparently the "ipod clickwheel" feel is popular. This is optional with a define and only affects color change mode.

```cpp
// Use this define to turn on an optional "iPod® Clickwheel" sound when rotating the hilt in smooth colorchange mode (Color Wheel).
// ccchange.wav will play, so the sound should probably be a very subtle "tick" sound.
// The value is the number of ticks per 360 hilt rotation. Higher number = more ticks per angle change.
// Suggested Range 50 - 120. Using zero or not defining will turn ticks off.

#define SMOOTH_COLORCHANGE_TICKS_PER_REVOLUTION 120
```

*Updated to reflect new usage as of https://github.com/profezzorn/ProffieOS/pull/731/commits/874f09090a63cb7732169819fe0682d3005adfc1

### DIFF
--- a/modes/smooth_mode.h
+++ b/modes/smooth_mode.h
@@ -5,8 +5,8 @@
 #include "select_cancel_mode.h"
 #include "../common/angle.h"
 
-#if defined(SMOOTH_COLORCHANGE_TICK_SOUND_SENS) && SMOOTH_COLORCHANGE_TICK_SOUND_SENS > 0
-#define SOUND_TICK_ANGLE ((M_PI * 2 / 12) / SMOOTH_COLORCHANGE_TICK_SOUND_SENS)
+#if defined(SMOOTH_COLORCHANGE_TICKS_PER_REVOLUTION) && SMOOTH_COLORCHANGE_TICKS_PER_REVOLUTION > 0
+#define SOUND_TICK_ANGLE (M_PI * 2 / SMOOTH_COLORCHANGE_TICKS_PER_REVOLUTION)
 #endif
 
 namespace mode {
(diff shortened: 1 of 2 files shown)


NoSloppy:
I suppose SMOOTH_CC_TICK_SENS might be good enough. Less cumbersome, that's for sure.

profezzorn:
Never had an ipod, but I would imagine that the clicks on that wheel actually signifies moving between different songs or something?

NoSloppy:
> Never had an ipod, but I would imagine that the clicks on that wheel actually signifies moving between different songs or something?

Yes, it's stepped. very small steps. look:
https://youtu.be/9-mWbV3lZH0?si=NG-y5EZacrDZg47V&t=10

NoSloppy:
ShtokyD tried it out. (actually is who requested)
https://www.facebook.com/shtokcustomworx/videos/1818698408662918/

modes/smooth_mode.h
```diff
@@ -5,6 +5,10 @@
 #include "select_cancel_mode.h"
 #include "../common/angle.h"
 
+#if defined(SMOOTH_COLORCHANGE_TICK_SOUND_SENS) && SMOOTH_COLORCHANGE_TICK_SOUND_SENS > 0
+#define SOUND_TICK_ANGLE ((M_PI * 2 / 12) / SMOOTH_COLORCHANGE_TICK_SOUND_SENS)
```
profezzorn, line 9: Why not just use steps per revolution like everything else?

NoSloppy: uh.. 
Well, this was ported from working well in 7.14
Also, we're IN smoothmode, so using anything from stepped didn't occur to me.
Not sure how that would work.

profezzorn: I just meant to change the unit (and maybe the name) of this define. I did not mean to use anything from stepped mode.

tl;dr; don't divide by 12

Also, the define should not have COLORCHANGE in the name, because this will affect all smooth_mode menus.

NoSloppy: ok. How about now?

It's set to only execute when in smooth colorchange so as NOT to interfere with other modes.
`    if (SaberBase::GetColorChangeMode() == SaberBase::COLOR_CHANGE_MODE_SMOOTH) {`

profezzorn: that is not the right way to do this.
If we only want this to affect color change mode, then the code should go in the color change mode class, not here.

NoSloppy: > If we only want this to affect color change mode, then the code should go in the color change mode class, not here.

Makes sense.
Moved, tested. Working just as well.
Updated PR files.

PR #734, opened 2025-01-29
BCprop - Add blade array switching

NoSloppy:
Should probably limit this to only if there's more than 1 blade config array?

- guard the array switching code, done.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -74,7 +74,7 @@ public:
     current_mode = this;
 #ifdef MENU_SPEC_TEMPLATE
     MKSPEC<MENU_SPEC_TEMPLATE>::SoundLibrary::init();
-#endif    
+#endif
   }
   BladeStyle* current_style() {
 #if NUM_BLADES == 0
@@ -514,9 +514,7 @@ public:
     SetPreset(0, true);
   }
 
-#ifdef BLADE_DETECT_PIN
   bool blade_detected_ = false;
-#endif
 
   // Use this helper function, not the bool above.
   // This function changes when we're properly initialized
@@ -538,7 +536,7 @@ public:
   }
 
   // Measure and return the blade identifier resistor.
-  float id(bool announce = false) {
+  virtual float id(bool announce = false) {
     EnableBooster();
     BLADE_ID_CLASS_INTERNAL blade_id;
     float ret = blade_id.id();
@@ -559,7 +557,7 @@ public:
       return ret;
   }
 
-  size_t FindBestConfig(bool announce = false) {
+  virtual size_t FindBestConfig(bool announce = false) {
     static_assert(NELEM(blades) > 0, "blades array cannot be empty");
 
     size_t best_config = 0;
@@ -617,7 +615,7 @@ public:
     }
 
   // Must be called from loop()
-  void PollScanId() {
+  virtual void PollScanId() {
     if (find_blade_again_pending_) {
       find_blade_again_pending_ = false;
       int noblade_level_before = current_config->ohm / NO_BLADE;
@@ -637,19 +635,43 @@ public:
   void PollScanId() {}
 #endif // BLADE_ID_SCAN_MILLIS
 
+  void SelectBladeConfig(size_t best_config) {
+    PVLOG_STATUS << "blade = " << best_config << "\n";
+    current_config = blades + best_config;
+  }
+
+  void ActivateBlades() {
+    bool activation_failed = false;
+#define ACTIVATE(N) do {                               \
+    if (!current_config->blade##N) {                  \
+      activation_failed = true;                       \
+    } else {                                          \
+      current_config->blade##N->Activate(N);          \
+    }                                                 \
+  } while(0);
+    ONCEPERBLADE(ACTIVATE);
+#if NUM_BLADES != 0
+    if (activation_failed) {
+      ProffieOSErrors::error_in_blade_array();
+    }
+#endif
+  }
+
+  void DeactivateBlades() {
+    ONCEPERBLADE(UNSET_BLADE_STYLE)
+#define DEACTIVATE(N) do {                      \
+    if (current_config->blade##N)               \
+      current_config->blade##N->Deactivate();   \
+  } while(0);
+    ONCEPERBLADE(DEACTIVATE);
+  }
+
   // Called from setup to identify the blade and select the right
   // Blade driver, style and sound font.
   void FindBlade(bool announce = false) {
     size_t best_config = FindBestConfig(announce);
-    PVLOG_STATUS << "blade = " << best_config << "\n";
-    current_config = blades + best_config;
-
-#define ACTIVATE(N) do {     \
-    if (!current_config->blade##N) goto bad_blade;  \
-    current_config->blade##N->Activate(N);          \
-  } while(0);
-
-    ONCEPERBLADE(ACTIVATE);
+    SelectBladeConfig(best_config);
+    ActivateBlades();
     RestoreGlobalState();
 #ifdef SAVE_PRESET
     ResumePreset();
@@ -666,6 +688,20 @@ public:
   bad_blade:
     ProffieOSErrors::error_in_blade_array();
 #endif
+  }
+
+  void FindBladeAgain() {
+    if (!current_config) {
+      // FindBlade() hasn't been called yet - ignore this.
+      return;
+    }
+    // Reverse everything that FindBlade does.
+
+    // First free all styles, then allocate new ones to avoid memory
+    // fragmentation.
+    DeactivateBlades();
+    SaveVolumeIfNeeded();
+    FindBlade(true);
   }
 
   SavePresetStateFile savestate_;
@@ -751,26 +787,7 @@ public:
 #endif
   }
 
-  void FindBladeAgain() {
-    if (!current_config) {
-      // FindBlade() hasn't been called yet - ignore this.
-      return;
-    }
-    // Reverse everything that FindBlade does.
 
-    // First free all styles, then allocate new ones to avoid memory
-    // fragmentation.
-    ONCEPERBLADE(UNSET_BLADE_STYLE)
-
-#define DEACTIVATE(N) do {                      \
-    if (current_config->blade##N)               \
-      current_config->blade##N->Deactivate();   \
-  } while(0);
-
-    ONCEPERBLADE(DEACTIVATE);
-    SaveVolumeIfNeeded();
-    FindBlade(true);
-  }
 
   bool CheckInteractivePreon() {
     #define USES_INTERACTIVE_PREON(N) \
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -331,7 +331,7 @@ public:
 #endif
 #ifdef SAVE_CLASH_THRESHOLD
       || GetCurrentClashThreshold() != saved_global_state.clash_threshold
-#endif  
+#endif
       ) {
       SaveGlobalState();
     }
@@ -514,7 +514,9 @@ public:
     SetPreset(0, true);
   }
 
+#ifdef BLADE_DETECT_PIN
   bool blade_detected_ = false;
+#endif
 
   // Use this helper function, not the bool above.
   // This function changes when we're properly initialized
@@ -525,9 +527,9 @@ public:
 
   virtual void SpeakBladeID(float id) {
 #ifdef DISABLE_TALKIE
-#ifdef SPEAK_BLADE_ID    
+#ifdef SPEAK_BLADE_ID
     #error You cannot define both DISABLE_TALKIE and SPEAK_BLADE_ID
-#endif    
+#endif
 #else
     talkie.Say(spI);
     talkie.Say(spD);
@@ -557,7 +559,7 @@ public:
       return ret;
   }
 
-  virtual size_t FindBestConfig(bool announce = false) {
+  size_t FindBestConfigForId(float resistor) {
     static_assert(NELEM(blades) > 0, "blades array cannot be empty");
 
     size_t best_config = 0;
@@ -574,6 +576,10 @@ public:
       }
     }
     return best_config;
+  }
+
+  size_t FindBestConfig(bool announce = false) {
+    return FindBestConfigForId(id(announce));
   }
 
 #ifdef BLADE_ID_SCAN_MILLIS
@@ -615,7 +621,7 @@ public:
     }
 
   // Must be called from loop()
-  virtual void PollScanId() {
+  void PollScanId() {
     if (find_blade_again_pending_) {
       find_blade_again_pending_ = false;
       int noblade_level_before = current_config->ohm / NO_BLADE;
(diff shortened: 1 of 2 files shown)


NoSloppy:
> Should probably limit this to only if there's more than 1 blade config array?
> 
> * guard the array switching code, done.

Nope. Reverted.

props/saber_BC_buttons.h
```diff
@@ -2042,6 +2064,154 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
+  // Manual blade array selection
+  // even with Real TIme Blade ID active
+  bool manual_blade_array_active = false;
+  size_t last_real_best_config = NELEM(blades);
+  size_t real_best_config = NELEM(blades);
+  size_t fake_best_config = NELEM(blades);
+  size_t prev_real_best_config = NELEM(blades);
+
+#ifdef BLADE_ID_SCAN_MILLIS
+  // Must be called from loop()
```
profezzorn, line 2076: "must be called from loop" is the same comment as PollScanId() in prop base, but presumably this is not the same function, or you wouldn't have overridden it. Please update the comment to explain what this function does differently from the base function, otherwise the reader (me) will have to compare manually.

props/saber_BC_buttons.h
```diff
@@ -2042,6 +2064,154 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
 
 #endif  // BC_DUAL_BLADES
 
+  // Manual blade array selection
+  // even with Real TIme Blade ID active
+  bool manual_blade_array_active = false;
+  size_t last_real_best_config = NELEM(blades);
```
profezzorn, line 2070: Each of these variables need a comment to explain what they do and why.

NELEM(blades) is probably not a valid value for these variables, I assum that is on purpose?

props/saber_BC_buttons.h
```diff
@@ -2042,6 +2064,154 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
+  }
+
+  void NextBladeArray() {
+    manual_blade_array_active = true; // Enable manual mode
+    PVLOG_NORMAL << "** Manually Switching to Blade Array: " << (current_config - blades + 1) << "\n";
+    FakeFindBladeAgain();
+  }
+
+  // Manual Blade Array Selection version of FindBladeAgain()
+  void FakeFindBladeAgain() {
```
profezzorn, line 2158: Seems like it would be easier to just spoof the blade ID and then call the real FindBladeAgain()

props/saber_BC_buttons.h
NoSloppy: "must be called from loop"  was from the initial copy . removed.
Comments added to hopefully explain the functions.

props/saber_BC_buttons.h
NoSloppy: Pretty sure I tried simple.
Activation and deactivation of blades i think were the issue. Possibly something else but I needed to duplicate some.
Sorry, a little rusty on this one at first glance.
I'l re-read the thread when this was going on to refresh.

props/saber_BC_buttons.h
NoSloppy: changed to SIZE_MAX

NoSloppy:
I snuck in a couple of small edits to the instructions comments as opposed to a dedicated PR.

profezzorn:
This PR now has conflicts, also, have you considered this method:
https://crucible.hubbe.net/t/switching-between-many-more-than-2-preset-lists/6065/27
?

NoSloppy:
> have you considered this method: 

Playing with it now.
I assume you meant SaberBase -> PropBase, 
and `next_id_ = blades[n].id;` should be 'ohm'.

However, while this is much shorter/simpler, it doesn't seem to play nice with BladeID also being active (at first tests).
Regardless of what's manually set, it's always overridden by the reading.

That is the intent here: to have manual NextBladeArray, manual TriggerBladeID again, and automatic Blade ID active all at the same time.
That's what the code I submitted does well.
It was a lot of work to get it to do that.
I very well may have a bunch of superfluous stuff in the code, which can be optimized if possible, but it seems like the suggestion from the Crucible thread is too simplified to do all this?

Not sure you knew the full expectations of the code.

profezzorn:
Yes, I meant PropBase::id()

Did you make it virtual? (because the symptoms you describe is what would happen if it wasn't.)

> Not sure you knew the full expectations of the code.

No, I don't. The code duplication makes it really really difficult to tease out what has changed and what is merely duplicated, so I haven't done that.

NoSloppy:
> The code duplication makes it really really difficult to tease out what has changed and what is merely duplicated...

Ok so I refactored significantly, cleaning up where I could. By splitting PropBase::FindBlade() up into helper functions, I can just call them in the way needed, making it easier to read hopefully. 
I tried to incorporate your overriding id() suggestion, but it still requires the flags and tracking I had in place to allow for a spoofed Blade ID for manual switching while Blade ID is active.

I also unguarded the blade_detected_ bool (required BLADE_DETECT_PIN defined) , and removed additional changes not specifically relevant to this update. Those can be a small touchup PR to follow instead.

props/saber_BC_buttons.h
```diff
@@ -1823,7 +1845,7 @@ void DoSavedTwist() {
     PVLOG_NORMAL << (battle_mode_ ? "** Entering" : "** Exiting") << " Battle Mode\n";
     if (battle_mode_) {
       if (!hybrid_font.PlayPolyphonic(&SFX_bmbegin)) {
-        hybrid_font.DoEffect(EFFECT_FORCE, 0);
+        hybrid_font.PlayPolyphonic(&SFX_force);
```
NoSloppy, line 1848: I don't know why this shows up here as a diff and no conflict.

props/saber_BC_buttons.h
```diff
@@ -2007,6 +2029,159 @@ void DoSavedTwist() {
...
+      PVLOG_DEBUG << "***** Spoofing Blade ID to: " << next_id_ << "\n";
+      return next_id_;
+    }
+
+    float ret = PropBase::id(announce);  // Read real ID
+
+#ifdef BLADE_DETECT_PIN
+    if (!blade_detected_) {
+      PVLOG_STATUS << "NO ";
+      ret += NO_BLADE;
```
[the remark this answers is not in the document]
NoSloppy: With this total override though, the base id() wouldn't get used, so I made sure to include all it does after the custom stuff here (without plain calling PropBase::id() because that would undermine the purpose here)

props/saber_BC_buttons.h
NoSloppy: > I'm going to see if I can come up with a simpler way.

That's great

props/saber_BC_buttons.h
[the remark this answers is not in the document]
profezzorn: But you're calling PropBase::id() on line 2082?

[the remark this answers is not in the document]
NoSloppy: Oh I think I meant to put that before the early return of use_next_id is true. 
Getting sloppy. Also not at my computer at the moment.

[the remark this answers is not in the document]
NoSloppy: If you're going to be doing something simpler, this might all change so not sure how much I should bother tweaking here.

props/saber_BC_buttons.h
```diff
@@ -2007,6 +2029,159 @@ void DoSavedTwist() {
...
+    if (!SaberBase::IsOn()) {
+      SaberBase::DoBladeDetect(blade_detected_);
+    } else {
+      SaberBase::DoNewFont();
+    }
+  }
+
+  // Play array.wavs if they exist and we manually switched.
+  // Fallback to playing font.wav.
+  void HandlePresetSound() {
```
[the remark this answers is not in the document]
NoSloppy: renaming

NoSloppy:
Ok. Refactored again based on your suggestions. All seems to be working correctly in testing.
All array detection and switching works right.
Manually triggering a BLADE ID scan works right.
Blade in/out choosing correct config works right.

This was the desired behavior as far as sounds goes
- When manually switching arrays, array.wav files should be used if they exist, otherwise play bladein/out sounds. Fallback to font.wav
- When a REAL blade inserted or removed, bladein/out sounds should play. Fallback to font.wav

The only time the sound doesn't "work" is when you're already manually switched to an array which a real blade insertion/removal would also trigger (like being on array 2, and inserting a blade that triggers array 2.)
No sound is played in this scenario (I can live with that if need be?)

Otherwise, tested with/without the sounds, all fallback and work as expected otherwise.
Thanks.

props/prop_base.h
```diff
@@ -668,6 +672,27 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
 #endif
   }
 
+  void FindBladeAgain() {
```
profezzorn, line 675: Looks like this was moved?
(Can we undo that to make the diff smaller?)

profezzorn:
> The only time the sound doesn't "work" is when you're already manually switched to an array which a real blade insertion/removal would also trigger (like being on array 2, and inserting a blade that triggers array 2.) No sound is played in this scenario (I can live with that if need be?)

If you want to fix this, you would need to check in id() if returning real_id is going to result in a change or not. If it wouldn't you would set a bool variable, and then play a sound in loop() if that variable is set. (I don't think it's a good idea to to DoEffect() directly from id() since it's sometimes called from odd places.)

It would look something like this: 
```
  if (current_config - blade == real_best_config) play_no_change_sound_ = true;
```
(this would go right before "return real_id;")

Oh, one more thing; you would want to reset play_no_change_sound_  to false in PlayArraySound().

props/prop_base.h
NoSloppy: Yes I just moved it to group it with the other relative functions. I can move it back and add that last piece of puzzle tomorrow. Thanks.

NoSloppy: moved back

props/prop_base.h
```diff
@@ -514,9 +514,9 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
     SetPreset(0, true);
   }
 
-#ifdef BLADE_DETECT_PIN
+// #ifdef BLADE_DETECT_PIN
```
profezzorn, line 517: Why comment out the #ifdef ?

NoSloppy: I am finishing up the edits to my prop - needed blade_detected_ in there, but doing without it instead.
The aboves above will revert in a moment.

props/prop_base.h
```diff
@@ -74,7 +74,7 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
     current_mode = this;
 #ifdef MENU_SPEC_TEMPLATE
     MKSPEC<MENU_SPEC_TEMPLATE>::SoundLibrary::init();
-#endif    
+#endif
```
NoSloppy, line 77: Sorry all these whitespace corrections showed up. I just overpasted the whole file.

NoSloppy:
> If you want to fix this, you would need to check in id() if returning real_id is going to result in a change or not. If it wouldn't you would set a bool variable, and then play a sound in loop() if that variable is set.

Ok I have tested this works now, thanks.
I also restored FindBladeAgain() to it's original placement in prop_base.

props/prop_base.h
NoSloppy: done.

NoSloppy:
> Oh, one more thing; you would want to reset play_no_change_sound_ to false in PlayArraySound().

play_no_change_sound_ will always already be false when reaching PlayArraySound(). It gets set false as soon as it becomes true. Is it ok as I have it? (not set false in PlayArraySound())  Or is this better coding, like a "just incase" defensive thing to add ?

props/saber_BC_buttons.h
```diff
@@ -1567,6 +1589,25 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
     if (scroll_presets_beep_delay_timer_.isTimerExpired()) {
         SaberBase::DoEffect(EFFECT_NEWFONT, 0);
     }
+    // Blade array changing helper. See id() for usage.
+    if (play_no_change_sound_) {
+      play_no_change_sound_ = false;
+
+      float real_id = PropBase::id(false);
+      size_t real_best_config = FindBestConfigForId(real_id);
```
profezzorn, line 1597: if no change happened, then this should be the same as current_config - blade.

props/saber_BC_buttons.h
```diff
@@ -1567,6 +1589,25 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
+
+      if (real_best_config == 0) {
+        // Real blade event resulted in no blade array, so blade removal
```
profezzorn, line 1600: This comment doesn't make any sense.
real_best_config == 0 just means the first entry in the blade array.

profezzorn: Maybe you wanted to check current_config->ohm >= NO_BLADE ?

props/saber_BC_buttons.h
```diff
@@ -1567,6 +1589,25 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
...
+        if (!hybrid_font.PlayPolyphonic(&SFX_bladeout)) {
```
profezzorn, line 1601: I don't understand why we would play bladeout/bladein if there was no change?
Shouldn't this play the array sound?

profezzorn:
> play_no_change_sound_ will always already be false when reaching PlayArraySound(). It gets set false as soon as it becomes true. Is it ok as I have it? (not set false in PlayArraySound()) Or is this better coding, like a "just incase" defensive thing to add ?

It's fine, I was thinking about an edge case, but now that I look at the code, that edge case is nearly impossible to trigger, so it's fine.

props/saber_BC_buttons.h
NoSloppy: Even if the blade array doesn’t actually change, I'm trying to distinguish between blade insertion and removal.
Yes, this is assumed that NO_BLADE is first in the BladeConfig.
That's how I always run it. I could add a "rule" to the prop comments saying that. But if it's unacceptable then I need to work on it, possibly again unguarding blade_detected_.

props/saber_BC_buttons.h
NoSloppy: This is assumed that NO_BLADE is first in the BladeConfig.
If the blade is removed, that's where we land, so a blade must have been removed.
See comment below.

NoSloppy: > Maybe you wanted to check current_config->ohm >= NO_BLADE ?

This is good.
Done

props/saber_BC_buttons.h
NoSloppy: Yes, thanks.

props/saber_BC_buttons.h
profezzorn: So, are you trying to capture the difference between the current state and the previous "fake" state?
Or are you thinking something else?

NoSloppy: Again, I was going with "if a real blade is removed, we will be arriving on NO_BLADE, so must be bladein.wav.
If a blade is inserted, we will be switching to something *not* NO_BLADE, so play bladein.wav."
It is working.
I did change 
```cpp
if (real_best_config == 0) {
        // Real blade event resulted in no blade array, so blade removal
        ```
        
   to use `current_config->ohm >= NO_BLADE` instead, so as not to force NO_BLADE to be the first array.

Comment (profezzorn) [reply props/saber_BC_buttons.h:1598]: I think I'm a bit slow today, it makes sense now.

Comment (NoSloppy) [reply props/saber_BC_buttons.h:1598]: Oh, well I just found that I still get font.wav if going from an array that's not zero, to another one (like be on array 1 , insert blade, go to array 3, but font.wav plays instead of expected bladein.wav. 
I can't nip it.

File: props/saber_BC_buttons.h [hunk 13]
```diff
@@ -2007,6 +2045,55 @@ void DoSavedTwist() {
...
+  }
+
+  void TriggerBladeID() {
+    use_fake_id_ = false;
+    FindBladeAgain();
+    PlayArraySound();
+  }
+
+  void NextBladeArray() {
+    best_config_before_faking_ = FindBestConfigForId(PropBase::id(false));
```
profezzorn, line 2079: I realized that this doesn't work if the blade is on.
Your key bindings are all for MODE_OFF, so maybe it doesn't matter, but the blade can still be on if people have idle effects, long fadeouts, or battery effects.

I think we can replace it with:
```
 if (!use_fake_id_) best_config_before_faking_ = current_config - blades;
```

NoSloppy: Great. In my limited testing, this works the same, so hopefully it helps other scenarios as well.
Edited.

PR #736, opened 2025-02-01
Add mclick

NoSloppy:
The ticking smoothmode option doesn't need to "take over" ccchange that might be used elsewhere.
A click sound might not be appropriate/desired, such as in stepped mode, or menu change etc...
Since I already incorporated mclick in the V2 voicepacks, this can work "out of the box".
Lastly, if no mclick exists, a similar sound in the form of a beep is used.
All of this of course if the define to use a sound in smoothmode is active).

### DIFF
--- a/modes/color_change_modes.h
+++ b/modes/color_change_modes.h
@@ -12,10 +12,10 @@
 namespace mode {
 
 template<class SPEC>
-struct SteppedVariationMode : public SPEC::SteppedMode {
+struct SteppedVariationMode : public SPEC::SteppedModeBase {
   void mode_activate(bool onreturn) override {
     saved_ = get();
-    SPEC::SteppedMode::mode_activate(onreturn);
+    SPEC::SteppedModeBase::mode_activate(onreturn);
     SaberBase::SetColorChangeMode(SaberBase::COLOR_CHANGE_MODE_STEPPED);
   }
   void exit() override {
@@ -29,10 +29,6 @@ struct SteppedVariationMode : public SPEC::SteppedMode {
   }
   void next() override { SaberBase::UpdateVariation(1); }
   void prev() override { SaberBase::UpdateVariation(-1); }
-
-  void update() override {
-    hybrid_font.PlayPolyphonic(&SFX_ccchange);
-   }
 
   int get() { return SaberBase::GetCurrentVariation(); }
   void set(int x) { SaberBase::SetVariation(x); }
--- a/modes/color_change_modes.h
+++ b/modes/color_change_modes.h
@@ -67,7 +67,7 @@ struct SmoothVariationMode : public SPEC::SmoothWraparoundMode {
       if (tick_sound_angle_smooth > M_PI)  tick_sound_angle_smooth -= 2.0f * M_PI;
       if (tick_sound_angle_smooth < -M_PI) tick_sound_angle_smooth += 2.0f * M_PI;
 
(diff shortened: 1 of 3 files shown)


profezzorn:
So, your comment explains what this change does, which is good, but I'm not sure about the "why".
Is it bad to "take over" ccchange?
Do we really need a fallback if mclick isn't there? (all other V2 voice pack sounds are NOT optional)

Please note, I'm not trying to be negative here, I'm just trying to understand if this is really the right way to control this or not.

NoSloppy:
I'll attempt to clarify.
The "tick" during color change works best when it is an extremely short, sharp sound (ideally less than 50ms). It's pretty much meant to be just a specific purpose tick sound or similar.
If it's a different, longer "swoosh" or other "sound effect", the overlap of the sound and the longer duration makes the effect lack that association with a virtual detent point in the rotation. I know because I tried a few. It just gets muddy and feels imprecise.
Many fonts come with sounds for the ccchange, or color.wav files. They tend to thematically match the font.
Take the Transformers sound for example. https://www.youtube.com/watch?v=o7dXLJaGqKM
Playing that during a ColorSelect event works great, but not so much for 120 ticks per rotation where it's just ridiculous.

> Is it bad to "take over" ccchange?

By "taking over", I meant that forcing the ccchange sound to contain the click sound would limit other events to the same tick, where a different sound would likely fit better.

So in the end, I thought it best to designate mclick as just that... a ticky sound for the purpose of the smooth colorchange mode, available to turn on or off, available already in a Voicepack. It could also have future use like a button press or something I suppose.
Leaving ccchange for more singular events like stepped mode or menu change things seems good so as to still allow for the variety and purpose that fonts that supply them had intended.

profezzorn:
But regular stepped menus currently use ccchange and they assumes that the sounds are short, because it won't start the sound that says what the menu is until after cccchange has played. Having a long ccchange won't work well in that use case either.

By your logic, the stepped mode should also use mclick.wav

Only menues that have no audible labels would work well with long "tick" sounds. (Which pretty much only applies to the stepped color change menu and nothing else.)

I do wonder if having the fallback being no sound would be better?

NoSloppy:
> Which pretty much only applies to the stepped color change menu

Agreed. Do you want to change the EFFECT_MENU_CHANGE to mclick as well, and then make stepped color change be the place that ccchange is used?

> I do wonder if having the fallback being no sound would be better?

Sure. Again, that was just an afterthought. 
Removed.

NoSloppy:
ok how about now?
Tested, working as expected. 
Makes sense.

modes/color_change_modes.h
```diff
@@ -30,6 +30,10 @@ struct SteppedVariationMode : public SPEC::SteppedMode {
   void next() override { SaberBase::UpdateVariation(1); }
   void prev() override { SaberBase::UpdateVariation(-1); }
 
+  void update() override {
```
profezzorn, line 33: This all seems good to me, but if we're not going to use say() and fadeout(), then we should probably inherit from SteppedModeBase rather than SteppedMode.

NoSloppy: oh wait, we don't even need to play ccchange here! I just noticed that makes it play twice because hybrid_font already has this:
```
      case SaberBase::CHANGE_COLOR:
        if (!PlayPolyphonic(&SFX_ccchange)) {
          beeper.Beep(0.05, 2000.0);
        }
```
So no need to call the sound, just inherit from the base instead to avoid both a click and a ccchangeand it's good.

profezzorn: :+1:

PR #738, opened 2025-02-06
BCprop - use sound library instead...

NoSloppy:
... for mzoom, 
and volup->mclick in BCChangeBladeLength mode.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -1140,11 +1140,9 @@ EFFECT(dim);        // for EFFECT_POWERSAVE
 EFFECT(battery);    // for EFFECT_BATTERY_LEVEL
 EFFECT(bmbegin);    // for Begin Battle Mode
 EFFECT(bmend);      // for End Battle Mode
-EFFECT(volup);      // for increse volume
 EFFECT(push);       // for Force Push gesture
 EFFECT(tr);         // for EFFECT_TRANSITION_SOUND, use with User Effects.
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -1847,7 +1845,9 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
   void ToggleSpamBlast() {
     spam_blast_ = !spam_blast_;
     PVLOG_NORMAL << (spam_blast_ ? "** Entering" : "** Exiting") << " Spam Blast Mode\n";
-    if (!hybrid_font.PlayPolyphonic(&SFX_mzoom)) {
+    if (SFX_mzoom) {
```
profezzorn, line 1848: mzoom is required, so you don't really need to check if it exists or not
(Which means you can also remove EFFECT(mzoom))

NoSloppy: Even better. And yes, I had removed EFFECT(mzoom).
done.

NoSloppy:
That said, should I just remove all the checks for "required" sounds, like I have mnum checked in a bunch of places.
Maybe emphasize in the instructions up top that a Voicepack is **_required_** to use the prop?

profezzorn:
I mean it's up to you, but the intent is that when you call init() on the sound pack, you'll get an error if the sound pack isn't present. Having some additional instructions in your prop might be helpful though.

NoSloppy:
Done and done.
Moving forward, not looking back.

PR #739, opened 2025-02-14
BCprop account for newer idle sound handling

NoSloppy:
Also in here:
- Since STEPS_PER_REVOLUTION is now a user define setting, BCmodes don't need to override that function.
- Removed some dead comment lines, and added a few spacing newlines.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -1147,21 +1147,18 @@ struct BCScrollPresetsMode : public SPEC::SteppedMode {
 
   void next() override {
     beeper.Beep(0.05, 4000);
-#ifdef ENABLE_IDLE_SOUND
-    SFX_font.SetFollowing(nullptr);
-#endif 
     prop_next_preset();
   }
 
   void prev() override {
     beeper.Beep(0.05, 3000);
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -1288,24 +1283,6 @@ struct BCVolumeMode : public SPEC::SteppedMode {
     SPEC::SteppedMode::exit();
   }
 
-  bool Parse(const char *cmd, const char* arg) override {
```
profezzorn, line 1291: Why did this go away?

NoSloppy: Okay I'm not at my computer right now, but this doesn't belong in the mode here, so it's just been moved down into the saber BC buttons class. However, I thought it was appropriate in the other PR but your comment there seems like no. I just want to move it.

NoSloppy: > Why did this go away?

Added in #741 instead.

NoSloppy: ugh. I got mixed up. 
Ok #741 is correct now (moving the code block).
Restored here so it's no longer a diff.

PR #742, opened 2025-02-20
Move current_alternative setting

NoSloppy:


props/prop_base.h
```diff
@@ -1862,7 +1863,11 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
   void MovePreset(int position) {
     current_preset_.SaveAt(position);
   }
-  
+
+  virtual void SetCurrentAlternative() {
```
profezzorn, line 1867: How about ResetCurrentAlternative() ?

profezzorn:
You should probably add the changes to the BC blaster prop to this PR too, as that makes it obvious why you are doing it.

NoSloppy:
> You should probably add the changes to the BC blaster prop to this PR too, as that makes it obvious why you are doing it.

Well I totally was going to do that until I messed up and didn't make a new branch in my haste. So this is just a "patch" of my master. I'd like to get back in sync so I will have to have do that separately, sorry.
Coming right up.

PR #743, opened 2025-02-20
Fixed PR. current alt move

NoSloppy:
I think I just discarded incorrect commits and was able to edit BCblaster., so this is good now?

### DIFF
--- a/props/blaster_BC_buttons.h
+++ b/props/blaster_BC_buttons.h
@@ -538,6 +538,9 @@ public:
     Off();
   }
 
+  // Prevent alt setting to alt000 for blaster.
+  void ResetCurrentAlternative() override {};
+
   // Pull in parent's SetPreset, but set the correct alt sounds.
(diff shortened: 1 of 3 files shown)


props/blaster_BC_buttons.h
```diff
@@ -538,11 +538,19 @@ class BlasterBCButtons : public Blaster {
     Off();
   }
 
-  // Pull in parent's SetPreset, but set the correct alt sounds.
-  void SetPreset(int preset_num, bool announce) override {
-    Blaster::SetPreset(preset_num, announce);
-    SaberBase::DoEffect(EFFECT_ALT_SOUND, 0.0, blaster_mode);
+void ResetCurrentAlternative() override {
```
profezzorn, line 541: why not just current_alternative = (int)BLASTER_DEFAULT_MODE; ?

NoSloppy: No reason other than I'm not that slick.
Thanks. Done.

PR #744, opened 2025-02-23
BCblaster prop   adjustments

NoSloppy:
Adjust comments.

AUTOFIRE mode enabled just by SFX_auto existing:
  Fallback to STUN if AUTO mode defined as default but has no auto sounds.
  Needed to fully override NextBlasterMode to bypass define BLASTER_ENABLE_AUTO

Fix alt sound persisting through preset change:
  This switch was indeed needed, just not related to BLASTER_DEFAULT_MODE.
  Added back in setting current_alternative based on current blaster mode.

### DIFF
--- a/props/blaster_BC_buttons.h
+++ b/props/blaster_BC_buttons.h
@@ -342,6 +342,10 @@ public:
 
   void Setup() override {
     MKSPEC<BCBlasterMenuSpec>::SoundLibrary::init();
+    if (blaster_mode == MODE_AUTO && !SFX_auto) {
+      // BLASTER_DEFAULT_MODE is AUTO, but no auto.wav files found. Falling back to STUN.
+      blaster_mode = MODE_STUN;
+    }
   }
 
   void EnterVolumeMenu() {
--- a/props/blaster_BC_buttons.h
+++ b/props/blaster_BC_buttons.h
@@ -342,10 +342,6 @@ public:
 
   void Setup() override {
     MKSPEC<BCBlasterMenuSpec>::SoundLibrary::init();
-    if (blaster_mode == MODE_AUTO && !SFX_auto) {
-      // BLASTER_DEFAULT_MODE is AUTO, but no auto.wav files found. Falling back to STUN.
-      blaster_mode = MODE_STUN;
-    }
   }
 
   void EnterVolumeMenu() {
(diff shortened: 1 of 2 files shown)


props/blaster_BC_buttons.h
```diff
@@ -539,14 +541,42 @@ class BlasterBCButtons : public Blaster {
   }
 
 void ResetCurrentAlternative() override {
-  current_alternative = BLASTER_DEFAULT_MODE;
+  switch (blaster_mode) {
```
profezzorn, line 544: why not just do:

current_alternative = blaster_mode;

?

Also, this will keep the mode when you change presets, is that the intention?

NoSloppy: Yes, the intent is to have not only mode persist (which it already did) but to also keep the alt sounds in sync.
I keep thinking in terms of the mode names, not as their enum equivalent to an int.
Changed, and working correctly.

props/blaster_BC_buttons.h
```diff
@@ -344,6 +342,10 @@ class BlasterBCButtons : public Blaster {
 
   void Setup() override {
     MKSPEC<BCBlasterMenuSpec>::SoundLibrary::init();
+    if (blaster_mode == MODE_AUTO && !SFX_auto) {
```
profezzorn, line 345: I think you want to do this in SetPreset().
Setup() is only called once on startup.

NoSloppy: I tried that initially, but I think the sequence of SetPreset(), SetBlasterMode() and ResetCurrentAlternative() all come into play.
It will still use the alt sound of the BLASTER_DEFAULT_MODE unless It's set first at boot in Setup().
Since the only way to get into MODE_AUTO if no auto sounds exist is by the BLASTER_DEFAULT_MODE being set as AUTO, I figured one time at boot was sufficient.

profezzorn: But what happens if you change preset from one that has SFX_auto to one that doesn't have SFX_auto?

NoSloppy: Well, you're right of course. I only tested with one preset.
Would it be ok to move the new call to ResetCurrentAlternative() in chdir() down a line to *after* 
`Effect::ScanCurrentDirectory();`? Or would that mess other things up?

If yes, then I could just put this part in my ResetCurrentAlternative() override instead of needing to add SetPreset() and do this same routine in 2 places.

I tested like that and it works well. But unsure how it might affect other things using current_alternative.

profezzorn: Should be fine.

NoSloppy: done.

PR #745, opened 2025-02-23
Fix mount sd

NoSloppy:
Forces turning OFF if ON when Mount SD is allowed via `sd 1` command.
Also prevents idle.wav from playing if enabled and exists when `sd 1` was issued while ON.

### DIFF
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -483,7 +483,9 @@ public:
     bool most_blades = location.on_blade(0);
     // SFX_in.SetFollowing( most_blades ?  &SFX_pstoff : nullptr );
 #ifdef ENABLE_IDLE_SOUND
-    if (most_blades) {
+    if (LSFS::GetAllowMount()) {
+      SFX_in.SetFollowing(nullptr);
+    } else if (most_blades) {
         if (SFX_pstoff) {
             SFX_in.SetFollowing(&SFX_pstoff);
             SFX_pstoff.SetFollowing(&SFX_idle);
--- a/sound/hybrid_font.h
+++ b/sound/hybrid_font.h
@@ -483,7 +483,7 @@ public:
     bool most_blades = location.on_blade(0);
     // SFX_in.SetFollowing( most_blades ?  &SFX_pstoff : nullptr );
 #ifdef ENABLE_IDLE_SOUND
-    if (LSFS::GetAllowMount()) {
+    if (GetAllowMountSD()) {
       SFX_in.SetFollowing(nullptr);
     } else if (most_blades) {
         if (SFX_pstoff) {
--- a/ProffieOS.ino
+++ b/ProffieOS.ino
@@ -409,6 +409,7 @@ void EnableAmplifier();
 bool AmplifierIsActive();
 void MountSDCard();
 const char* GetSaveDir();
+bool GetAllowMountSD();
 
 #include "common/lsfs.h"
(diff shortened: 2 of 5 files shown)


NoSloppy:
I don't get any difference if it changes to `if (IsOn()) Off(); else SaberBase::DoOff(OFF_IDLE, 0);`instead, but maybe that's just better to have it that way?

sound/hybrid_font.h
```diff
@@ -483,7 +483,9 @@ class HybridFont : public SaberBase, public Looper {
     bool most_blades = location.on_blade(0);
     // SFX_in.SetFollowing( most_blades ?  &SFX_pstoff : nullptr );
 #ifdef ENABLE_IDLE_SOUND
-    if (most_blades) {
+    if (LSFS::GetAllowMount()) {
```
profezzorn, line 486: This kind of thing happens in the OLED code and in the new display code as well...
We should probably use the same condition everywhere.
To make it easier to maintain, we might want a helper function somewhere in common/sd_card.h or something.

NoSloppy: > This kind of thing happens in the OLED code and in the new display code as well... We should probably use the same condition everywhere. To make it easier to maintain, we might want a helper function somewhere in common/sd_card.h or something.

Might need some help with that as I haven't played with the display stuff yet.
Something like
```cpp
    #ifdef ENABLE_RGB565_DISPLAY
    rgb565frame.ClearDisplay();
    #endif

    #ifdef ENABLE_SSD1306_DISPLAY
    ssd1306.ClearDisplay();
    #endif
    ```
    in a function in sd_card.h?
    Not sure how to throw in the idle sound stopping there, or how to remove it from how I have it currently working.

profezzorn: There is function called EscapeIfNeeded in ssd1306, my thinking is that maybe we move that to sd_card.h, add GetAllowMount() to it, rename it to something better, and then call it from here, and from ssd1306 

The color display code apparently doesn't do this right now, but it might need to... (I can fix that later.)

NoSloppy: would we need to #include .. common/sd_card.h in both ssd1306 and hybrid_font then?
That causes errors I assume due to the oder of inclusion at compile time.

profezzorn: We could do that, or we could just move it up in ProffieOS.ino.
It gets a little more complicated if both depend on each other.
Alternatively, we could could make GetAllowMount() a global function (which I would then call GetAllowMountSD()), put the definition of it in sd_card.h, and put a declaration of it in ProffieOS.ino so that it can be called from anywhere.

NoSloppy: Also, "moving" EsacpeIdleIfNeeded() is eluding me.
I see we should call a helper function instead of USBD_Conncted() so that `sd 1` works right, but the bool still serves a local purpose.
A function in sd_card.h named like ShouldStopIdle() that just checks GetAllowMount() seems like ... just centralizing it somewhere logical then?
Then SB_Off in hybrid_font calls that instead like 
`if (ShouldStopIdle())`? Which brings us back to how to call it with the inclusion order.

NoSloppy: > Alternatively, we could could make GetAllowMount() a global function

ok, how about now?

profezzorn:
This required me to think about how I wanted this to work, and I ended up doing a little bit different.
(Thank you for making me think it through. :)

PR #747, opened 2025-02-26
add SubBladeWithList

NoSloppy:
The POD will need to be updated.
While at that, shouldn't the ZZ and maybe BarBack be included in the comments section here and POD as well?

### DIFF
--- a/blades/sub_blade.h
+++ b/blades/sub_blade.h
@@ -29,6 +29,10 @@ Usage: SubBladeWithStride(first_led, last_led, stride_length, blade_definition)
 Like SubBlade, but LEDs are indexed with an additional 'stride' parameter,
 allowing you to "skip" over a regular number of pixels in the data chain. (Such as every other one)
 
+Usage: SubBladeWithList<int1, int2, ...>(blade_definition)
+Like SubBlade, but you provide a custom list of LED indices instead of a range.
+Useful for ring-based or irregular LED layouts.
+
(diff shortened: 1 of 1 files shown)


blades/sub_blade.h
```diff
@@ -311,6 +315,58 @@ class BladeBase* SubBladeZZ(int first_led, int last_led, int stride, int column,
...
+  if (first_subblade_wrapper) {
+    ret->SetNext(last_subblade_wrapper);
+    first_subblade_wrapper->SetNext(ret);
+    last_subblade_wrapper = ret;
+  } else {
+    ret->SetNext(ret);
+    first_subblade_wrapper = last_subblade_wrapper = ret;
+  }
+
+  ret->SetupSubBlade(blade, 0, count);
```
profezzorn, line 359: This will make primary() return true.
However, there is only supposed to be one primary subblade per string.
Currently, the "primary()" is based on whether the subblade has an offset of zero, which is why this will always return true right now.

We have two options:
1. We can modify this call so that the offset is only zero if the list of integers contains a zero.
2. We could add a primary_ bool and set it to true in SetNext() if next == this;

I'm not entirely certain which is better. The first one means fewer modifications to existing code, while the second one potentially fixes an existing problem.  (What if none of the subblades have offset == 0?)

The second is probably leaner code-wise, so maybe go with that?

NoSloppy: ok. How about now?

PR #750, opened 2025-03-03
BCprop - just play sound

NoSloppy:
For entering Battle Mode, don't actually do EFFECT_FORCE , just play the sound if need be.


profezzorn:
Why not do EFFECT_FORCE?

NoSloppy:
> Why not do EFFECT_FORCE?

If there's a blade effect for force, it's not really meant to be shown here. This is just a fallback sound use.

PR #752, opened 2025-03-04
Now includes array saving.

Sabersense-CC:
Adds extra save file to SD root (arraysve) which stores the array index, meaning you boot back in to the last array you were using.

### DIFF
--- a/props/saber_sabersense_buttons.h
+++ b/props/saber_sabersense_buttons.h
@@ -1,4 +1,4 @@
-/* V.8-154.
+/* V.8-155.
 ============================================================
 ================   SABERSENSE PROP FILE   ==================
 ================            by            ==================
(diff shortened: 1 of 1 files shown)


Sabersense-CC:
It just occurred to me, do we want the save option wrapped in its own #define? If so, am I allowed to have it enabled by default with a #ifndef DISABLE_ARRAY_SAVE, or does is have to be an explicit #ifdef ENABLE_ARRAY_SAVE?

profezzorn:
> It just occurred to me, do we want the save option wrapped in its own #define? If so, am I allowed to have it enabled by default with a #ifndef DISABLE_ARRAY_SAVE, or does is have to be an explicit #ifdef ENABLE_ARRAY_SAVE?

You can do whichever one you want, but it's gotta start with SABERSENSE_

Sabersense-CC:
Of course. Understood. Will get it sorted later today.

props/saber_sabersense_buttons.h
```diff
@@ -532,7 +564,15 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
       SaberBase::DoNewFont();  // Play font ident if 'array' sound file missing.
     }
 #endif
+#ifndef SABERSENSE_DISABLE_SAVE_ARRAY
+    SaveArrayState();  // After changing array, save new array index.
+  }
```
profezzorn, line 569: This is super messy. Please have your #if #else #endif stay within one function
(or wrap an entire function.)

PR #755, opened 2025-03-07
jetpack_Oli_buttons.h

olivierflying747-8:
I have not yet been able to test it because my Proffie board is still not yet connected but I am hoping for my setup to be ready by end March 2025. I know I said end February but it is taking me a lot longer then I thought to design a chassis that I will be happy with.

I am hoping for a head-start on the code review as I am hoping not to miss the OS8.x launch.

Thank you for your time.

I also have the "jetpackdisplaycontroler" but I will wait until it can be integrated in the prop directly rather than in ssd1306.h

### DIFF
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -1080,21 +1080,21 @@ public:
     CONFIG_VARIABLE2(maxclash, 16);
 #endif
 }
-  bool gestureon; // gesture controls on/off for use with "Gesture Sleep"
-  bool swingon; // Swing On ignition
-  int swingonspeed; // Swing On speed (200 ~ 500 range)
-  bool twiston; // Twist On ignition
-  bool thruston; // Thrust On ignition
-  bool stabon; // Stab On ignition
-  bool forcepush; // Force Push
-  uint32_t forcepushlen; // Force Push Length
-  uint32_t lockupdelay; // Lockup Delay (for Battle Mode)
-  bool twistoff; // Twist Off retraction
-  // disable PWR button for retraction, for use with "Power Lock" mode
-  // to prevent button turning saber off
+  bool gestureon;         // gesture controls on/off for use with "Gesture Sleep"
+  bool swingon;           // Swing On ignition
+  int swingonspeed;       // Swing On speed (200 ~ 500 range)
+  bool twiston;           // Twist On ignition
+  bool thruston;          // Thrust On ignition
+  bool stabon;            // Stab On ignition
+  bool forcepush;         // Force Push
+  uint32_t forcepushlen;  // Force Push Length
+  uint32_t lockupdelay;   // Lockup Delay (for Battle Mode)
+  bool twistoff;          // Twist Off retraction
+                          // disable PWR button for retraction, for use with "Power Lock" mode
+                          // to prevent button turning saber off
   bool powerlock;
-  int clashdetect; // maximum Clash Strength to detect Clash during Battle Mode (0 ~ 10 range)
-  int maxclash; // maximum Clash Strength for Clash Sound and Detection works with CLASH_THRESHOLD_G to create range of Clash Strength (8 ~ 16 range)
+  int clashdetect;        // maximum Clash Strength to detect Clash during Battle Mode (0 ~ 10 range)
+  int maxclash;           // maximum Clash Strength for Clash Sound and Detection works with CLASH_THRESHOLD_G to create range of Clash Strength (8 ~ 16 range)
 };
 
 GestureControlFile saved_gesture_control;
@@ -2536,7 +2536,7 @@ SaberFett263Buttons() : PropBase() {}
           SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
           SaberBase::DoBeginLockup();
           auto_lockup_on_ = true;
-	} else {
+        } else {
           SaberBase::DoClash();
         }
         clash_type_ = CLASH_NONE;
@@ -2603,7 +2603,7 @@ SaberFett263Buttons() : PropBase() {}
         if (!swinging_ && fusor.swing_speed() > saved_gesture_control.swingonspeed) {
           swinging_ = true;
           Event(BUTTON_NONE, EVENT_SWING);
-	}
+        }
       }
     }
     // EVENT_THRUST
@@ -4430,18 +4430,18 @@ SaberFett263Buttons() : PropBase() {}
         menu_sub_pos_ += direction;
         if (menu_sub_pos_ <= 0) menu_sub_pos_ = 2;
         if (menu_sub_pos_ > 2) menu_sub_pos_ = 1;
-	switch (menu_sub_pos_) {
-	  case EDIT_STYLE_SELECT:
+        switch (menu_sub_pos_) {
+          case EDIT_STYLE_SELECT:
             sound_library_.SaySelectStyle();
-	    break;
+            break;
           case EDIT_STYLE_SETTINGS:
             sound_library_.SayEditStyleSettings();
             break;
-	  default:
+          default:
             sound_library_.SaySelectOption();
-	    break;
-	}
-	break;
+            break;
+        }
+        break;
       case MENU_STYLE:
         ChangeStyleNumber(blade_num_, direction);
         current_preset_.Save();
@@ -4594,7 +4594,7 @@ SaberFett263Buttons() : PropBase() {}
           if (direction > 0) {
             sound_library_.SayUp();
           } else {
-	    sound_library_.SayDown();
+            sound_library_.SayDown();
           }
 #endif
           break;
@@ -4636,7 +4636,7 @@ SaberFett263Buttons() : PropBase() {}
         }
         switch (set_num_) {
           case STYLE_OPTION_ARG:
-	    sound_library_.SayStyleOptions();
+            sound_library_.SayStyleOptions();
             break;
           case STYLE_OPTION2_ARG:
             sound_library_.SayStyleOptions();
@@ -4647,14 +4647,14 @@ SaberFett263Buttons() : PropBase() {}
             sound_library_.SayNumber(3, SAY_WHOLE);
             break;
           case IGNITION_OPTION_ARG:
-	    sound_library_.SayIgnitionOptions();
+            sound_library_.SayIgnitionOptions();
             break;
           case IGNITION_OPTION2_ARG:
             sound_library_.SayIgnitionOptions();
             sound_library_.SayNumber(2, SAY_WHOLE);
             break;
           case IGNITION_TIME_ARG:
-	    sound_library_.SayIgnitionTime();
+            sound_library_.SayIgnitionTime();
             break;
           case IGNITION_POWER_UP_ARG:
             sound_library_.SayPowerUpOptions();
@@ -4663,14 +4663,14 @@ SaberFett263Buttons() : PropBase() {}
             sound_library_.SayIgnitionDelay();
             break;
           case RETRACTION_OPTION_ARG:
-	    sound_library_.SayRetractionOptions();
+            sound_library_.SayRetractionOptions();
             break;
           case RETRACTION_OPTION2_ARG:
             sound_library_.SayRetractionOptions();
             sound_library_.SayNumber(2, SAY_WHOLE);
             break;
           case RETRACTION_TIME_ARG:
-	    sound_library_.SayRetractionTime();
+            sound_library_.SayRetractionTime();
             break;
           case RETRACTION_COOL_DOWN_ARG:
             sound_library_.SayCoolDownOptions();
@@ -4679,10 +4679,10 @@ SaberFett263Buttons() : PropBase() {}
             sound_library_.SayRetractionDelay();
             break;
           case LOCKUP_POSITION_ARG:
-	    sound_library_.SayClashLockupPosition();
+            sound_library_.SayClashLockupPosition();
             break;
           case DRAG_SIZE_ARG:
-	    sound_library_.SayDragSize();
+            sound_library_.SayDragSize();
             break;
           case MELT_SIZE_ARG:
             sound_library_.SayMeltSize();
@@ -4691,16 +4691,16 @@ SaberFett263Buttons() : PropBase() {}
             sound_library_.SaySwingOption();
             break;
           case EMITTER_SIZE_ARG:
-	    sound_library_.SayEmitterSize();
+            sound_library_.SayEmitterSize();
             break;
           case OFF_OPTION_ARG:
             sound_library_.SayOffOption();
             break;
           case PREON_OPTION_ARG:
-	    sound_library_.SayPreonOptions();
+            sound_library_.SayPreonOptions();
             break;
           case PREON_SIZE_ARG:
-	    sound_library_.SayPreonSize();
+            sound_library_.SayPreonSize();
             break;
           default:
             sound_library_.SaySelectOption();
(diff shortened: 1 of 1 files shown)


profezzorn:
Can't we just call it "hev.h" ?

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,931 @@
...
+EFFECT2(stuttering,flightmode);      // jetpack stuttering sound  (in flight)
+EFFECT(selfdestruct);                // jetpack exploding sound   (pre-idle)
+EFFECT(meltdown);                    // jetpack melt-down sound   (pre-idle)
+EFFECT(dud);                         // jetpack not exploding or melting down sound "Dank farrik, it's a dud!"
+// == sounds for missile ==
+EFFECT(aiming);                      // viewfinder coming down "click" sound
+EFFECT(targetting);                  // viewfinder search/finds target sound
+EFFECT(missilelaunch);               // missile launch sound
+EFFECT(missilegoesboom);             // double explosion in the distance sound
+EFFECT(mandotalk);                   // "Nice shot! I was aiming for the other one!
```
profezzorn, line 296: What's wrong with "quote" ?

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,931 @@
...
+        blade_detected_ = false;
+        FindBladeAgain();
+        SaberBase::DoBladeDetect(false);
+        return true;
+#endif // BLADE_DETECT_PIN
+      // Jetpack ON AND Volume Up:
+      case EVENTID(BUTTON_POWER, EVENT_FIRST_SAVED_CLICK_SHORT, MODE_OFF):
+      case EVENTID(BUTTON_POWER, EVENT_FIRST_SAVED_CLICK_SHORT, MODE_ON):
+        if (!mode_volume_) {
+          if       (idle_) StartFlightMode(); // Transition from idle to flight  (rev more up)
```
profezzorn, line 528: This is not how you make an if statement easy to read, please remove extra spaces and put the function call on the next line.

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,931 @@
...
+          else if (flight_) StopFlightMode(); // Transition from flight to idle     (rev down)
+          else               StartIdleMode(); // Jetpack initial start from 0 to idle (rev up)
+          // if idle, it revs up / if up, it revs down to idle / if not idle or not up, it needs to rev "up" to idle.
+        } else VolumeUp();
```
profezzorn, line 532: If you have a } before the else, you should have one after the else too.

```
} else {
  VolumeUp();
}
```

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,931 @@
...
+        state_machine_.reset_state_machine();
+        PVLOG_NORMAL << "Missile Sequence Aborted.\n";
+        SaberBase::DoEffect(EFFECT_DISARM,0);
+        return true;
+
+      case EVENTID(BUTTON_POWER, EVENT_FIRST_CLICK_LONG, MODE_OFF):
+      // Next Preset (1 button):
+#if NUM_BUTTONS == 1
+        if (!mode_volume_) next_preset();
+        return true;
```
profezzorn, line 563: Since you have return true at the end of both the #if and the #else, you should just move it out to after the #endif.

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,931 @@
...
+#else
+        // Start or Stop Track (2 and 3 buttons):
+#ifdef JETPACK_TRACK_PLAYER
+        if (!istrack_playing_) StartTrackPlayer();
```
profezzorn, line 567: Don't do this:

```
if (A) B();
else C();
```
Do this instead:
```
if (A)
  B();
else
  C();
```

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,931 @@
...
+        else StopTrackPlayer();
+#else
+        StartOrStopTrack();
+#endif
+        return true;
+#endif // NUM_BUTTONS == 1
+      case EVENTID(BUTTON_POWER, EVENT_FIRST_CLICK_LONG, MODE_ON):
+      // Volume Down (1 button):
```
profezzorn, line 575: indent two more spaces

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,931 @@
...
+        }
+        return true;
+
+      // Color Change mode (1 button):
+#if NUM_BUTTONS == 1
+      case EVENTID(BUTTON_POWER, EVENT_THIRD_SAVED_CLICK_SHORT, MODE_ON):
+#ifndef DISABLE_COLOR_CHANGE
+        ToggleColorChangeMode();
+#endif
+        return true;
```
profezzorn, line 618: You probably shouldn't return true here if DISABLE_COLOR_CHANGE is defined, since it doesn't actually do anything in that case.

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,931 @@
...
+      // Color Change mode (2 buttons):
+#elif NUM_BUTTONS == 2
+#ifndef DISABLE_COLOR_CHANGE
+      case EVENTID(BUTTON_AUX, EVENT_FIRST_CLICK_SHORT, MODE_ON):
+        ToggleColorChangeMode();
+        return true;
+#endif
+#else
+      // Color Change mode (3 buttons):
+#ifndef DISABLE_COLOR_CHANGE
```
profezzorn, line 628: Seems like it would be better to do:
```
#ifndef DISABLE_COLOR_CHANGE
#if NUM_BUTTONS == 1
#elif NUM_BUTTONS == 2
#elif ...
#endif
#endif
```

instead of:

```
#if NUM_BUTTONS==1
#ifndef DISABLE_COLOR_CHANGE
#endif
#elif NUM_BUTTONS==2
#ifndef DISABLE_COLOR_CHANGE
#endif
#ifndef NUM_BUTTONS==3
#ifndef DISABLE_COLOR_CHANGE
#endif
#endif
```

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,931 @@
...
+  void DoMotion(const Vec3&, bool) override { }  // jetpack doesn't have "movement effects", you
+    // do not want to have to do front/back/side flips at your next cosplay to activate someting!
+    // Or just imagine, for an instant, someone trying to replicate stab, aka running face first
+    // into the nearest wall with their backpack/jetpack prop.
+
+  // Helper (to avoid repetition)
+  void EffectHelper (Effect* effect1,const char* msg1) {
+    hybrid_font.PlayPolyphonic(effect1);
+    PVLOG_STATUS << msg1 << "\n";
+    return;
```
profezzorn, line 854: return; at the end of a function is redundant

profezzorn (commented):
first pass

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+    Loop Function:
+        Monitors the 1-minute timer (or user defined) during idle mode to turn off the jetpack
+        completely if the time expires.
+        Plays idle or flight mode on loop for the required time.
+
+    To do: create animations for OLED for Missile Launch (based on Mando S02E06 - if you know, you know!)
+    & make a fun jetpack sound font.
+    Create DIM "mode" for "blades" in idle mode.
+
+*/
```
profezzorn, line 58: Why mess with separate comments, just have one large comment.

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+#ifndef PROPS_JETPACK_OLI_BUTTONS_H
+#define PROPS_JETPACK_OLI_BUTTONS_H
+
+#if NUM_BUTTONS < 1
+#error Your prop NEEDS 1 or more buttons to use jetpack_prop (2 is recomended)
+#endif
+
+#include "prop_base.h"
+
+#ifdef INCLUDE_SSD1306
```
profezzorn, line 124: use PROP_BOTTOM

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+#ifndef EXTRA_DISPLAY_CONTROLLER_INCLUDE
+#define EXTRA_DISPLAY_CONTROLLER_INCLUDE "props/additional_display_controllers.h"
+#endif
+#endif
+
+#define PROP_TYPE JetpackOliButtons
+
+#ifndef JETPACK_IDLE_TIME
+#define JETPACK_IDLE_TIME 1000 * 60 * 1 // Jetpack max idle mode time in millisec (default 1 min) before auto-shutdown
```
profezzorn, line 133: line is long, put comment before define.
Also might be worth pointing out how this is different from IDLE_OFF_TIME

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+#define HAVE_VMBEGIN    // for Begin Volume Menu
+#define HAVE_VMEND      // for End Volume Menu
+#define HAVE_VOLDOWN    // for decrease volume
+#define HAVE_VOLUP      // for increase volume
+#define HAVE_VOLMAX     // for maximum volume reached
+#define HAVE_VOLMIN     // for minimum volume reached
+#endif
+
+#ifdef PROPS_BLASTER_BC_BUTTONS_H
+#ifndef HAVE_BATTERY
```
profezzorn, line 205: Do you get warnings if you remove these #ifndef ?

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+EFFECT(volup);          // for increase volume
+#endif
+#ifndef HAVE_VOLMAX
+EFFECT(volmax);         // for maximum volume reached
+#endif
+#ifndef HAVE_VOLMIN
+EFFECT(volmin);         // for minimum volume reached
+#endif
+
+#ifdef HAVE_BATTERY
```
profezzorn, line 256: I think it's fine to #undef something that isn't defined, so the #ifdef isn't needed I think.

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+#endif
+#ifdef HAVE_VOLMAX
+#undef HAVE_VOLMAX      // for maximum volume reached
+#endif
+#ifdef HAVE_VOLMIN
+#undef HAVE_VOLMIN      // for minimum volume reached
+#endif
+
+// == sounds for jetpack ==
+EFFECT2(idlemode,idlemode);          // jetpack idle sound        (looping idle sound)
```
profezzorn, line 285: We have an EFFECT(idle) already (enabled with a define)
Can't we use that?

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+  void PreviousTrack() {
+    if (num_tracks_ <= 0) {
+      PVLOG_DEBUG << "No tracks available.\n";
+      return;
+    }
+    track_num_ = (track_num_ - 2 + num_tracks_) % num_tracks_ + 1; // Cycle to the previous track
+    PlayTrack();
+  }
+
+// only one prop can have chdir & Parse, so in a dual_prop.h/multi_prop.h "environment" SaberFett263Button already has them.
```
profezzorn, line 475: That's not true. multiprop can choose which one to use.

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+      case EVENTID(BUTTON_BLADE_DETECT, EVENT_LATCH_OFF, MODE_ANY_BUTTON | MODE_OFF):
+        blade_detected_ = false;
+        FindBladeAgain();
+        SaberBase::DoBladeDetect(false);
+        return true;
+#endif // BLADE_DETECT_PIN
+      // Jetpack ON AND Volume Up:
+      case EVENTID(BUTTON_POWER, EVENT_FIRST_SAVED_CLICK_SHORT, MODE_OFF):
+      case EVENTID(BUTTON_POWER, EVENT_FIRST_SAVED_CLICK_SHORT, MODE_ON):
+        if (!mode_volume_) {
```
profezzorn, line 533: This sort of uglyness can be avoided by using the new mode system for menus. (optional)

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+        talkie.Say(spPOINT);
+        talkie.SayDigit(((int)floorf(battery_monitor.battery() * 10)) % 10);
+        talkie.SayDigit(((int)floorf(battery_monitor.battery() * 100)) % 10);
+        talkie.Say(spVOLTS);
+        return true;
+    }  // switch (EVENTID)
+    return false;  // No action
+  }  // Event2
+
+  // I want the effects to play one after the other without overlapping or interrupting one another.
```
profezzorn, line 719: Ok, you want something, but what does that mean?
(This comment doesn't seem very helpful.)

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+    SaberBase::DoEffect(EFFECT_STARTFLIGHTMODE, 0); // start transition to flight mode
+    flight_ = true;
+    idle_   = false;
+    PVLOG_STATUS << "Jetpack Flying\n";
+  }
+
+  // Transition from Flying Mode (from flying to idle)
+  void StopFlightMode() {
+    if (jetpack_wav_player_.isPlaying()) {
+      return; // Wait for the current sound to finish
```
profezzorn, line 809: This comment is misleading, as there is no waiting involved.
This just disables this function if something else is happening, which seems like it could lead to bad user experiences.
(I want to stop, why isn't it working???)

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+    }
+    SaberBase::DoEffect(EFFECT_STOPFLIGHTMODE, 0); // start transition from flight mode
```
profezzorn, line 811: Please consider changing the ALT mode when transitioning to a different mode.
This makes it possible to alter *all* sounds based on the current mode.

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+
+  // Stop Idle Mode (Jetpack completely off - idle to off)
+  void StopIdleMode() {
+    if (jetpack_wav_player_.isPlaying()) {
+      return; // Wait for the current sound to finish
+    }
+    SaberBase::DoEffect(EFFECT_STOPIDLEMODE, 0); // start transition from idle mode
+    flight_ = false; // Should already be false but StopIdleMode can be used to turn the jetpack completely off from any state.
+    idle_   = false;
+    Off(OFF_FAST); // Fully stop the jetpack
```
profezzorn, line 824: There is also an OFF_IDLE, does that apply here?

props/jetpack_Oli_buttons.h
```diff
@@ -0,0 +1,945 @@
...
+  In case of problem, you can find us at: https://crucible.hubbe.net somebody will be there to help.
+  Distributed under the terms of the GNU General Public License v3.
+  http://www.gnu.org/licenses/
+
+Includes 1, 2 and 3 button controls. However, I would recommend a 2 or 3
+buttons setup because I think a 1 button setup is "cumbersome" to use!
+
+Buttons handeling was initially based on saber_sa22c_buttons.h
+
+Explanation:
```
profezzorn, line 19: It would be helpful to have an explanation of what the different "modes" are.

profezzorn:
Has this been actually tested yet?

PR #757, opened 2025-03-07
Multi prop

olivierflying747-8:
Just like my jetpack, it is not tested yet. Hopefully by the end March 2025, I will have a working saber to test it on.

Thanks in advance for an early review.

I initially added my multipropdisplaycontroller.h but I think it will be better in a separate PR (coming soon).

### DIFF
--- a/buttons/button_base.h
+++ b/buttons/button_base.h
@@ -27,6 +27,9 @@
 #define BUTTON_HELD_LONG_TIMEOUT 2000
 #endif
 
+#ifndef BUTTON_LONG_CLICK_TIMEOUT
+#define BUTTON_LONG_CLICK_TIMEOUT 2500
+#endif
 
 // Simple button handler. Keeps track of clicks and lengths of pushes.
 class ButtonBase : public Looper,
@@ -109,7 +112,7 @@ protected:
         current_modifiers &=~ button_;
         if (millis() - push_millis_ < BUTTON_SHORT_CLICK_TIMEOUT) {
           SendClick(EVENT_CLICK_SHORT);
-        } else if (millis() - push_millis_ < 2500) {
+        } else if (millis() - push_millis_ < BUTTON_LONG_CLICK_TIMEOUT) {
 	  // Long clicks cannot be "saved", so just emit immediately.
           Send(EVENT_CLICK_LONG);
         }


common/events.h
```diff
@@ -51,6 +51,7 @@ enum EVENT : uint32_t {
   EVENT_HELD,
   EVENT_HELD_MEDIUM,
   EVENT_HELD_LONG,
+  EVENT_HELD_XTRA_LONG,
```
profezzorn, line 54: Please calculate this in your prop instead.
I don't think we need more "HELD" events.
Actually, it looks like you *are* calculating this in the prop, so why add it here??

props/multi_prop.h
```diff
@@ -0,0 +1,501 @@
...
+#ifndef TWO_BUTTONS_X_LONG_PUSH
+#define TWO_BUTTONS_X_LONG_PUSH 1000 * 3.9
+#endif
+
+// Button combo state management
+unsigned long holdStartTime = 0;
+Prop_Mode currentMode = Prop_Mode::SABER;  // Initial state is Saber
+
+// Helper function to set FakeBladeID and re-detect the blade
+template <class PropBase>
```
profezzorn, line 123: Why are these functions outside of the prop class?

props/multi_prop.h
```diff
@@ -0,0 +1,501 @@
...
+      }
+      //  Reset  the  hold  state  if  either  button  is released
+      if (!powerPressed_ || !auxPressed_) {
+        holdStartTime = 0;
+        PVLOG_NORMAL << "One of the two (or both) button(s) was (were) released too early!\n";
+      }
+    }
+    // Check if both buttons have been held for the required duration
+    if (powerPressed_ && auxPressed_ && (millis() - holdStartTime) >= TWO_BUTTONS_X_LONG_PUSH) {
+      switchModes(this);               // Pass "this" MultiProp instance to switchModes & switch the prop mode
```
profezzorn, line 255: This comment is useless. The code already says the same thing.
A lot of comments in this code explains what the code does, but that's not the purpose of comments. The comments should focus on explaining *why* the code does what it does.
In most cases, what the code does is obvious from reading the code.

props/multi_prop.h
```diff
@@ -0,0 +1,501 @@
...
+      case Prop_Mode::JETPACK:   return   Jetpack::Event(button, event);
+    //case Prop_Mode::MORSECODE: return MorseCode::Event(button, event);
+    //case Prop_Mode::DROID:     return     Droid::Event(button, event);
+    //case Prop_Mode::VEHICLE:   return   Vehicle::Event(button, event);
+    }
+    return false;  // Event was not handled
+  }
+
+/******************************************************************************************\
+| I tried with lamdbas functions to reduce repetition but it was taking a lot more memmory.|
```
profezzorn, line 289: It's probably because you pulled in std::unordered_map, which is a large and complicated template, and also not required. A simple array would work better. Also, library functions that use "new" are generally no-no in ProffieOS as memory fragmentation can cause ProffieOS to run out of memory very quickly. 

In general, I'm a bit skeptical of the lambda approach. It seems like a lot of complexity for very little gain, and then it makes it harder if we actually need to fiddle with the arguments like we do in Event().

profezzorn (commented):
first pass

props/multi_prop.h
profezzorn: > None of the lambda's stuff was done by me, but by ChatGPT. 

Nope.

You put it in the PR. I don't care where it came from once you put it in the PR, it is your responsibility, not ChatGPTs.

profezzorn:
tested yet?

PR #758, opened 2025-03-07
Display controller for multi prop & jetpack prop

olivierflying747-8:
Only worth reviewing if jetpack and multiprop are going to be added.


display/ssd1306.h
```diff
@@ -947,6 +947,12 @@ class StandardDisplayController : public DisplayControllerBase<Width, col_t>, pu
   volatile Effect* current_effect_;
 };
 
+// The content of multipropdisplaycontroller.h should go in the respective prop files
+// but it is not yet possible at the moment!
+#ifdef PROPS_MULTI_PROP_H
+#include "multipropdisplaycontroller.h"
```
profezzorn, line 953: I think what we should do is to put the display controller file in the prop directory.
Then we'll add a define in the prop file, something like:

```cpp
#define EXTRA_DISPLAY_CONTROLLER_INCLUDE "props/multiprop_display_controller.h"
```

Then in ProffieOS.ino, after including display/ssd1306.h, we'll put:

```cpp
#ifdef EXTRA_DISPLAY_CONTROLLER_INCLUDE
#include EXTRA_DISPLAY_CONTROLLER_INCLUDE
#endif
```

props/additional_display_controllers.h
```diff
@@ -0,0 +1,398 @@
...
+  MultiPropDisplayEffects<PREFIX> img_;
+  MultiPropDisplayConfigFile &multiprop_font_config;
+  MultiPropDisplayController() :
+    img_(*getPtr<MultiPropDisplayEffects<PREFIX>>()),
+    multiprop_font_config(*getPtr<MultiPropDisplayConfigFile>()) {}
+
+  void SB_Effect2(EffectType effect, EffectLocation location) override {
+    switch (effect) {
+      case EFFECT_SABERMODE:
+        if (img_.IMG_sabermode)
```
profezzorn, line 78: This is not an ok way to indent, it should look like:

```
if (...) {
  ShowFile...
} else {
   this->SetMessage...
   this->SetScreenNow..
}
break;
```

props/additional_display_controllers.h
```diff
@@ -0,0 +1,398 @@
...
+  ShowFileWithSoundLength(&img_.IMG_blastermode,   multiprop_font_config.ProffieOSBlasterModeImageDuration);
+        else {
+  this->SetMessage("blaster/n mode");
+  this->SetScreenNow(SCREEN_MESSAGE);
+       } break;
+      case EFFECT_DETONATORMODE:
+        if (img_.IMG_detonatormode)
+  ShowFileWithSoundLength(&img_.IMG_detonatormode, multiprop_font_config.ProffieOSDetonatorModeImageDuration);
+       else {
+  this->SetMessage("detonator/n  mode");
```
profezzorn, line 95: /n ? Did you mean \n ?

props/additional_display_controllers.h
```diff
@@ -0,0 +1,398 @@
...
+        if (img_.selfdestruct) // Jetpack exploding animation
+  ShowFileWithSoundLength(&img_.selfdestruct,        jetpack_font_config.ProffieOSSelfDestructImageDuration);
+        else {
+  this->SetMessage("jetpack\nself destruct");
+  this->SetScreenNow(SCREEN_MESSAGE);
+       } break;
+      case EFFECT_MELTDOWN:
+        if (img_.meltdown) // Jetpack melting animation
+  ShowFileWithSoundLength(&img_.meltdown,            jetpack_font_config.ProffieOSMeltdownImageDuration);
+        else {
```
profezzorn, line 288: It should always be "} else {" not "else {" or "} else".

props/additional_display_controllers.h
```diff
@@ -0,0 +1,398 @@
...
+  this->SetMessage("jetpack\nmelting");
+  this->SetScreenNow(SCREEN_MESSAGE);
+       } break;
```
profezzorn, line 291: break should go on a separate line, usually with an empty line after it.

props/additional_display_controllers.h
profezzorn: If both sides are short one-liners, then "else" (no brackets) is also acceptable.

profezzorn:
Ok, I've made it so that the display controller can go in the prop file.
It just needs to go in the PROP_BOTTOM section. See blaster.h for an example.

profezzorn:
That is correct.

PR #759, opened 2025-03-08
Add SD_ACCESS to Edit Mode Menu

fdarosa2663:
adding support for MOUNT_SD_SETTING to Edit Mode Menu


props/saber_fett263_buttons.h
```diff
@@ -1797,6 +1797,7 @@ SaberFett263Buttons() : PropBase() {}
 
   void Setup() override {
     RestoreGestureState();
+    sound_library_.init();
```
profezzorn, line 1800: The idea is to call init on the highest version of the sound library you actually use, so this should probably be:

```
#ifdef MOUNT_SD_SETTING
   sound_library_v2.init();
#else
   sound_library_.init();
#endif
```

fdarosa2663: Updated.

PR #762, opened 2025-03-11
Create hev.h

jimsworld:
First commit.
This file is the most up to date version since our last PM.


jimsworld:
> Since it is a prop, it should probably go in \props.

Apologies, yes you're right. Let me rectify this. This is my first commit on another repo.

> Out of curiosity, what does hev stand for ?

Check out the thread I made on it!
https://crucible.hubbe.net/t/half-life-hev-suit-replica-costume/6386
Always keen to hear new concepts that could be implemented down the line.

props/hev.h
```diff
@@ -0,0 +1,417 @@
...
+
+    // Enforce minimum values
+    if (health_ < 0) health_ = 0;
+    if (armor_ < 0) armor_ = 0;
+
+    // Damage
+    if (!quiet) SaberBase::DoEffect(EFFECT_STUN, 0.0, damage);
+    
+    // Death Sound
+    if (health_ == 0 && !dead_) { // Check flag to avoid death sound spam
```
profezzorn, line 85: Instead of using dead_, you can just save the health_ into a local variable at the beginning, like:

  int health = health_;

Then this if statement just becomes:

   if (health_ == 0 && health > 0)

props/hev.h
```diff
@@ -0,0 +1,417 @@
...
+
+    // Stop Clashes if health is 0
+    if (health_ == 0) {
+      return;
+    }
+
+     // Strength multiplier
+    int damage = (int)(strength * 4);
+
+    // Play Clash sounds based on strength
```
profezzorn, line 136: FYI: This might not work super well, because the clash event usually triggers before the full strength of the clash is known.

props/hev.h
```diff
@@ -0,0 +1,417 @@
...
+    SFX_clash.SelectFloat(v);
+    SFX_clsh.SelectFloat(v);
+    SFX_stab.SelectFloat(v);
+
+    // Cap damage at 50
+    if (damage > 50) {
+      damage = 50;
+    }
+
+    // Play Major or Minor Detected Voice Lines
```
profezzorn, line 147: Why is this commented out?
Use an #ifdef maybe? Or just delete it?

props/hev.h
```diff
@@ -0,0 +1,417 @@
...
+      // return;
+  
+	// Random Hazard Sounds
+      case EFFECT_STUN:
+	hybrid_font.PlayCommon(&SFX_stun);
+	return;
+	
+	// Health Alert
+      case EFFECT_USER1:
+  if (dead_) return; // Don't queue health sounds if dead
```
profezzorn, line 402: bad indentation

props/hev.h
```diff
@@ -0,0 +1,417 @@
...
+public:
+  Hev() : PropBase() {}
+  const char* name() override { return "Hev"; }
+
+  int health_ = 100;
+  int armor_ = 100;
+
+  bool dead_ = false;
+  bool armor_depleted_ = false;
+  bool morphine_triggered_ = false;
```
profezzorn, line 44: This is not used anywhere.

profezzorn (commented):
First pass

props/hev.h
```diff
@@ -0,0 +1,417 @@
...
+      PropBase::LowBatteryOff();
+    }
+  }
+
+
+  // Calculate Physical Damage //////////////////////////////////////////////////
+  void DoDamage(int damage, bool quiet = false) {
+    int tens = health_ / 10;
+    if (armor_ >= damage) {
+        armor_ -= ceilf(damage * 0.80 / 2);  // Applies 80% divided by 2 of damage to armor, if armor is enough.
```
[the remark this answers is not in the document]
jimsworld: Fixed and fixed.

props/hev.h
jimsworld: I've deleted for now. Will revisit the functionality later down the line.

props/hev.h
jimsworld: Fixed.

props/hev.h
jimsworld: Removed. Will revisit the morphine voice line later.

props/hev.h
```diff
@@ -0,0 +1,417 @@
...
+
+class Hev : public PROP_INHERIT_PREFIX PropBase {
+public:
+  Hev() : PropBase() {}
+  const char* name() override { return "Hev"; }
+
+  int health_ = 100;
+  int armor_ = 100;
+
+  bool dead_ = false;
```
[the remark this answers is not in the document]
jimsworld: Yes I guess it is. I'm hesitant to remove this as it's used in other parts.

Should I:

1. Remove it
2. For every method that uses `dead_`, should I remove `dead_` and add a local variable at the start, like:

`int health = health_;`?

props/hev.h
jimsworld: Please ref my replies on `dead_` and `armor_depleted_` as they're at the root of this use case.

The change seems simple, but would need to clarify everywhere it's used to make sure the functionality stays the same.

props/hev.h
jimsworld: I'm not too sure what to suggest here. Although, I can see that clash damage may not be calculated correctly, or the clash wav isn't chosen correctly.

However I can confirm that it seems to work okay from more than 20 minutes of bashing tests.

I think this may need to be revisited and revised to ascertain whether the clash damage and clash wav are being accurately calculated...

props/hev.h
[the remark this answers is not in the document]
profezzorn: in most places, you would just replace dead_ with health == 0
It's just DoDamage that needs a before-and-after-check.

props/hev.h
profezzorn: If it works, that's good.
I just wanted to warn you that it might not.

props/hev.h
profezzorn: Doesn't seem fixed.

PR #763, opened 2025-03-15
Fix IsGreaterThan

fdarosa2663:
This is actually a bug in OS7 that was missed, IsGreaterThan was working identical to IsLessThan.


fdarosa2663:
Something else is going on, ignore.

PR #778, opened 2025-03-20
make BCprop compile again

NoSloppy:
- Fix BladeID if BLADE_ID_SCAN_MILLIS not defined.
- Also, add 2 POD links for menu info.
- 1 trailing whitespace removed.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -92,7 +92,7 @@ Optional #defines:
 * NEW as of OS8:
 #define MENU_SPEC_TEMPLATE         - BC Volume Menu, Scroll Presets, ColorChange, and BC Blade Length adjust are used by default.
                                      Use this define to override and add access the the OS System Menu for editing presets, colors etc...
-                                     Several template choices are available to set how the menus are used. See Documentation.
+                                     Several template choices are available to set how the menus are used. See https://pod.hubbe.net/howto/menus.html.
 #define DYNAMIC_BLADE_LENGTH       - This is required for onboard menu driven blade length adjustments.
                                      The blade length maximum should be set in the user config file's BladeConfig section.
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -2057,13 +2071,16 @@ void DoSavedTwist() {
...
     if (!use_fake_id_) best_config_before_faking_ = current_config - blades;
-    size_t next_array = (current_config - blades + 1) % NELEM(blades);
-    fake_id_ = blades[next_array].ohm;
+    fake_id_ = blades[((current_config - blades) + 1) % NELEM(blades)].ohm;
     use_fake_id_ = true;
-
     FindBladeAgain();
     PlayArraySound();
+  #else
+    current_config = blades + (((current_config - blades) + 1) % NELEM(blades));
```
profezzorn, line 2081: This won't work.
This doesn't tear down blades, change directory or do any of the things that FindBladeAgain() does.

Also, I'm not sure what BLADE_ID_SCAN_MILLIS has to do with it... It should work anyways, no?

NoSloppy: ugh I left out FindBladeAgain(). I meant to have both calls outside the #endif.
However, that's not all the problem.

This PR is a result of me not testing thoroughly enough.
The way it stands, if BLADE_ID_SCAN_MILLIS is not defined, BCprop tries to override GetNoBladeLevelBefore(), but that doesn't exist.
If just adding a guard to my override to try to fix that, NextBladeArray() breaks.
 
Also, I hadn't tested using multiple preset arrays and different fonts, so it "worked" as I first submitted here, but then found the issues.
 
I've since refactored it, tested "all" the scenarios, and it seems correct now.
Does it look right / make sense?

PR #781, opened 2025-03-21
SpeakBladeId() to use sound lib?

NoSloppy:


props/prop_base.h
```diff
@@ -527,6 +527,10 @@ class PropBase : CommandParser, Looper, protected SaberBase, public ModeInterfac
   }
 
   virtual void SpeakBladeID(float id) {
+    if (&SFX_mnum) {
```
profezzorn, line 530: Why?
Speak blade id is a debug function, it doesn't need to sound good.
Also, it doesn't  compile, because the { } are unbalanced.

NoSloppy:
Ok that's right. This has come up in the past.

PR #782, opened 2025-03-25
Add TransitionLoopWhileL<>

fdarosa2663:


transitions/loop.h
```diff
@@ -103,4 +103,69 @@ class TrLoopUntil {
...
+  PONUA SVFWrapper<CONDITION> condition_;
+
+public:
+auto getColor(int led) -> decltype(
+    MixColors(end_tr_.getColor(loop_tr_.getColor(RGBA_um_nod::Transparent(), RGBA_um_nod::Transparent(), led), RGBA_um_nod::Transparent(), led),
+        loop_tr_.getColor(loop_tr_.getColor(RGBA_um_nod::Transparent(), RGBA_um_nod::Transparent(), led), RGBA_um_nod::Transparent(), led), 1, 1)) {
+    if (run_) {
+      return loop_tr_.getColor(RGBA_um_nod::Transparent(),
+          RGBA_um_nod::Transparent(), led);
+    } else if (end_) {
```
profezzorn, line 162: We shouldn't have a hard transition between "looping" and "end".
Instead, we should keep running the loop until "end" is done, and layer the end transition on top of the loop transition.

transitions/loop.h
```diff
@@ -103,4 +103,69 @@ class TrLoopUntil {
...
+// Continuously runs LOOP_TRANSITION while CONDITION > 0
+// Makes more sense if TRANSITION is a TrConcat, as this will
+// transition to/from the intermediate steps in a loop.
+// If CONDITION <= 0, runs END_TRANSITION and stops.
+
+template<class LoopTr, class EndTr, class CONDITION>
+class TransitionLoopWhileL {
+
+public:
+void run(BladeBase* blade) {
```
profezzorn, line 120: this whole function should be indented two steps

transitions/loop.h
```diff
@@ -103,4 +103,69 @@ class TrLoopUntil {
...
+// return value: COLOR
+//
+// Continuously runs LOOP_TRANSITION while CONDITION > 0
+// Makes more sense if TRANSITION is a TrConcat, as this will
+// transition to/from the intermediate steps in a loop.
+// If CONDITION <= 0, runs END_TRANSITION and stops.
+
+template<class LoopTr, class EndTr, class CONDITION>
+class TransitionLoopWhileL {
+
```
profezzorn, line 118: remove empty line

fdarosa2663:
Had in the wrong .h file, this belongs in transition_loop.h, not loop.h

transitions/loop.h
fdarosa2663: fixed

transitions/loop.h
fdarosa2663: fixed

transitions/loop.h
profezzorn: Basically like:

decltype(SAME AS RETURN TYPE) ret = RGBA_um_nod::Transparent();
if (run_) {
  ret = loop_tr_.getColor(RGBA_um_nod::Transparent(), RGBA_um_nod::Transparent(), led);
  if (end_)  ret = loop_tr_.getColor(ret, RGBA_um_nod::Transparent(), led);
}
return ret;


The MixColors inside decltype() is just calculating the return type of the function, it doesn't do anything else.

fdarosa2663: Thanks, I "think" I did it right BUT the end transition does seem to actually run, it just ends when the condition is false.  I tried moving the if (end_) outside the if(run_) but it did the same thing, so it must be something else?  Any ideas?

fdarosa2663:
I rearranged things and fixed how end_ worked and it's running as expected now.

styles/transition_loop.h
```diff
@@ -34,4 +34,66 @@ class TransitionLoopL {
...
+// Continuously runs LOOP_TRANSITION while CONDITION > 0
+// Makes more sense if TRANSITION is a TrConcat, as this will
+// transition to/from the intermediate steps in a loop.
+// If CONDITION <= 0, runs END_TRANSITION and stops.
+
+template<class LoopTr, class EndTr, class CONDITION>
+class TransitionLoopWhileL {
+public:
+  void run(BladeBase* blade) {
+    condition_.run(blade);
```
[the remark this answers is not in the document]
fdarosa2663: Updated.

styles/transition_loop.h
```diff
@@ -34,4 +34,61 @@ class TransitionLoopL {
...
+  EndTr end_tr_;
+  PONUA SVFWrapper<CONDITION> condition_;
+
+public:
+auto getColor(int led) -> decltype(
+    MixColors(end_tr_.getColor(loop_tr_.getColor(RGBA_um_nod::Transparent(), RGBA_um_nod::Transparent(), led), RGBA_um_nod::Transparent(), led),
+        loop_tr_.getColor(loop_tr_.getColor(RGBA_um_nod::Transparent(), RGBA_um_nod::Transparent(), led), RGBA_um_nod::Transparent(), led), 1, 1)) {
+    decltype(MixColors(end_tr_.getColor(loop_tr_.getColor(RGBA_um_nod::Transparent(), RGBA_um_nod::Transparent(), led), RGBA_um_nod::Transparent(), led),
+        loop_tr_.getColor(loop_tr_.getColor(RGBA_um_nod::Transparent(), RGBA_um_nod::Transparent(), led), RGBA_um_nod::Transparent(), led), 1, 1)) ret = RGBA_um_nod::Transparent();
+    if (end_) ret = end_tr_.getColor(loop_tr_.getColor(RGBA_um_nod::Transparent(), RGBA_um_nod::Transparent(), led), RGBA_um_nod::Transparent(), led);
```
profezzorn, line 88: line 88 and 89 needs to switch places.
The end transition needs to run "on top of" the loop, which means it needs to be after in the code.
Also, the first input to the end end transition should be "ret", not transparent.

the return type needs to be fixed in a similar manner too..

fdarosa2663: "Also, the first input to the end end transition should be "ret", not transparent."

I didn't quite get this, how should it be?

profezzorn: I just realized one more thing, if run_ and loop_ are both true, then this will run loop_tr_ twice, I think it should be like this:

```
if (run_) {
  ret = loop_tr_.getColor(RGBA_um_nod::Transparent(), RGBA_um_nod::Transparent(), led); 
  if (end_) ret = end_tr_.getColor(ret, RGBA_um_nod::Transparent(), led);
}
```

PR #783, opened 2025-03-26
Add FETT263_REPLACE_CC_COLOR_LIST

fdarosa2663:
Allow users to use Edit Hue in place of Color List for COLOR_ARG styles in ColorChange Mode.


props/saber_fett263_buttons.h
```diff
@@ -5924,7 +5987,13 @@ SaberFett263Buttons() : PropBase() {}
             return true;
           }
 #endif
-          if (color_mode_ == CC_COLOR_LIST) {
+          if (
+#ifdef FETT263_REPLACE_CC_COLOR_LIST
```
profezzorn, line 5991: There seems to be a lot of these.
Why not just do something like somewhere higher up:

```
#ifdef FETT263_REPLACE_CC_COLOR_LIST
#define ACTIVE_COLOR_MODE CC_EDIT_COLOR
#else
#define ACTIVE_COLOR_MODE CC_COLOR_LIST
#endif
```

Then here you can just do color_mode_ == ACTIVE_COLOR_MODE

Or something like that?

fdarosa2663: I replaced where it made sense, in other locations the define is changing which functions are used and/or how the mode works as the Color List and Edit Color versions function differently.  They're not interchangeable.

props/saber_fett263_buttons.h
```diff
@@ -2073,6 +2099,20 @@ SaberFett263Buttons() : PropBase() {}
       handles_color_change |= current_config->blade##N->current_style() && current_config->blade##N->current_style()->IsHandled(HANDLED_FEATURE_CHANGE_TICKED);
       ONCEPERBLADE(USES_COLOR_CHANGE)
       if (!handles_color_change) {
+#ifdef FETT263_REPLACE_CC_COLOR_LIST
+        color_mode_ = CC_EDIT_COLOR;
+        show_color_all_.Start();
+        for (int i = 1; i <= NUM_BLADES; i++) {
+          if (style_parser.UsesArgument(current_preset_.GetStyle(i), BASE_COLOR_ARG + 2)) {
+            ShowColorStyle::SetColor(GetColorArg(i, BASE_COLOR_ARG));
+            saved_color_ = GetColorArg(i, BASE_COLOR_ARG);
```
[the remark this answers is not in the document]
fdarosa2663: saved_color_ gets used in Edit Mode across Color Editing functions to be able to revert to the original color.  This mode is just going in and out of one mode, but in Edit Mode it will go across several to change Hue, White Level and Black Level.

PR #784, opened 2025-03-27
Add hev.h

jimsworld:
Upload of hev.h. This branch has taken into account the amendments from the old branch.


jimsworld:
Delaying Hazard damage allows time for the Hazard voice line to play out.
This is the sequence:

- random Hazard called.
- delay Hazard damage for 6s.
- start Hazard damage.
- Hazard damage happens every 1s until stopped.

The current timings are just for testing.

props/hev.h
```diff
@@ -0,0 +1,440 @@
...
+
+  // Random Hazards ////////////////////////////////////////////////////////////
+  uint32_t last_random_draw_ = millis();
+  uint32_t post_death_cooldown_ = 0;
+
+  // Check and activation
+  void CheckRandomEvent() {
+    // Don't allow Hazards if dead or during post-death cooldown.
+    // This prevents Hazards immediately after healing.
+    if (health_ == 0 || (post_death_cooldown_ &&
```
profezzorn, line 170: millis will eventually return zero, it's better to have a separate bool for if the cooldown is "active" instead of re-using the timeout value.

props/hev.h
```diff
@@ -0,0 +1,440 @@
...
+  // Armor Readout /////////////////////////////////////////////////////////////
+  void armor_readout() {
+    PVLOG_NORMAL << "Current Armor: " << armor_ << "\n"; // Debug logging
+    SFX_armor.SelectFloat(armor_ / 100.0);
+    hybrid_font.PlayCommon(&SFX_armor);
+  }
+
+
+  // Clashes ///////////////////////////////////////////////////////////////////
+  uint32_t last_clash_time_ = 0;
```
profezzorn, line 112: Since you're doing a lot with timers, it might be helpful to have a helper class, something like:

```
struct HEVTimer {
   bool reset() { active_ = false; }
   void start() { active_ = true; start_ = millis(); }
   bool check(uint32_t timeout) {
     if (!active_) return false;
     if (millis() - start_ > timeout) return true;
     active_ = false;
     return false;
   }
   uint32_t start_;
   bool active_ = false;
};
```

props/hev.h
```diff
@@ -0,0 +1,440 @@
...
+	if (SaberBase::Lockup()) {
+	  SaberBase::DoEndLockup();
+	  SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+	  return true;
+	}
+	break;
+
+  // Hold POWER to start recharging armor
+      case EVENTID(BUTTON_POWER, EVENT_HELD_MEDIUM, MODE_ON):
+    if (!SaberBase::Lockup()) {
```
profezzorn, line 363: fix indentation

props/hev.h
```diff
@@ -0,0 +1,440 @@
...
+      case EVENTID(BUTTON_AUX, EVENT_RELEASED, MODE_ANY_BUTTON | MODE_ON):
+      case EVENTID(BUTTON_AUX, EVENT_RELEASED, MODE_ANY_BUTTON | MODE_OFF):
+	if (SaberBase::Lockup()) {
+	  SaberBase::DoEndLockup();
+	  SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+	  return true;
+	}
+	break;
+
+  // Hold POWER to start recharging armor
```
profezzorn, line 361: fix indentation

props/hev.h
```diff
@@ -0,0 +1,440 @@
...
+      case EVENTID(BUTTON_POWER, EVENT_HELD_MEDIUM, MODE_ON):
+    if (!SaberBase::Lockup()) {
+      SaberBase::SetLockup(SaberBase::LOCKUP_LIGHTNING_BLOCK);
+      SaberBase::DoBeginLockup();
+      return true;
+    }
+    break;
+    // Release POWER to stop recharging armor (or wait until full).
+        case EVENTID(BUTTON_POWER, EVENT_RELEASED, MODE_ANY_BUTTON | MODE_ON):
```
profezzorn, line 370: fix indentation

props/hev.h
```diff
@@ -0,0 +1,453 @@
...
+  void armor_readout() {
+    PVLOG_NORMAL << "Current Armor: " << armor_ << "\n"; // Debug logging
+    SFX_armor.SelectFloat(armor_ / 100.0);
+    hybrid_font.PlayCommon(&SFX_armor);
+  }
+
+  // Clashes
+  void Clash(bool stab, float strength) override {
+    // Don't allow Clashes if dead or during debounce period.
+    if (health_ == 0 || !clash_timer_.check(HEV_CLASH_DEBOUNCE_MS)) {
```
profezzorn, line 154: prop_base already has debounce for clashes, why not just use that?

jimsworld: I just didn't really know how to implement it tbh! I'll take a look into it.

jimsworld:
millis() are now all handled in the HEVTimer class.

props/hev.h
```diff
@@ -0,0 +1,463 @@
...
+// death.wav - for when health is 0
+// clash**.wav - physical clash sounds
+// health00.wav to health100.wav - for health alert sounds
+// armor00.wav to armor100.wav - for armor readout sounds
+//
+// Random Hazard sounds:
+// These are played using the altchng method. So create 7 directories:
+// alt00, alt01, alt02, alt03, alt04, alt05, alt06.
+// Each directory should contain two folders, altchng and stun.
+// In alt01 to alt06, place the hazard detection voice lines (bio, blood toxins, chemical, radiation, shock, fire) in altchng.
```
profezzorn, line 32: line is too long

props/hev.h
```diff
@@ -0,0 +1,463 @@
...
+
+class Hev : public PROP_INHERIT_PREFIX PropBase {
+public:
+  Hev() : PropBase() {}
+  const char* name() override { return "Hev"; }
+
+  int health_ = 100;
+  int armor_ = 100;
+
+  HEVTimer clash_timer_;
```
profezzorn, line 97: Am I right to assume that only some of these timers needs the "next_tick_" ?
If so, we might want to create a subclass with the additional functionality and only use that class where it is needed for clarity.

props/hev.h
```diff
@@ -0,0 +1,463 @@
...
+  void DoDamage(int damage, bool quiet = false) {
+    int tens = health_ / 10;
+    if (armor_ >= damage) {
+      armor_ -= ceilf(damage * 0.80 / 2);
+      health_ -= ceilf(damage * 0.20);
+    } else if (armor_ > 0) {
+      int excess_physical = damage - armor_;
+      health_ -= excess_physical;
+      // Make armor compromised sound play before setting armor to 0
+      SaberBase::DoEffect(EFFECT_USER2, 0.0, 0);
```
profezzorn, line 127: This explicitly tries to play sound 0 rather than playing a random sound.
If that's not what you want, just remove the last zero.

props/hev.h
```diff
@@ -0,0 +1,463 @@
...
+    // Enforce minimum values
+    if (health_ < 0) health_ = 0;
+    if (armor_ < 0) armor_ = 0;
+
+    // Damage
+    if (!quiet) SaberBase::DoEffect(EFFECT_STUN, 0.0, damage);
+    
+    // Death Sound
+    if (health_ == 0) {
+      SaberBase::DoEffect(EFFECT_EMPTY, 0.0, 100);
```
profezzorn, line 143: Playing sound 100?
Why?

props/hev.h
```diff
@@ -0,0 +1,463 @@
...
+    } else {
+      health_ -= damage;
+    }
+
+    // Enforce minimum values
+    if (health_ < 0) health_ = 0;
+    if (armor_ < 0) armor_ = 0;
+
+    // Damage
+    if (!quiet) SaberBase::DoEffect(EFFECT_STUN, 0.0, damage);
```
profezzorn, line 139: It might be better to do SFX_stun.SetFloat(max<float>(1.0, damage/100.0)) instead of using the three-argument version of DoEffect().

SetFloat scales the number to the number of available sounds, while the three-argument version just picks the sound number directly.

profezzorn (commented):
... still didn't get through the whole file...

props/hev.h
```diff
@@ -0,0 +1,463 @@
...
+EFFECT(death);
+EFFECT(armor_alarm);
+EFFECT(armor_compromised);
+EFFECT(major);
+EFFECT(minor);
+EFFECT(morphine);
+
+struct HEVTimer {
+  uint32_t start_ = 0;
+  uint32_t next_tick_ = 0;
```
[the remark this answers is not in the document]
jimsworld: I believe I was having issues with damage per second being consistent, where hazard damage was being applied at the rate of the value of the _initial delay_ we made a while ago, HEV_HAZARD_DELAY_MS, which allowed the hazard alert voice line to play out before actual hazard sound effects interrupted.

next_tick_ was a way to ensure a guaranteed schedule for damage. Currently it is only used for the hazard damage sequence in combination with HEV_HAZARD_DECREASE_MS. Ultimately it's there to prevent any mishaps or timing drift.

Later on, I'm interested on expanding damage per second to be more... random, where damage could drop at 1hp, 2hp or 3hp every 1s or 2s... or 4s, instead of just the fixed 1s it is now.

I will add a comment to state what this is.

props/hev.h
jimsworld: Ammended

props/hev.h
```diff
@@ -0,0 +1,463 @@
...
+  void reset() { active_ = false; next_tick_ = 0; }
+  
+  void start() { active_ = true; start_ = next_tick_ = millis(); }
+
+  // Returns true if timer is inactive or timeout exceeded
+  bool check(uint32_t timeout) {
+    return !active_ || (millis() - start_ > timeout);
+  }
+
+  // Manages hazard warning and damage sequence timing
```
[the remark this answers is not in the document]
jimsworld: This goes hand-in-hand with the next_tick_ comment above. it's all a part of the idea to delay hazard damage so that the hazard alert voice lines can playout without the burning damage overpowering the suit's voice, or the other stun sounds.

**warning_period** is basically a grace period. I suppose that could be named better. This is is the _initial delay_ which is basically the HEV_HAZARD_DELAY_MS.
**tick_interval** used HEV_HAZARD_DECREASE_MS as an argument, which is set to 1000ms, at the moment.

The hazard sequence method returns **true** when:

- The timer is active (active_ is true)
- The warning period has elapsed (now - start_ >= warning_period)
- It's time for the next damage tick (now >= next_tick_)

It returns **false** when:

- The timer is not active (active_ is false)
- The warning period hasn't elapsed yet
- It's not time for the next damage tick

To be honest I still think hazard damage as a whole (the function and the timing) is still in it's early stages at the moment, I mean conceptionally. I've sort of added to this to help draft together the basis of the hazard damage function. So it's possible a re-write could happen for this but I just decided to focus on putting together other functions.

I think moving forwards this idea could be developed a bit better. For example, the voice lines would all be handled by a future HEVQueue class which handles the ordering and queuing of the voice lines. HEVTimer will still be responsible for various timings on different things.

Having control over unwanted sound overlaps is the goal here, but perhaps what if we lowered the sound of the stun wavs **WHILE** HEVQueue is busy playing a voice line?

props/hev.h
```diff
@@ -0,0 +1,463 @@
...
+  bool hazard_sequence(uint32_t warning_period, uint32_t tick_interval) {
+    if (!active_) return false;
+    uint32_t now = millis();
+    
+    return (now - start_ >= warning_period) && 
+           ((now >= next_tick_) ? (next_tick_ = now + tick_interval, true) : false);
+  }
+
+  // Returns true if timer is active and interval has elapsed
+  bool ready(uint32_t interval) {
```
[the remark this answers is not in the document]
jimsworld: Okay, I have made changes to the HEVTimer and subsequently any method that calls the timer, for the next commit. Changes we mainly for better categorization and cleanliness.

props/hev.h
jimsworld: I went ahead and created the subclass and renamed the timers for clarity. Hopefully this helps with readability by seperating some of the basic methods into HEVTimerBase and another class, HEVTimerHazard, for specific hazard damage timing.

Currently, if all looks okay to you, I think this works for now, but could be expanded on in the future.

props/hev.h
jimsworld: Thanks for the tip! There is only 1 armor compromised voice line. But I shall remove it anyway as I could have a bit of fun with the font in the future.

props/hev.h
jimsworld: Uhhhhh... hmmmm... no idea. I'll remove the 💯

props/hev.h
jimsworld: So I added:

`    if (!quiet) SFX_stun.SetFloat(max(1.0, damage/100.0));`

But I got a compile error. SelectFloat seemed to compile correctly, but I'm not sure what the difference is? I will leave it as:

`    if (!quiet) SaberBase::DoEffect(EFFECT_STUN, 0.0);`

for the moment. Thankfully, by going through removing the third argument in the other DoEffect's, one of them fixed the persistent stun02.wav playing per every hazard damage 👍

props/hev.h
```diff
@@ -0,0 +1,534 @@
...
+  // Armor Readout
+  void armor_readout() {
+    PVLOG_NORMAL << "Current Armor: " << armor_ << "\n"; // Debug logging
+    SFX_armor.SelectFloat(armor_ / 100.0);
+    hybrid_font.PlayCommon(&SFX_armor);
+  }
+
+  // Clashes
+  void Clash(bool stab, float strength) override {
+    // Don't process clashes if dead or during cooldown.
```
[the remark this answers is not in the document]
jimsworld: Okay cool, I've removed most of the comment bloat and the subsequent comments throughout the rest of the Clash function. Going to commit now.

props/hev.h
```diff
@@ -0,0 +1,512 @@
...
+        return true;
+      case EVENTID(BUTTON_AUX, EVENT_THIRD_SAVED_CLICK_SHORT, MODE_ANY_BUTTON | MODE_ON):
+      case EVENTID(BUTTON_AUX, EVENT_THIRD_SAVED_CLICK_SHORT, MODE_ANY_BUTTON | MODE_OFF):
+        previous_preset();
+        return true;
+
+      // Hold AUX to start healing
+      case EVENTID(BUTTON_AUX, EVENT_HELD_MEDIUM, MODE_ON):
+        if (!SaberBase::Lockup()) {
+          SaberBase::SetLockup(SaberBase::LOCKUP_NORMAL);
```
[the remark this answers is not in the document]
jimsworld: I've added both of these at the top.

Should I go through the entire file and replace LOCKUP_NORMAL and LOCKUP_LIGHTNING_BLOCK with LOCKUP_HEALING and LOCKUP_FILL_ARMOR, respectively?

props/hev.h
```diff
@@ -0,0 +1,512 @@
...
+  void configure(uint32_t interval) {
+    interval_ = interval;
+  }
+
+  // Returns true if timer is inactive or timeout exceeded.
+  bool check() {
+    return !active_ || (millis() - start_ > interval_);
+  }
+
+  // Returns true if timer is active and interval has elapsed.
```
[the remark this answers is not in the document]
jimsworld: No problem. Removed.

props/hev.h
```diff
@@ -0,0 +1,512 @@
...
+    armor_++;
+    timer_armor_increase_.start();
+    PVLOG_NORMAL << "Armor: " << armor_ << "\n";
+  }
+
+  // Main Loop
+  void Loop() override {
+    CheckRandomEvent();
+    HazardDecrease();
+    if (SaberBase::Lockup() == SaberBase::LOCKUP_NORMAL) {
```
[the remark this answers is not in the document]
jimsworld: Okay, I've added this line at the top of the IncreaseHealth function, same for IncreaseArmor too but using LOCKUP_FILL_ARMOR instead.

I've also called both functions in the main loop. Tested and everything seems to work as normal.

props/hev.h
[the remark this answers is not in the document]
profezzorn: Yes, please do the replace.

props/hev.h
```diff
@@ -0,0 +1,513 @@
...
+    }
+
+    // Start sequence if not running
+    if (!timer_hazard_delay_.active_) {
+      timer_hazard_delay_.start();
+      return;
+    }
+
+    // Check sequence and apply damage
+    if (timer_hazard_delay_.hazard_sequence()) {
```
profezzorn, line 300: Here is what I think we should do:

1. Replace timer_hazard_delay_ with a regular timer.
2. This lines becomes 'if (timer_hazard_delay_.done())`
3. when we start the timer the first time (line 295) we'll also call timer_hazard_delay_.configure(HEV_HAZARD_DELAY_MS)
4. Inside this if statement, we add another call to configure with (HEV_HAZARD_DECREASE_MS) and a call to timer_hazard_delay_.start()
5. remove TimerHazard
6.

props/hev.h
```diff
@@ -0,0 +1,513 @@
...
+  HEVTimerBase timer_hazard_damage_;
+  HEVTimerBase timer_health_increase_;
+  HEVTimerBase timer_armor_increase_;
+  HEVTimerHazard timer_hazard_delay_;
+
+  Hev() : PropBase() {
+    // Configure all timers with their respective timings from hev_config.h
+    // Clash timer uses clash_timeout_ from PropBase for debounce
+    timer_clash_.configure(this->clash_timeout_);
+    timer_post_death_cooldown_.configure(HEV_POST_DEATH_COOLDOWN_MS);
```
profezzorn, line 156: I think all of these defines should be described in the top comment.
Also, each define should probably have a default value, which means having something like this somewhere:

```cpp
#ifndef HEV_POST_DEATH_COOLDOWN_MS
#define HEV_POST_DEATH_COOLDOWN_MS  3000 
#endif
```

props/hev.h
```diff
@@ -0,0 +1,512 @@
...
+      SaberBase::SetLockup(SaberBase::LOCKUP_NONE);
+      return;
+    }
+
+    if (!SaberBase::Lockup() || !timer_health_increase_.ready()) {
+      return;
+    }
+
+    if (health_ == 0) {
+      timer_post_death_cooldown_.start();
```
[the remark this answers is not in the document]
jimsworld: Okay, I've added a new function called Revive() and added timer_random_event.reset().

I've removed this from the IncreaseHealth() function:

```
    if (health_ == 0) {
      timer_post_death_cooldown_.start();
```

props/hev.h
jimsworld: Okay looks like we'll remove the TimerHazard, no problem!

Could I grab a hand with some of the code implementation please? I tried my hand at it and have ended up making the hazard damage almost instantly kill the user... lol

props/hev.h
jimsworld: I shall give the top comment a revamp tomorrow, including the sound files section similar to how I explained the other day.

with the defines themselves, they're actually in hev_config.h.... should we move _all_ specific hev defines into hev.h?

jimsworld: Top comment has been overauled.

jimsworld:
Caught up with last batch of reviews.

props/hev.h
profezzorn: The top comment has a lot now, but it doesn't describe this define or how to use it.

props/hev.h
```diff
@@ -0,0 +1,667 @@
...
+// ▪ All alt**/ folders (alt00–alt06) must have the **same number** of //
+//    WAVs inside including **all** sub folders.                       //
+//                                                                     //
+//=====================================================================//
+
+//=====================================================================//
+//                       TIMER CONFIGURATION                           //
+//                    ─────────────────────────                        //
+//                                                                     //
+// - timer_clash_               - Debounce timer for physical Clashes. //
```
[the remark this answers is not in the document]
jimsworld: No problem, I've moved the comment block above those timers.

props/hev.h
```diff
@@ -0,0 +1,667 @@
...
+//                                                                     //
+//------------------------- Health Alerts -----------------------------//
+//                                                                     //
+// ▪ When Health reaches the below thresholds, the suit will           //
+//   randomly say different (HEV VOICE LINES):                         //
+//     - ≤= 50: "Seek Medical Attention"        (50% chance)           //
+//     - ≤= 30: "Vital Signs Critical"          (60% chance)           //
+//     - ≤= 6:  "User Death Imminent"           (100% chance)          //
+//                                                                     //
+// ▪ The chances have been configured directly in the sound files,     //
```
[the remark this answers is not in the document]
jimsworld: Same as above. I''ve made a table displaying exactly what files have audio and which do not. People would have to customize this themselves if they want higher or lower chance of hearing these particular voice lines.

jimsworld:
Renamed `HEV_POST_DEATH_COOLDOWN_MS` to `HEV_HAZARD_AFTER_REVIVE_MS`

In retrospect, it was a brilliant source of confusion. Hoping the new name is better.

props/hev.h
```diff
@@ -0,0 +1,513 @@
...
+    
+    // Death Sound
+      case EFFECT_EMPTY:
+        if (health_ == 0) {
+          SOUNDQ->clear_pending();
+        }
+        hybrid_font.PlayCommon(&SFX_death);
+        return;
+    }
+  }
```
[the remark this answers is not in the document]
jimsworld: There's still a few more voice lines to be queued but not yet implemented.

In the current implementation, there's only two other voice lines that could be queued, boot.wav and armor**.wav.

I'm don't think it's necessary to queue boot.wav? But I think armor** should be. However, it does play an _effect_ (HEV UI Sound) just before armor** plays. I have some code ready for it as it.

Being that fuzz** is technically a HEV UI Sound, I think it should be queued before armor**.

Here's the small addition to armor_readout(). Does this look okay?
I'm not sure if that delay would be okay...
I'm also not sure exactly where it should be situated, like down with the rest of the actual queue in SB_Effect ?

```
  void armor_readout() {
    PVLOG_NORMAL << "Current Armor: " << armor_ << "\n";
    
    // Play random "fuzz" sound only if armor is above 0
    if (armor_ > 0) {
      SOUNDQ->Play(&SFX_fuzz);

      // Delay to ensure fuzz finishes before armor readout plays
      HEVTimer().delay(475);
    }

    // Play armor value
    SFX_armor.SelectFloat(armor_ / 100.0);
    SOUNDQ->Play(&SFX_armor);
  }
```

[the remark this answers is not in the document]
profezzorn: First of all, I think I made a small mistake previously. The "EFFECT_ALT_SOUND" is supposed to be in SB_Effect2 to work reliably. That's why I wrote out the whole function. (But then I called it SB_Effect, instead of SB_Effect2..)

The armor readout code is weird though.

This part:
```
      HEVTimer().delay(475);
```

is complete bogus, because HEVTimer() doesn't have a delay function, and if it did, that would not be how you call it. Also, we don't need to delay anything if we're just adding things to a queue. Once one sound is done playing, it will play the next sound in the queue. If we want to add silence to the queue, we can do that by adding quiet to the end of a wav file, or we can have a 475ms silent sound file to play in between the sounds.

We could also just add the fuzz effect to the armour readout sounds directly, which might be the most flexible solution for anybody who wants to play something else.

props/hev.h
```diff
@@ -0,0 +1,788 @@
...
+//        Default from 0 - 100 Health is 10 seconds.                   //
+//        Lower = faster healing.                                      //
+//                                                                     //
+//  HEV_ARMOR_INCREASE_MS                                              //
+//      - Time (ms) between each armor recharge tick (hold POWER).     //
+//        Default from 0 - 100 Armor is 10 seconds.                    //
+//        Lower = faster recharge.                                     //
+//=====================================================================//
+
+#define HEV_RANDOM_EVENT_INTERVAL_MS 60000
```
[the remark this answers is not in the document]
jimsworld: Okay will do.

props/hev.h
```diff
@@ -0,0 +1,795 @@
...
+
+    // Enforce minimum values
+    if (health_ < 0) health_ = 0;
+    if (armor_ < 0) armor_ = 0;
+
+    // Damage
+    if (!quiet) SaberBase::DoEffect(EFFECT_STUN, 0.0);
+    
+    // Death Sound
+    if (health_ == 0) {
```
profezzorn, line 456: Quirk #1

We can fix this by only playing the death sound when health_ went from a non-zero value, to a zero value.
Basically, at the beginning of this function, we ave the health in a local variable, something like:

```cpp
   int previous_health = health;
```

Then here, we replace this if statement with:

```cpp
    if (health_ == 0 && prevous_health > 0) {
```

That way, if we were already out of health when DoDamage is called, the death sound will not play again.

props/hev.h
```diff
@@ -0,0 +1,811 @@
...
+          int excess_physical = damage - armor_;
+          health_ -= excess_physical;
+          armor_ = 0;
+        } else { health_ -= damage; }
+        break;
+
+      case DAMAGE_HAZARD:
+        if (armor_ > 0) {
+          armor_ -= damage;
+        } else { health_ -= damage; }
```
profezzorn, line 458: health_ -= damage; shouldbe on a line by itself.

If statements should always follow one of these patterns:
```
if (CONDITION) {
  CODE;
}
```
or, if there is an else:
```
if (CONDITION) {
  CODE
} else {
  OTHER CODE
}
```

if everything fits on one line, you can use:
```
if (CONDITION) CODE;
```

if there is no else, and the code is short enough, you can do:
```
if (CONDITION)
  CODE;
```

This is also allowed for repeated indentations, like, but there must be no brackets and each line must be relatively short:
```
if (CONDITION)
  for (A;B;C)
    if (CONDITION)
     CODE;
```

Most of the time, it's best to just stick with the first one.

props/hev.h
```diff
@@ -0,0 +1,811 @@
...
+    switch (type) {
+      case DAMAGE_PHYSICAL:
+        if (armor_ >= damage) {
+          armor_ -= ceilf(damage * 0.80 / 2);
+          health_ -= ceilf(damage * 0.20);
+        } else if (armor_ > 0) {
+          int excess_physical = damage - armor_;
+          health_ -= excess_physical;
+          armor_ = 0;
+        } else { health_ -= damage; }
```
profezzorn, line 452: Move health_ -= damage; to it's own line (see comment below)

props/hev.h
```diff
@@ -0,0 +1,811 @@
...
+        break;
+    }
+
+    // Enforce minimum values
+    if (health_ < 0) health_ = 0;
+    if (armor_ < 0) armor_ = 0;
+
+    // (HEV VOICE LINE) Logic for Armor Compromised
+    if (previous_armor > 0 && armor_ == 0) {
```
profezzorn, line 467: should we skip this if the health_ == 0 ?

profezzorn (commented):
almost there...

props/hev.h
jimsworld: Thank you for the tips Prof, I've made the changes.

PR #785, opened 2025-03-27
Add sound queue.h addition

jimsworld:
A small addition in sound_queue.h to be able to purge/clear **any pending** sounds from the queue.


sound/sound_queue.h
```diff
@@ -86,6 +86,11 @@ class SoundQueue {
     fadeout_ = true;
     fadeout_len_ = len;
   }
+  void clear_pending() {
```
profezzorn, line 89: This doesn't stop the current sound, just stops starting new ones, which could be really weird.
Is that what you want?

jimsworld: Yep, that's the intended behaviour. When health is 0 and the flatline sound is triggered, it would sound odd if the suit kept playing queued up sounds. It should only finish the currently playing sound once health is 0.

sound/sound_queue.h
```diff
@@ -86,6 +86,11 @@ class SoundQueue {
...
+    if (busy_) {
```
profezzorn, line 90: Why is there an "if (busy_) {" here?

jimsworld: I think it was there as a safety mechanism in testing. I'll remove it then test 👍

PR #786, opened 2025-03-27
Add hev_config.h

jimsworld:
Uploading WIP hev config file.

### DIFF
--- a/config/hev_config.h
+++ b/config/hev_config.h
@@ -1,0 +1,43 @@
+#ifdef CONFIG_TOP
+#include "proffieboard_v2_config.h"
+#define NUM_BLADES 1
+#define NUM_BUTTONS 2
+#define VOLUME 30
+const unsigned int maxLedsPerStrip = 144;
+#define CLASH_THRESHOLD_G 3.0
+#define ENABLE_AUDIO
+#define ENABLE_MOTION
+#define ENABLE_WS2811
+#define ENABLE_SD
+#define DISABLE_NO_REPEAT_RANDOM
+#define HEV_POST_DEATH_COOLDOWN_MS 30000
+#define HEV_RANDOM_EVENT_INTERVAL_MS 30000
+#define HEV_RANDOM_HAZARD_CHANCE 50
+#define HEV_HAZARD_DELAY_MS 6000
+#define HEV_HAZARD_DECREASE_MS 1000
+#define HEV_HEALTH_INCREASE_MS 100
+#define HEV_ARMOR_INCREASE_MS 100
+#endif
+
+#ifdef CONFIG_PROP
+#include "../props/hev.h"
+#endif
+
+#ifdef CONFIG_PRESETS
+Preset presets[] = {
+  { "_hev;common_hev", "common_hev/tracks/hl1_ost/10 Valve Theme [Extended].wav",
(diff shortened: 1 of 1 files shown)


config/hev_config.h
```diff
@@ -0,0 +1,43 @@
+#ifdef CONFIG_TOP
+#include "proffieboard_v2_config.h"
+#define NUM_BLADES 1
+#define NUM_BUTTONS 2
+#define VOLUME 30
+const unsigned int maxLedsPerStrip = 144;
+#define CLASH_THRESHOLD_G 3.0
+#define ENABLE_AUDIO
```
profezzorn, line 8: These four ENABLE_.. are not needed in ProffieOS 8.x

config/hev_config.h
```diff
@@ -0,0 +1,43 @@
...
+#define ENABLE_MOTION
+#define ENABLE_WS2811
+#define ENABLE_SD
+#define DISABLE_NO_REPEAT_RANDOM
```
profezzorn, line 12: Maybe add a comment as to why you think it is better to disable no repeat random?

config/hev_config.h
```diff
@@ -0,0 +1,43 @@
...
+#define HEV_RANDOM_EVENT_INTERVAL_MS 30000
```
profezzorn, line 13: These ones are all the same as in the prop file, right?
(If so, we don't really need them here.)

config/hev_config.h
```diff
@@ -0,0 +1,43 @@
...
+#define HEV_ARMOR_INCREASE_MS 100
+#endif
+
+#ifdef CONFIG_PROP
+#include "../props/hev.h"
+#endif
+
+#ifdef CONFIG_PRESETS
+Preset presets[] = {
+  { "_hev;common_hev", "common_hev/tracks/hl1_ost/10 Valve Theme [Extended].wav",
```
profezzorn, line 28: why start with an _ ?
Also, for someone who wants to build their own HEV, where can they find these sounds?

config/hev_config.h
jimsworld: No problem, I will remove them.

config/hev_config.h
jimsworld: Yes indeed they are. The ones in the prop file are the "default" values. These ones were meant to be configurable but I suppose there's only really one hev prop file unlike a multitude of saber files. I'll remove them.

config/hev_config.h
profezzorn: Comment is good.

config/hev_config.h
```diff
@@ -0,0 +1,37 @@
...
+#define CLASH_THRESHOLD_G 3.0
+//If not disabled, Armor Readout play back floats around current Armor value.
+#define DISABLE_NO_REPEAT_RANDOM
+#endif
+
+#ifdef CONFIG_PROP
+#include "../props/hev.h"
+#endif
+
+// All HALF-LIFE sfx audio can be found in the game's "sound" directory.
```
profezzorn, line 16: This comment is really helpful, but I would imagine that you still have to convert sounds to 44.1khz PCM mono and edit and rename some sounds.  It would be great to have a guide that explains what needs to be done to make a functional font from the game directories.  Although, that information should probably be in the prop file, not here. (And it could be in a different PR.)

PR #787, opened 2025-04-02
Add Blade Detect functionality to manual Sabersense Array Selector...

Sabersense-CC:
Adds Blade Detect functionality when using SABERSENSE_ARRAY_SELECTOR. When using Blade Detect, the manual array selector skips the NO_BLADE array. 
Also allows the user to set a default blade array via a define. This can be useful if using arrays to set different blade lengths, as you can have a logical progression of arrays from long to short for example, then set a default for the most often used blade or the blade supplied with the hilt.

### DIFF
--- a/props/saber_sabersense_buttons.h
+++ b/props/saber_sabersense_buttons.h
@@ -297,7 +297,7 @@ GESTURE CONTROLS
   constexpr size_t NUM_ARRAYS = sizeof(blades) / sizeof(blades[0]);
   static_assert(SABERSENSE_DEFAULT_ARRAY >= 1 && SABERSENSE_DEFAULT_ARRAY <= NUM_ARRAYS,
 
-    // Show error below if array number is invalid. (Returns and indents for readability).
+    // Show error below if array number is invalid. (Returns and indents for IDE readability).
     "\n  ERROR:\n    SABERSENSE_DEFAULT_ARRAY number invalid. Ensure array number specified "
     "does not exceed number of blade arrays in config.\n "
     "   Arrays should be numbered sequentially. First array should be zero (or NO_BLADE if "
--- a/props/saber_sabersense_buttons.h
+++ b/props/saber_sabersense_buttons.h
@@ -1,4 +1,4 @@
-/* V7/8-167.
+/* V7/8-177.
 ============================================================
 =================   SABERSENSE PROP FILE   =================
 =================            by            =================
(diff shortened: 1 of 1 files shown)


props/saber_sabersense_buttons.h
```diff
@@ -275,27 +288,53 @@ GESTURE CONTROLS
...
 
 #ifdef SABERSENSE_ARRAY_SELECTOR
+
+#ifndef SABERSENSE_DEFAULT_ARRAY
+#define SABERSENSE_DEFAULT_ARRAY 1
+#endif
+
+  // Check that user-specified SABERSENSE_DEFAULT_ARRAY is within valid range.
+  constexpr size_t NUM_ARRAYS = sizeof(blades) / sizeof(blades[0]);
+  static_assert(SABERSENSE_DEFAULT_ARRAY >= 1 && SABERSENSE_DEFAULT_ARRAY <= NUM_ARRAYS,
```
profezzorn, line 298: Why can't it be zero?

props/saber_sabersense_buttons.h
```diff
@@ -275,27 +288,53 @@ GESTURE CONTROLS
...
+    "using Blade Detect) then 1, 2, 3, etc.\n");
+
 #ifndef SABERSENSE_DISABLE_SAVE_ARRAY
 class SaveArrayStateFile : public ConfigFile {
 public:
   void iterateVariables(VariableOP *op) override {
-    CONFIG_VARIABLE2(sabersense_array_index, 0);  // Default array if no save file.
+
+#ifndef BLADE_DETECT_PIN
+    // Minus 1 allows for extra NO_BLADE array for consistent numbering for end user.
```
profezzorn, line 312: What if they have more than one NO_BLADE entry?

props/saber_sabersense_buttons.h
```diff
@@ -275,27 +288,53 @@ GESTURE CONTROLS
...
       }
-      static void cycle() {
-        return_value = (return_value + 1) % NELEM(blades);
-      }
-    };
-    int SabersenseArraySelector::return_value;
-    
+
+    static void cycle() {
+      int i = return_value;
```
profezzorn, line 329: if return_value is an ohm value, you probably want to scan the array to find the index of the entry that has the given return value, like maybe:

```
int i = 0; for(;i < (int)NELEM(blades); i++) if (blades[i].ohm == return_value) break;
```

props/saber_sabersense_buttons.h
Sabersense-CC: I did try that, but under certain conditions, it stopped the array cycling working when the system didn't use Blade Detect. I haven't quite figured out why, but what I ended up with does work so I kind of left it.

props/saber_sabersense_buttons.h
Sabersense-CC: I did try that too, like NO_BLADE and NO_BLADE + 1, but it didn't do anything and didn't cycle through the NO_BLADE arrays, so I settled on the system only working with one NO_BLADE array. I guess I should add a compile error if people try to use more than one NO_BLADE array?

props/saber_sabersense_buttons.h
profezzorn: "it works" is not a good enough reason.
The code also has to make sense.
It seems like you set the default to zero if you don't set the define, and *that* works, right?

props/saber_sabersense_buttons.h
profezzorn: It's difficult to make a compile error for that.
It's easier to just make the code work with multiple NO_BLADE entries.
Also, I think doing the -1 thing makes the numbering *less* consistent.

Sabersense-CC: OK, I'm working on this by trying to make it so that with a blade detected it cycles through the numbered (ohm) arrays and skips all NO_BLADE arrays, and with no blade detected it does the opposite, i.e. cycles through NO_BLADE arrays and skips ohm arrays.
Is there a flag already available that the system can use to know whether a blade is detected or not, or do I need to make one (like bool isBladeDetected or something like that)?

props/saber_sabersense_buttons.h
Sabersense-CC: > It seems like you set the default to zero if you don't set the define, and _that_ works, right?

Well, yes, the intention is to have the system use the first array as the default if no define is present in the config. But when I set it to zero it didn't work because in arrays with a NO_BLADE array, there was no zero array. When I tried to add a zero array to an array with NO_BLADE (Like NO_BLADE, 0, 1, 2, 3 etc.) it stopped the cycling working. Hence I settled on 1 as default, but that meant when NOT using Blade Detect, it landed on the second array as the default, which in turn prompted the whole minues 1 thing.

Of course reading this back, it feels more hacky and workaroundy than it felt at the time.

props/saber_sabersense_buttons.h
profezzorn: https://github.com/profezzorn/ProffieOS/blob/a0fa24b6a460cf8fba4ded6fd22dfcb77b1f2884/props/prop_base.h#L525

PR #788, opened 2025-04-03
Add OriginalColorChangeMode()

fdarosa2663:
Adding OS7 ColorChangeMode() (as OriginalColorChangeMode() function) to support existing prop features and Edit Mode capabilities properly.


props/saber_fett263_buttons.h
```diff
@@ -5784,7 +5890,11 @@ SaberFett263Buttons() : PropBase() {}
 #ifndef DISABLE_COLOR_CHANGE
         if (EndColorZoom()) return true;
         if (SaberBase::GetColorChangeMode() != SaberBase::COLOR_CHANGE_MODE_NONE) {
+#ifdef MENU_SPEC_TEMPLATE
```
profezzorn, line 5893: Since this is done in many places, maybe have a helper function for it.
Or maybe just do:

```
#ifndef MENU_SPEC_TEMPLATE
void ToggleColorChangeMode() override {
  OriginalColorChangeMode();
}
#endif
```

PR #792, opened 2025-04-15
Update detonator.h

olivierflying747-8:
Added "blank" custom display controller for the detonator.

What effect could be "displayable" on OLED for the detonator?

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -628,7 +628,7 @@ public:
       int noblade_level_before = GetNoBladeLevelBefore();
       FindBladeAgain();
       int noblade_level_after = current_config->ohm / NO_BLADE;
-      
+
       if (noblade_level_before < noblade_level_after) {
         SaberBase::DoEffect(EFFECT_BLADEOUT, 0);
       } else if(noblade_level_before > noblade_level_after) {
@@ -1227,7 +1227,7 @@ public:
     if (b & BUTTON_LEFT) STDOUT.print("Left");
     if (b & BUTTON_RIGHT) STDOUT.print("Right");
     if (b & BUTTON_SELECT) STDOUT.print("Select");
-    if (b & BUTTON_BLADE_DETECT) STDOUT.print("Select");
+    if (b & BUTTON_BLADE_DETECT) STDOUT.print("BladeDetect");
     if (b & MODE_ON) STDOUT.print("On");
   }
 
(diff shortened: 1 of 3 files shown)


common/saber_base.h
```diff
@@ -114,8 +114,11 @@ extern SaberBase* saberbases;
     DEFINE_EFFECT(FONT_DIRECTORY_NOT_FOUND)     \
     /* menu effects, triggered when you go from one menu entry to the next/previous */                             \
     DEFINE_EFFECT(MENU_CHANGE)                  \
-
-
+    /* Detonator  effects */                    \
+    DEFINE_EFFECT(BGNARM)                       \
+    DEFINE_EFFECT(ARMHUM)                       \
+    DEFINE_EFFECT(ENDARM)                       \
```
profezzorn, line 120: Detonator doesn't use endarm, just OFF.

props/detonator.h
```diff
@@ -134,3 +138,174 @@ class Detonator : public PROP_INHERIT_PREFIX PropBase {
...
+  X(endarm)                           \
+  X(boom) // Add a backslash to this line when adding the next line.
+          // The last line in the list needs no backslash at the end!
+//X(***xxx1***)
+//X(***xxx2***)
+//X(***xxx3***)
+
+#ifdef INCLUDE_SSD1306
+
+struct DetonatorDisplayConfigFile : public ConfigFile {
```
profezzorn, line 164: So, is anybody actually going to use this?
A detonator with a display seems weird.
I'm not totally against adding it, but adding something that nobody will use is generally... useless?

props/detonator.h
```diff
@@ -134,3 +138,174 @@ class Detonator : public PROP_INHERIT_PREFIX PropBase {
...
+  DetonatorDisplayConfigFile() { link(&font_config); }
+  void iterateVariables(VariableOP *op) override {
+    CONFIG_VARIABLE2(ProffieOSBgnarmImageDuration, 1000.0f);
+    CONFIG_VARIABLE2(ProffieOSEndarmImageDuration, 1000.0f);
+    CONFIG_VARIABLE2(ProffieOSBoomImageDuration, 1000.0f);
+  //CONFIG_VARIABLE2(ProffieOS***Xxx1***ImageDuration, 1000.0f);
```
profezzorn, line 170: All these extra comments needs to go.

props/detonator.h
profezzorn: Very well, but please remember: no untested code....

common/saber_base.h
profezzorn: If you're going to add EFFECT_* values, you're also going to need to add code that triggers them, otherwise they will do nothing.

PR #793, opened 2025-04-16
Update button_base.h

olivierflying747-8:
extra spaces removed


buttons/button_base.h
```diff
@@ -109,7 +109,7 @@ class ButtonBase : public Looper,
         current_modifiers &=~ button_;
         if (millis() - push_millis_ < BUTTON_SHORT_CLICK_TIMEOUT) {
           SendClick(EVENT_CLICK_SHORT);
-        } else if (millis() - push_millis_ < 2500) {
+        } else if ((millis() - push_millis_) < (BUTTON_HELD_LONG_TIMEOUT + 500)) {
```
profezzorn, line 112: This is not correct.
It either needs to be 2500 or a new define.

PR #795, opened 2025-04-20
Fix increment.h usage comment

NoSloppy:


profezzorn:
If you haven't already, please fix the POD as well.

PR #801, opened 2025-04-22
Add Blackpower for Blade Plug Charging

Sabersense-CC:
Blackpower leaves the pogo pins on a neopixel blade connector live without displaying anything on the blade.
Used for blade plug charging to keep the FETs powered.

This has long been possible by using a few lines of code in the user's config, but since ChargingStylePtr is now baked into ProffieOS, I figured it would be good to have Blackpower as part of it too.

### DIFF
--- a/ProffieOS.ino
+++ b/ProffieOS.ino
@@ -522,6 +522,7 @@ struct is_same_type<T, T> { static const bool value = true; };
 #include "styles/edit_mode.h"
 #include "styles/pixelate.h"
 #include "styles/display.h"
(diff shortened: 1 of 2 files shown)


profezzorn:
ChargingStylePtr<> and ChargingStyle<> already keeps the pins on. That's the whole point of those classes, so you can just use them with "Black", and that should do the same thing.

Sabersense-CC:
Ah, sorry - I didn't realise that ChargingStylePtr handled that already. Apologies.

PR #802, opened 2025-04-24
Bugfix saber_sabersense_buttons.h

Sabersense-CC:
Slight bugette found in which first Blade-In array gets selected twice after reboot or system reset when using Blade Detect with array selector.
Now fixed.

### DIFF
--- a/props/saber_sabersense_buttons.h
+++ b/props/saber_sabersense_buttons.h
@@ -1,4 +1,4 @@
-/* V7/8-201.
+/* V7/8-214.
 ============================================================
 =================   SABERSENSE PROP FILE   =================
 =================            by            =================
@@ -294,6 +294,7 @@ GESTURE CONTROLS
 #define PROPS_SABER_SABERSENSE_BUTTONS_H
 
 #ifdef SABERSENSE_ARRAY_SELECTOR
+
 #ifndef SABERSENSE_DISABLE_SAVE_ARRAY
 class SaveArrayStateFile : public ConfigFile {
 public:
@@ -309,12 +310,23 @@ public:
       float id() {
         return return_value;
       }
+
       static void cycle() {
+#ifndef BLADE_DETECT_PIN
         return_value = (return_value + 1) % NELEM(blades);
       }
     };
     int SabersenseArraySelector::return_value;
-    
+#else   // Improved logic fixes early Array1 repetition when using Blade Detect.
+        return_value++;
+        if (return_value == 0 || return_value >= NELEM(blades)) {
+          return_value = 1;
+        }
+      }
+    };
+    int SabersenseArraySelector::return_value = 1;
+#endif
+
 #undef BLADE_ID_CLASS_INTERNAL
 #define BLADE_ID_CLASS_INTERNAL SabersenseArraySelector
 #undef BLADE_ID_CLASS


props/saber_sabersense_buttons.h
```diff
@@ -309,12 +310,23 @@ class SaveArrayStateFile : public ConfigFile {
       float id() {
         return return_value;
       }
+
       static void cycle() {
+#ifndef BLADE_DETECT_PIN
```
profezzorn, line 315: Please change this so that the { and } are balanced within the ifdefs, or you're going to confuse the reader, and some editors.

Easiest way is probably to use two ifdefs.

Sabersense-CC: Done.

Sabersense-CC: Soz - just wanted to tweak a line space for clarity.
Now definitely done done.

props/saber_sabersense_buttons.h
```diff
@@ -310,11 +311,23 @@ class SaveArrayStateFile : public ConfigFile {
         return return_value;
       }
       static void cycle() {
+#ifndef BLADE_DETECT_PIN
         return_value = (return_value + 1) % NELEM(blades);
+#else   // Improved logic fixes early Array1 repetition when using Blade Detect.
+        return_value++;
```
profezzorn, line 317: how about:
```cpp
   return_value = 1 + return_value % (NELEM(blades) - 1);
```

props/saber_sabersense_buttons.h
```diff
@@ -310,11 +311,23 @@ class SaveArrayStateFile : public ConfigFile {
...
+        }
+#endif
       }
     };
+
+#ifndef BLADE_DETECT_PIN
     int SabersenseArraySelector::return_value;
-    
+#else    
+    int SabersenseArraySelector::return_value = 1;
```
profezzorn, line 328: Didn't I see this in the other review too?

props/saber_sabersense_buttons.h
Sabersense-CC: Sorry for the delay, but yes, I've tested that and it does indeed work well.

But in testing that, I found another... bug is the wrong word, because it isn't, but it is an aspect that might preclude certain legit uses...

Specifically, if the user accidentally attempts an array select with no blade in the hilt (remember the prop only permits one no-blade array), then it includes that attempt in its count, meaning that when you subsequently fit a blade, it will always go to blade array 2. This means if you've used the array select as a means of accommodating different blade lengths - let's say array 4 matches your blade - and you mistakenly attempt to cycle arrays with no blade fitted, when you refit the blade, it will now be set to the wrong length.

I've been trying to implement a lock so that if using Blade Detect while no blade is detected, attempting to select a new array gets ignored, but no success so far. I thought there was a global blade_present() function, but my memory must be playing tricks because it doesn't work and I can't find it in the OS.

Sabersense-CC: Doh! Just found blade_detected - which I'm guesing is the key. Bear with.....

Sabersense-CC: Don't you just hate it when a really simple solution eludes you for several hours!
Now sorted.

Sabersense-CC:
OK, your shorter code line has been aded and I've also added that lockout to prevent invalid no_blade array cycling messing up array values.

PR #809, opened 2025-05-05
Add to and improve preset skipping - Sabersense prop

Sabersense-CC:
Adds 5 and 10 preset skipping to 1 button systems;
All skipping now disableable via define;
Two levels of skip value, both can be user defined in config;
Improves and standardizes skipping logic.


Sabersense-CC:
Sorry, I think I did that wrong - it pulled the Blackpower thing over which I didn't want. Still getting my head around GitHub.....

PR #810, opened 2025-05-05
Improved preset skipping in Sabersense Prop.

Sabersense-CC:
Adds 5 and 10 preset skipping to 1 button systems;
All skipping now disableable via define;
Two levels of skip value, both can be user defined in config;
Improves and standardizes skipping logic.


props/saber_sabersense_buttons.h
```diff
@@ -310,11 +329,23 @@ class SaveArrayStateFile : public ConfigFile {
         return return_value;
       }
       static void cycle() {
+#ifndef BLADE_DETECT_PIN
         return_value = (return_value + 1) % NELEM(blades);
+#else   // Improved logic fixes early Array1 repetition when using Blade Detect.
+        return_value++;
```
profezzorn, line 335: why not:

```cpp
  return_value = 1 + return_value % (NELEM(blades) - 1);
```

?

props/saber_sabersense_buttons.h
```diff
@@ -310,11 +329,23 @@ class SaveArrayStateFile : public ConfigFile {
...
+        }
+#endif
       }
     };
+
+#ifndef BLADE_DETECT_PIN
     int SabersenseArraySelector::return_value;
-    
+#else    
+    int SabersenseArraySelector::return_value = 1;
```
profezzorn, line 346: if you're going to have =1 here you should probably have = 0 above.

props/saber_sabersense_buttons.h
```diff
@@ -798,25 +847,77 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
...
+
+#if NUM_BUTTONS == 1
+    // First skip number (define A - default 5)
+    case EVENTID(BUTTON_POWER, EVENT_SECOND_HELD_MEDIUM, MODE_OFF): {
+      if (fusor.angle1() > M_PI / 6) {
+        // Pointing up, skip forward
+        SetPreset(current_preset_.preset_num + SABERSENSE_SKIP_A, true);
+      } else if (fusor.angle1() < -M_PI / 6) {
+        // Pointing down, skip backward
+        SetPreset(current_preset_.preset_num - SABERSENSE_SKIP_A, true);
```
profezzorn, line 863: This doesn't actually work when the presets are loaded from presets.ini.
Only -1 works, -2, -3, etc. does not work.
You'll need to figure out how many presets there are and then do something like:

```
 SetPreset((current + number_of_presets - skip) % number_of_presets);
```

props/saber_sabersense_buttons.h
```diff
@@ -798,25 +847,77 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
...
+      }
+#ifdef SAVE_PRESET
+      SaveState(current_preset_.preset_num);
+#endif
+      return true;
+    }
+
+    // Second skip number (define B - default 10)
+    case EVENTID(BUTTON_POWER, EVENT_THIRD_HELD_MEDIUM, MODE_OFF): {
+      if (fusor.angle1() > M_PI / 6) {
```
profezzorn, line 877: Sems like this is mostly the same as above, so it would be good to have a helper function that just takes the skip amount as an argument...

props/saber_sabersense_buttons.h
```diff
@@ -825,15 +926,14 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
     case EVENTID(BUTTON_AUX, EVENT_FIRST_HELD_LONG, MODE_OFF):
 #endif
     case EVENTID(BUTTON_POWER, EVENT_FIRST_HELD_LONG, MODE_OFF):
-#define DEGREES_TO_RADIANS (M_PI / 180)
-      if (fusor.angle1() > 45 * DEGREES_TO_RADIANS) {
-        // If pointing up
+      if (fusor.angle1() > M_PI / 6) {
```
profezzorn, line 929: this could also use the helper function, but with 1 as the skip, right?

props/saber_sabersense_buttons.h
```diff
@@ -362,6 +393,25 @@ class SaveArrayStateFile : public ConfigFile {
...
+#ifndef SABERSENSE_SKIP_A
+#define SABERSENSE_SKIP_A 5
+#endif
+
+#ifndef SABERSENSE_SKIP_B
+#define SABERSENSE_SKIP_B 10
+#endif
+
+#ifndef SABERSENSE_BATTERY_CHECK
+#define SABERSENSE_BATTERY_CHECK()                                         \
```
profezzorn, line 405: Why is this a define instead of a function?

props/saber_sabersense_buttons.h
Sabersense-CC: As above - will sort with the other PR.

props/saber_sabersense_buttons.h
Sabersense-CC: I'm not sure it can. The last stage (horizontal) of this one is different to the bool driving the other two - in those horizontal is a battery level indicator, in this one it's a skip to the middle preset.
That said, I have tweaked it a bit to make it more consistent with the others.

props/saber_sabersense_buttons.h
Sabersense-CC: I'm confused. I've just tested it as per my latest update, and it seems to work flawlessly under all conditions. I tried colour changing a preset in order to create the presets.ini file, switched around some presets, shutdown, booted up swicthed around some more, in singles and multis, and it all worked perfectly. I can't make it _not_ work with the line in question. What am I missing here?

props/saber_sabersense_buttons.h
Sabersense-CC: Helper function done.

props/saber_sabersense_buttons.h
Sabersense-CC: It isn't any more - now a function and working well. :)

props/saber_sabersense_buttons.h
```diff
@@ -466,6 +495,28 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
 #endif
   }
 
+    // Used for 1 button Preset Skipping.
+    // Uses SABERSENSE_SKIP_A/B for skip values.
+#ifndef SABERSENSE_DISABLE_SKIPPING
```
profezzorn, line 500: There is no need to use #ifndef for the function. If it's never called, it will not take up any space.

props/saber_sabersense_buttons.h
```diff
@@ -466,6 +495,28 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
...
+  bool HandleSkipEvent(int skip_value) {
+    if (fusor.angle1() > M_PI / 6) {
+      // Pointing up, skip forward
+      SetPreset(current_preset_.preset_num + skip_value, true);
+    } else if (fusor.angle1() < -M_PI / 6) {
+      // Pointing down, skip backward
+      SetPreset(current_preset_.preset_num - skip_value, true);
```
profezzorn, line 507: This doesn't work, as discussed earlier.

props/saber_sabersense_buttons.h
```diff
@@ -466,6 +495,28 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
...
+    } else {
+      // Horizontal, perform battery check
+      BatteryChecker();
+      return true;
+    }
+#ifdef SAVE_PRESET
+    SaveState(current_preset_.preset_num);
```
profezzorn, line 514: This can interrupt the "font" sound, so it's better to do it before SetPreset if possible.

props/saber_sabersense_buttons.h
Sabersense-CC: #ifndef removed.

props/saber_sabersense_buttons.h
Sabersense-CC: Sorry, I'm still confused by this. What is it that doesn't work about it? I've tested it under every condition I can think of - with and without a presets.ini, booting after saves, leaving it in different arrays or on different presets, cycling presets singly and in multiples - and it always work perfectly. So I'm not quite sure what I need to aim at to change.

props/saber_sabersense_buttons.h
Sabersense-CC: I'll try re-ordering and do some tests....

Sabersense-CC: Hang on - if we do SaveState _before_ we do SetPreset, doesn't that mean the system will save the preset we just left rather than saving the preset we just landed on?

Sabersense-CC: How about a setup like we did for playing font sounds after array sounds? i.e. I could build it so the SaveState only happens after the font sound effect has finished playing - that way there's no risk one impacting upon the other. Would that work as a solution?

Sabersense-CC: Hmmm... I've been looking into this some more, and although it can work, we have to allow for a lot of variables. Switching presets will just play font.wav, but we want the same SaveState to happen when we switch arrays, and switching arrays can play font.wav, array.wav, bladeid.wav, array.wav followed by font.wav, or bladeid.wav followed by font.wav.  If we're going to tell the system to wait until the correct combination of wavs has played before saving, it's going to take a fair amount of code (I'm up to around 80 lines and I suspect it will grow a little more before it's nailed down).
It makes me wonder if it's worth it given that I've yet to experience any issues with this particular feature as it is?

Sabersense-CC: Or another approach - we could tell it, "Once you've had half a second of no sound, do the SaveState." That way we don't need to track which of the three sounds above is playing. But of course that's not without its problems. For instance the save could now interfere with the user actually using the blade. Or alternatively if the user lights up a quarter second after the font wav, the save gets delayed again. If they shut off and switch font or array very rapidly, we're trying to move to a new preset without having saved the last one and now we're getting pretty lost.
All of which I think precludes any of these solutions, and takes us back to your original suggestion and SaveState before SetPreset, but I can't see how that can work - it's like leaving your house to drive to a hotel, and then being given your house address to tell you where to go.
Again, what am I missing here?

Sabersense-CC:
OK, I've revisited this with fresh eyes and the benefit of a few days away from it, and I think the penny has finally dropped as to the nature of the first issue you originally flagged - it's to do with wrapping isn't it. Now that I can see the problem, I can try and find the answer.
The second issue about the savestate possibly interefering with font playback I'm still puzzled by.
But one step at a time - I'll look into the wrapping issue over the next few days and try and figure that one out, and we'll go from there.

Sabersense-CC:
OK, I'm hoping I've fixed the edge case problems with this. I've set it so that with big skips (5 or 10 presets at a time) it caps at the ends instead of wraps - I think it would be odd for a user to skip forward from preset 33 on a 36 preset system and find themselves back at preset 7.
I've hopefully also sorted the save/font file thing, so I think the whole thing is now ready to go! Fingers crossed! :)

props/saber_sabersense_buttons.h
```diff
@@ -465,6 +490,61 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
...
 #endif
   }
+  
+  // PRESET SKIPPING
+  // Skip Value definable.
+  // One Button System - Up, Horizontal, Down.
+  bool OneBtnMultiFontSkip(int skip_value) {
+      int count = 0;
+      CurrentPreset tmp;
+      for (int i = 0; ; i++) {
```
profezzorn, line 500: There is an easier way to count presets:

You can see it here:

https://github.com/profezzorn/ProffieOS/blob/3c60a823d4da314134870e2e9c91966132b7b91d/props/saber_fett263_buttons.h#L2988

props/saber_sabersense_buttons.h
```diff
@@ -465,6 +490,61 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
...
+
+      if (angle < -M_PI / 6) { // Pointing down
+          new_index -= skip_value;
+      } else if (angle > M_PI / 6) { // Pointing up
+          new_index += skip_value;
+      } else { // Horizontal
+          SpeakBatteryLevel();
+          return true;
+      }
+      // Capping, not wrapping - less disorientating for user.
```
profezzorn, line 516: This seems very strange to me, but whatever.

props/saber_sabersense_buttons.h
```diff
@@ -465,6 +490,61 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
...
+      if (new_index < 0) new_index = 0;
+      if (new_index >= count) new_index = count - 1;
+      // Save state before setting the preset
+      SetPreset(new_index, true);
+#ifdef SAVE_PRESET
+      SaveState(new_index);
```
profezzorn, line 522: SaveState is usually done before SetPreset to avoid glitches while playing font.wav

props/saber_sabersense_buttons.h
```diff
@@ -465,6 +490,61 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
...
+#endif
+      return true;
+  }
+
+  // Two Button System - Up, Down.
```
profezzorn, line 527: Why is this different from one-button skipping?

Also, why not just do something like:

```
#if NUM_BUTTONS == 2
bool MultiFontSkip(int delta) {
    ...
}
#elif NUM_BUTTONS == 1
bool MultiFontSkip(int delta) {
}
#endif
```

Not sure, but it might simplify some code below.

Sabersense-CC: With two-button I've deliberately left it with looser angle requirements, so that just above horizontal counts as up, i.e. users can benefit from more ease of use. But with one button, there's no choice but to demand stricter angles in order to pack all those features in, hence the differing angle logic. I know it takes a little more code and perhaps comes at the expense of code readability, but I stand by the premise.
Anyway I'll get those other tweaks sorted in the next day or so.

Sabersense-CC: OK, update, those changes you mentioned are now implemented, subject to the angle/clamping caveats mentioned above and in the other PR.
Sorry about the crossover between PRs. (I think I'll stick to one at a time in future, as it's easier to keep track). Not sure if it helps or not but I have a complete prop file myself with all the latest versions of everything we've done across all these PRs which I can send across if it helps.

props/saber_sabersense_buttons.h
```diff
@@ -798,50 +889,68 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
...
-    case EVENTID(BUTTON_POWER, EVENT_FIRST_HELD_LONG, MODE_OFF):
-#define DEGREES_TO_RADIANS (M_PI / 180)
-      if (fusor.angle1() > 45 * DEGREES_TO_RADIANS) {
-        // If pointing up
-        SetPreset(0, true);
-      } else if (fusor.angle1() < -45 * DEGREES_TO_RADIANS) {
-        // If pointing down
-        SetPreset(-1, true);
+    case EVENTID(BUTTON_POWER, EVENT_FIRST_HELD_LONG, MODE_OFF): {
+      CurrentPreset tmp;
```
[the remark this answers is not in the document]
Sabersense-CC: > Isn't this the MultiFontSkip code again?
Not quite. MutliFontSkip jumps forward or backward a specified number of presets, i.e. it has to count how many, as it were, and then figure out where to land. Hot skipping by contrast jumps to a specific named preset - first middle or last by default, with middle and last being overridable by the user with a define if they want it. So the two functions, although related, are distinct from one another.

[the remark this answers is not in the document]
profezzorn: Right. Isn't this code in a different PR?
What is it doing here?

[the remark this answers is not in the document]
Sabersense-CC: I fear I've careered into the buffers of my limited GitHub knowledge. I thought I'd kept them separate as far as possible, but the nature of it means there was a bit of crossover. Sincere apologies Prof, but I don't quite know where to go from here. All of font skipping - both leaping over set numbers of fonts and hot skipping to named fonts, is handled in two chunks of code.
Will definitely make sure in future that if I have more than one PR on the go at once, that thy are about completely distinct, separate functions or features.
Would it help if I deleted these two PRs and created a new one covering all of font skipping?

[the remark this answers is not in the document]
Sabersense-CC: OK, #815 should now be correct with just the Hot Skip code included therein.
Hoping I've done it right.

props/saber_sabersense_buttons.h
```diff
@@ -476,6 +501,45 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
...
+    return tmp.preset_num + 1;
+  }
+
+  // Two Button system: Up/Down only - loose angle settings for ease of use.
+  bool MultiFontSkip(int skip_value) {
+      float angle = fusor.angle1();
+      int delta = 0;
+
+#if NUM_BUTTONS == 2
+      delta = (angle < -M_PI / 4) ? -skip_value : skip_value;
```
profezzorn, line 518: shouldn't this be angle < 0 ? (above/below horizon)

Right now the cutoff is 45 degrees I think?

props/saber_sabersense_buttons.h
```diff
@@ -808,50 +882,68 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
...
-      if (fusor.angle1() > 45 * DEGREES_TO_RADIANS) {
-        // If pointing up
-        SetPreset(0, true);
-      } else if (fusor.angle1() < -45 * DEGREES_TO_RADIANS) {
-        // If pointing down
-        SetPreset(-1, true);
+    case EVENTID(BUTTON_POWER, EVENT_FIRST_HELD_LONG, MODE_OFF): {
+      CurrentPreset tmp;
+      tmp.SetPreset(-1);
+      int num_presets = tmp.preset_num + 1;
```
profezzorn, line 917: use GetNumberOfPresets()

props/saber_sabersense_buttons.h
```diff
@@ -808,50 +882,68 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
...
+      if (angle > M_PI / 6) {
+        target_preset = 0;
+      } else if (angle < -M_PI / 6) {
+        target_preset = hot_skip_down;
       } else {
-        // If horizontal
-        CurrentPreset tmp;
-        tmp.SetPreset(-1);
-        SetPreset(tmp.preset_num / 2, true);
+        target_preset = hot_skip_level;
```
profezzorn, line 937: Put all the hot_skip_level calculations here.

props/saber_sabersense_buttons.h
```diff
@@ -808,50 +882,68 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
...
+        hot_skip_level = num_presets / 2;
+      }
+
+      int target_preset;
+      float angle = fusor.angle1();
+
+      if (angle > M_PI / 6) {
+        target_preset = 0;
+      } else if (angle < -M_PI / 6) {
+        target_preset = hot_skip_down;
```
profezzorn, line 935: Put all the hot_skip_down calculations down.

PR #812, opened 2025-05-16
add get_variation

NoSloppy:


props/prop_base.h
```diff
@@ -1764,6 +1764,10 @@ class PropBase : CommandParser, protected Looper, protected SaberBase, public Mo
       ToggleColorChangeMode();
       return true;
     }
+    if (!strcmp(cmd, "get_variation")) {
+      STDOUT.println(current_preset_.variation);
```
profezzorn, line 1768: This should probably print the variation from SaberBase. (They are probably the same most of the time though.)

NoSloppy: I had considered that. This was a request from ShtokyD, as he wanted to be able to get current variation via serial on the phone. I kind of assumed to use current preset's. He said it works great like this.

profezzorn: The "variation" command uses SaberBase::SetVariation, and unless there is a very good reason for it, this command should be doing the inverse of that.

"it works" is not a good enough reason.

NoSloppy: Ok. How about now?

props/prop_base.h
```diff
@@ -1709,7 +1709,7 @@ class PropBase : CommandParser, protected Looper, protected SaberBase, public Mo
     }
     if (!strcmp(cmd, "get_volume")) {
 #ifdef ENABLE_AUDIO
-      STDOUT.println(dynamic_mixer.get_volume());
```
profezzorn, line 1712: Actually. this (and below) should not use PVLOG, because no log level should be able to turn it off. Both should use STDOUT.

NoSloppy: adjusted.

props/prop_base.h
```diff
@@ -1709,6 +1709,7 @@ class PropBase : CommandParser, protected Looper, protected SaberBase, public Mo
     }
     if (!strcmp(cmd, "get_volume")) {
 #ifdef ENABLE_AUDIO
+      STDOUT.print("volume ");
```
profezzorn, line 1712: This will break the ProffieOS workbench.

NoSloppy:
I've asked ShtokyD to chime in here, as it's really he who can explain what he can and can't do on his end.

profezzorn:
These arguments makes no sense.

PR #815, opened 2025-05-17
Add User Definable "Home" Preset

Sabersense-CC:
Intended for users who may have several sabers and they want all the same fonts on all sabers to be in the same order, so that preset navigation is the same for each hilt. But they still want a 'default' font to be the appropriate font for that hilt. This feature makes that possible by letting the user define the home preset which replaces the skip-to-last preset. (Skip-to-last is easiest to lose because skip to first takes you one preset away from last, should you want it, due to wrapping).

### DIFF
--- a/props/saber_sabersense_buttons.h
+++ b/props/saber_sabersense_buttons.h
@@ -1,4 +1,4 @@
-/* V7/8-201.
+/* V7/8-201d.
 ============================================================
 =================   SABERSENSE PROP FILE   =================
 =================            by            =================
@@ -360,6 +360,10 @@ public:
 
 #ifndef BUTTON_HELD_LONG_TIMEOUT
 #define BUTTON_HELD_LONG_TIMEOUT 2000
+#endif
+
+#ifndef SABERSENSE_USER_HOME_FONT
+#define SABERSENSE_USER_HOME_FONT 0
 #endif
 
 EFFECT(dim);      // for EFFECT_POWERSAVE
@@ -819,29 +823,37 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
       return true;
 #endif
 
-    // Skips to first preset (up) or last preset (down)
-    // or middle preset if horizontal:
+    // Skips to first preset (up), last or user-defined preset (down),
+    // or middle preset (horizontal):
 #if NUM_BUTTONS == 2
     case EVENTID(BUTTON_AUX, EVENT_FIRST_HELD_LONG, MODE_OFF):
 #endif
-    case EVENTID(BUTTON_POWER, EVENT_FIRST_HELD_LONG, MODE_OFF):
-#define DEGREES_TO_RADIANS (M_PI / 180)
-      if (fusor.angle1() > 45 * DEGREES_TO_RADIANS) {
-        // If pointing up
+    case EVENTID(BUTTON_POWER, EVENT_FIRST_HELD_LONG, MODE_OFF): {
+      // Calculate user font, defaulting to -1 for last preset if not set by user.
+      int user_font = (SABERSENSE_USER_HOME_FONT > 0) ? (SABERSENSE_USER_HOME_FONT - 1) : -1;
+      CurrentPreset tmp;
+      tmp.SetPreset(-1);
+      int num_presets = tmp.preset_num + 1;
+      // Check user_font is valid (0 to num_presets-1). Clamp to last if not.
+      if (user_font >= num_presets || user_font < 0) {
+        user_font = num_presets - 1;
+      }
+      if (fusor.angle1() > M_PI / 6) {
+        // Pointing up, skip to first (home) preset
         SetPreset(0, true);
-      } else if (fusor.angle1() < -45 * DEGREES_TO_RADIANS) {
-        // If pointing down
-        SetPreset(-1, true);
+      } else if (fusor.angle1() < -M_PI / 6) {
+        // Pointing down, skip to last (or user-defined) preset
+        SetPreset(user_font, true);
       } else {
-        // If horizontal
-        CurrentPreset tmp;
-        tmp.SetPreset(-1);
-        SetPreset(tmp.preset_num / 2, true);
+        // Horizontal, skip to middle preset
+        SetPreset(num_presets / 2, true);
       }
 #ifdef SAVE_PRESET
-      SaveState(current_preset_.preset_num);
+        SaveState(current_preset_.preset_num);
 #endif
-      return true;
+        break;
+    }
+#endif
 
     // BLADE ID OPTIONS AND ARRAY NAVIGATION
     // Blade ID on-demand scanning with BladeID audio idents.


Sabersense-CC:
Sorry, slight tweak - it occurred to me that some people might want to define both horizontal and down preset skips. Pointing up stays fixed at skipping back to the first font/preset.

props/saber_sabersense_buttons.h
```diff
@@ -819,29 +848,43 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
...
-#define DEGREES_TO_RADIANS (M_PI / 180)
-      if (fusor.angle1() > 45 * DEGREES_TO_RADIANS) {
-        // If pointing up
+    case EVENTID(BUTTON_POWER, EVENT_FIRST_HELD_LONG, MODE_OFF): {
+      CurrentPreset tmp;
+      tmp.SetPreset(-1);
+      int num_presets = tmp.preset_num + 1;
+
+      // Get and check user values. Clamp to last preset if invalid.
+      int hot_skip_down = (SABERSENSE_HOT_SKIP_DOWN > 0) ? (SABERSENSE_HOT_SKIP_DOWN - 1) : -1;
```
profezzorn, line 862: This code is a bit weird.
It calculates two different values, then depending on some if statements below, it uses one of those values (or zero), and the other value is never used.

It would be better to do it sort of like this:

```
   int target = 0;
   if (up) {
      target = 0;
   } else if (down) {
      target = calculate hot skip down;
   } else {
      target = calculate hot skip;
   }
   SetPreset(target, true);
```

This way we only calculate the value we need, and there is only one call to SetPreset with that value.

Sabersense-CC: OK, I've done some digging and if I may be so bold, I'm going to try to defend it as it is. The crux of it is that I want slightly different behaviours between one and two button systems.
As mentioned on the other PR, with two-button I want it with looser angle requirements and to only have up or down options, so that just above horizontal counts as up, i.e. users can benefit from less strict, easier usage.
But with one button, there's no choice but to have three options - up, level and down - in order to pack all the features in. Hence we have to demand stricter angles in order for it to work, but it's a shame to impose those stricter angles on 2 button when they're not necessary.
But although we have dual angle handling, we still want the capping/clamping to work the same.
Hence it made sense to have the capping/clamping code centralized in one place which is what we ended up with.
That said, I gather there is a slight computational penalty, but my understanding is that it's very minimal and that the code clarity was preferable.
Full disclosure too - I'm swimming a little out of my depth here, so apologies if I'm grasping the wrong end of the stick.

profezzorn: Did you maybe respond to the wrong comment? This comment seems like it is relating to a different PR?

Sabersense-CC: Sorry, yes, there's a bit of an overlap between the two remaining PRs as they both deal with different bits of kind of the same thing. Would it help if I somehow uploaded my own file which has the most recent of everything incorporated into it?

profezzorn: that sounds like it would muddle things up even more

Sabersense-CC: OK, now updated and hopefully good to go.

Sabersense-CC:
This PR should hopefully now be good to go as it just has the Hot Skipping code included in it - pending any necessary relevant corrections of course.

props/saber_sabersense_buttons.h
```diff
@@ -368,8 +384,13 @@ class SaveArrayStateFile : public ConfigFile {
 #define BUTTON_HELD_MEDIUM_TIMEOUT 1000
 #endif
 
-#ifndef BUTTON_HELD_LONG_TIMEOUT
-#define BUTTON_HELD_LONG_TIMEOUT 2000
+// Hot Skip values are placeholders only.
+#ifndef SABERSENSE_HOT_SKIP_DOWN
+#define SABERSENSE_HOT_SKIP_DOWN 0
```
profezzorn, line 389: Why not have this zero-indexed and default it to -1?
Then you need no calculations below.

props/saber_sabersense_buttons.h
```diff
@@ -829,29 +850,43 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
...
+        hot_skip_level = num_presets / 2;
+      }
+
+      int target_preset;
+      float angle = fusor.angle1();
+
+      if (angle > M_PI / 6) {
+        target_preset = 0;
+      } else if (angle < -M_PI / 6) {
+        target_preset = hot_skip_down;
```
profezzorn, line 879: Move all the calculations for "hot_skip_down" to here.

props/saber_sabersense_buttons.h
```diff
@@ -829,29 +850,43 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
...
       } else {
-        // If horizontal
-        CurrentPreset tmp;
-        tmp.SetPreset(-1);
-        SetPreset(tmp.preset_num / 2, true);
+        target_preset = hot_skip_level;
```
profezzorn, line 881: Move all the calculations for "hot_skip_level" to here.

props/saber_sabersense_buttons.h
```diff
@@ -368,8 +384,13 @@ class SaveArrayStateFile : public ConfigFile {
...
 #endif
 
-#ifndef BUTTON_HELD_LONG_TIMEOUT
-#define BUTTON_HELD_LONG_TIMEOUT 2000
+// Hot Skip values are placeholders only.
+#ifndef SABERSENSE_HOT_SKIP_DOWN
+#define SABERSENSE_HOT_SKIP_DOWN 0
+#endif
+
+#ifndef SABERSENSE_HOT_SKIP_LEVEL
```
profezzorn, line 392: I wonder if these hot skip things should maybe be percentages?
That way you could do:

```
#define SABERSENSE_HOT_SKIP_DOWN 33
#define SABERSENSE_HOT_SKIP_LEVEL 66
```

to get to 1/3rd and 2/3rds of the list...

Sabersense-CC: I don't think that would be very user friendly as people would need to calculate how far in their chosen preset is. If they have a simple list of, say, 20 presets and they want number 12 as their hot skip number, that's super-simple and understandable. I think anything else adds too much complexity for the user.

props/saber_sabersense_buttons.h
Sabersense-CC: I think that makes it less intuitive for the user. If they have a paper list of presets and they want the sixth one on the list to be the skip target, they want to enter 6, not 5. I appreciate this creates another calculation for the system, but I think it's fair to let the system do the legwork to make is super-simple for the user.

props/saber_sabersense_buttons.h
Sabersense-CC: Yep, will get that sorted and updated by teatime tomorrow.

props/saber_sabersense_buttons.h
Sabersense-CC: As above - willdo...

profezzorn:
Seems to have a lot of overlap with the other code review?
(Or is it just very similar?)

profezzorn:
Right, I forgot.
Still some comments that need to be sorted out in this one though.

Sabersense-CC:
> Right, I forgot. Still some comments that need to be sorted out in this one though.

OK, should be done now. Apologies - I thought I'd done that re-order thing but it seems I hadn't. Should be there now.

PR #817, opened 2025-05-22
Fallthrough boot to font

NoSloppy:
Does this make sense to do?


profezzorn:
I think it will just muddle up how people make fonts.

PR #821, opened 2025-05-31
BCprop - fix vol menu sounds

NoSloppy:
Sounds would "pile up" in the queue due to new calls accruing during old sound playback.
Now they are not queued if sounds are playing or if rotating very fast.
Maximum and minimum sounds take precedence.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -1219,12 +1219,6 @@ struct BCVolumeMode : public SPEC::SteppedMode {
   int initial_percentage_ = 0;
   int percentage_ = 0;
 
-  // Debounce timer for sounds during rotation steps
-  uint32_t trigger_time_ = millis();
-  // Track whether a range‐end prompt is currently playing
-  enum class RangeState { None, SayingMin, SayingMax };
-  RangeState rangeState_ = RangeState::None;
-
   void mode_activate(bool onreturn) override {
     initial_volume_ = dynamic_mixer.get_volume();
     initial_percentage_ = round((initial_volume_ / max_volume_) * 10) * 10;
@@ -1257,10 +1251,10 @@ struct BCVolumeMode : public SPEC::SteppedMode {
     mode::getSL<SPEC>()->SayVolumeMenuEnd();
     SPEC::SteppedMode::mode_deactivate();
 
-    // Exit printout already handled in select() / exit()
+    // "Exit Volume Menu" printout already handled in select() / exit()
     if (percentage_ <= 10) PVLOG_NORMAL << "** Minimum Volume\n";
     else if (percentage_ >= 100) PVLOG_NORMAL << "** Maximum Volume\n";
-    else PVLOG_NORMAL << "** Volume " << percentage_ << "%\n";
+    else PVLOG_NORMAL << "** Final Volume " << percentage_ << "%\n";
   }
 
   void next() override {
@@ -1270,11 +1264,9 @@ struct BCVolumeMode : public SPEC::SteppedMode {
       if (current_volume_ >= max_volume_) {
         current_volume_ = max_volume_;
         QuickMaxVolume();
-      } else if (millis() - trigger_time_ > 100 && !sound_library_.busy()) {
-        // If anything is playing or 100ms hasn't passed, wait to queue SayVolumeUp
+      } else {
+        sound_library_.fadeout(0.2);
         mode::getSL<SPEC>()->SayVolumeUp();
-        rangeState_ = RangeState::None;
-        trigger_time_ = millis();
       }
       // Always update the volume even if we skipped SayVolumeUp
       dynamic_mixer.set_volume(current_volume_);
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -1253,9 +1270,13 @@ struct BCVolumeMode : public SPEC::SteppedMode {
       if (current_volume_ >= max_volume_) {
         current_volume_ = max_volume_;
         QuickMaxVolume();
-      } else {
+      } else if (millis() - trigger_time_ > 100 && !sound_library_.busy()) {
```
profezzorn, line 1273: Why do you need to check both busy() and trigger_time_ ?
Seems like a suspenders-and-belt solution?

props/saber_BC_buttons.h
```diff
@@ -1253,9 +1270,13 @@ struct BCVolumeMode : public SPEC::SteppedMode {
...
+        // If anything is playing or 100ms hasn't passed, wait to queue SayVolumeUp
```
profezzorn, line 1274: This comment doesn't seem to make sense. I mean, if we are here in the code, then nothing is playing and 100ms has indeed passed...

props/saber_BC_buttons.h
```diff
@@ -1267,28 +1288,43 @@ struct BCVolumeMode : public SPEC::SteppedMode {
...
+        rangeState_ = RangeState::None;
+        trigger_time_ = millis();
       }
       dynamic_mixer.set_volume(current_volume_);
     }
   }
 
   void QuickMaxVolume() {
     dynamic_mixer.set_volume(max_volume_);
+    // Always play immediately when Max reached
```
profezzorn, line 1302: If we're always playing immediately, then why not just call fadeout() unconditionally?
Why check rangeState_ ?

props/saber_BC_buttons.h
NoSloppy: Trigger Time 100 typically ends up being longer than the actual sound length of the small click sound (that's good). Because I wanted to space those click sounds out but also handle if a longer prompt such as maximum volume was spoken.

profezzorn: I don't see how this spaces out click sounds?

NoSloppy: Because if the click sound is only 20 milliseconds, too many queue up with fast rotation. This is limiting calls to sound library to a minimum of 100 milliseconds.

profezzorn: but how? There is no click playing in here

props/saber_BC_buttons.h
NoSloppy: We're not always playing immediately. Only max/min. See longer reply above.

props/saber_BC_buttons.h
profezzorn: Wait, I think I'm confused "SayVolumeDown" is the tick, isn't it?
I was thinking that this was the min/max case, let me take another look at the code.

props/saber_BC_buttons.h
```diff
@@ -1267,28 +1288,43 @@ struct BCVolumeMode : public SPEC::SteppedMode {
...
+      sound_library_.fadeout(SaberBase::sound_length);
+    }
     PVLOG_NORMAL << "** Maximum Volume\n";
     mode::getSL<SPEC>()->SayMaximumVolume();
+    rangeState_ = RangeState::SayingMax;
   }
 
   void QuickMinVolume() {
     dynamic_mixer.set_volume(min_volume_);
+    if (rangeState_ == RangeState::SayingMin || rangeState_ == RangeState::SayingMax &&
```
profezzorn, line 1314: It seems very unlikely that rangeState_ would be RangeState::SayingMax when you get here. Like, that would only happen when you skip from max to min without any of the steps in between.

It also seems like if it's SayingMin, we should just let it finish rather than fading it out?

To me, it seems like the logic should be:

```
   if (rangeState_ == RangeState::SayingMin && sound_ibrary_.busy()) return;
   sound_library_.fadeout(0.1);
```

props/saber_BC_buttons.h
profezzorn: I'm not a big fan of hard-coding time limits.
But I guess it's ok since there is also a busy() check in there.

It's not what I would have done though. This just drops the sound, which means that users can end up changing the volume several steps but only get one sound.  I would have just called fadeout(0.2) to get rid of the previous sound and then played the new sound instead.  However, if you are happy with how this works, that's fine with me.

props/saber_BC_buttons.h
profezzorn: If we're not always playing immediately, then maybe the comment shouldn't say "Always play immediately" ?

props/saber_BC_buttons.h
NoSloppy: Calling fadeout is probably the right and more simple way to handle most of this. I guess I couldn't figure out how to target the sound that's playing? I'll have to look again.

I will play around when I get back.

PR #822, opened 2025-06-02
Fix scanid for serial devices

NoSloppy:


props/prop_base.h
```diff
@@ -545,17 +545,17 @@ class PropBase : CommandParser, protected Looper, protected SaberBase, public Mo
...
     SpeakBladeID(ret);
 #endif // SPEAK_BLADE_ID
     }
 #ifdef BLADE_DETECT_PIN
     if (!blade_detected_) {
-      PVLOG_STATUS << "NO ";
+      STDOUT << "NO ";
       ret += NO_BLADE;
     }
-    PVLOG_STATUS << "Blade Detected\n";
```
profezzorn, line 558: Isn't it enough to just change line 548? (And add a comment that specifies why it needs to be STDOUT.)

NoSloppy: Sure, done.

PR #823, opened 2025-06-06
HEV - Expanded Clash function

jimsworld:
Custom behavior has been added to the Clash function in hev.h.
Strength still applies to the intensity of the clash sound.
Laid some groundwork to be picked up by a future HEVQueue class which will handle the chance, order and queuing of which follow-up voice lines might play.

Please read the notes inside the file for full details on the custom behavior.
(the top section will prettier eventually).


profezzorn:
Close for now...

PR #831, opened 2025-07-04
Allow idle to fade in

NoSloppy:
This is to handle a one time fade-in of the looping idle sound when it begins, without having a fade directly in the idle sound (and hearing a fade-in on each restart) nor having to bake in a manual "bgn-idle" fade to the end of boot, font, in, or postoff.

This is based on my original proposal for idle, where a boolean gets flipped for one cycle in loop() to trigger idle to start  playing.
We ended up (the current way) leveraging file linking in effect.h to have it follow boot and font, and used SetFollowing() in SB_Off so that in.wav and preon could finish before it auto-started playing.
That does work, but to make it fade in, we need to start it manually with a fade time set, and I didn't see a way to intercept that linking routine cleanly just for this small "feature".
This does not address adding bgnidle yet. If that becomes a thing, we can check if it exists, and bypass the fade-in part.
I added a font.config parameter so users can edit the fade time, and also kept the AvoidIdleSDAccess() check.

Lastly, in volume_overlay, the fade speed was clamped to a slowest speed of 1 unit/sample. 
I added a fractional stepping to allow for longer fade times, and I tested it with various times, some as long as 10 seconds. It uses a quadratic ease-in curve on each sample, because trying basic linear sounded too sudden to my ear.

It's working well in my opinion. I should be low on additional cycles/memory consumption, and safe from integer truncation by allowing a minimum of 10ms.

If this can be done better, by all means let me know. If it's horrible, also please let me know. :)

P.S. I also added the parameter (and a few other OS8.x ones) to the POD for review
https://github.com/profezzorn/ProffieOSDocs/pull/70

### DIFF
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -712,14 +712,12 @@ EFFECT(preon);
 EFFECT(pstoff);
 
 // Monophonic fonts
+// Idle effect, plays when blade is off.
 #ifdef ENABLE_IDLE_SOUND
 EFFECT2(idle, idle);
-EFFECT2(boot, idle);
-EFFECT2(font, idle);
-#else
+#endif
 EFFECT(boot);
 EFFECT(font);      // also polyphonic
-#endif
 EFFECT(bladein);   // also polyphonic
 EFFECT(bladeout);  // also polyphonic
 EFFECT2(hum, hum);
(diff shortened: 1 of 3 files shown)


profezzorn:
I think I like the linking version better, we just have to add a bgnidle in the chain of links.

NoSloppy:
Close in lieu of just using bgnidle instead.

PR #832, opened 2025-07-06
Add DEFAULT ARRAY define

Sabersense-CC:
Useful for busy installers as you can have a standard array for all common blade lengths, then just define which one you want as the default.

### DIFF
--- a/props/saber_sabersense_buttons.h
+++ b/props/saber_sabersense_buttons.h
@@ -1,4 +1,4 @@
-/* V7/8-261.
+/* V7/8-262.
 ============================================================
 =================   SABERSENSE PROP FILE   =================
 =================            by            =================
@@ -238,6 +238,22 @@ COLOUR CHANGE FUNCTIONS WITH BLADE ON
       { 3, ... }
       etc.
 
+#define SABERSENSE_DEFAULT_BLADE_ARRAY 3
+  This feature is really intended for busy installers
+  and sets the default blade array in multi-array systems.
+  Using this feature, you can have a blade array with all
+  the common blade lengths included, and then select which
+  length to default to on first boot (i.e. until a save
+  file is written).
+  The alternative is to simply re-order the blade array
+  manually, but then you have to re-order the array1.wav,
+  array2.wav etc. files on the SD card to match, which
+  is a hassle.
+  Note that the define uses zero-based numbering, the
+  same as the blade array itself, so you must use the
+  number as it is shown in the blade array - i.e. number 3
+  for the fourth array down the full list.
+
 #define SABERSENSE_DISABLE_SAVE_ARRAY
   By default, SABERSENSE_ARRAY_SELECTOR saves the current
   array so that the saber will always boot into the last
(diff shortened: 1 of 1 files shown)


props/saber_sabersense_buttons.h
```diff
@@ -335,6 +363,14 @@ class SaveArrayStateFile : public ConfigFile {
     struct SabersenseArraySelector {
       static int return_value;  // Tracks current array index.
       float id() {
+#ifdef BLADE_DETECT_PIN
+        if (return_value == 0) {
```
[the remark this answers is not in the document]
Sabersense-CC: On reading again, I think I see where you're coming from - how can it be zero when the system will add one under Blade Detect? In which case, I concede it's unlikely, but it is still possible because if the user defines zero, that will override the +1, leaving the define invalid.
So yes, this check is guarding against a rare edge case, but since it adds negligible overhead, I figured it was worth having for the sake of robustness.

Sabersense-CC:
Just spotted a bug with disabling array saving, as the default array defines had been placed in the wrong place relative to the disable define. Now fixed.

props/saber_sabersense_buttons.h
```diff
@@ -322,19 +338,46 @@ GESTURE CONTROLS
...
+    SABERSENSE_DEFAULT_BLADE_ARRAY != 0,
+    "[Sabersense] ERROR: "
+    "SABERSENSE_DEFAULT_BLADE_ARRAY must be 1 or higher when using Blade Detect."
+  );
+#endif
+
     struct SabersenseArraySelector {
       static int return_value;  // Tracks current array index.
       float id() {
+        if (return_value >= NELEM(blades)) {
```
profezzorn, line 378: How can this ever happen?
At least print out a warning or something, otherwise you're just hiding bugs potentially.

Sabersense-CC: I admit it's a little bit belt and braces, but it protects against someone messing with the arrysave file on the SD card and changing the array to an invalid one. Or perhaps someone reflashes with fewer arrays than before, but then the system tries to use a saved array which is now invalid (as reflashes don't delete save files).
Anyway the system will now auto-correct by reverting to the default array, and I've added a serial alert telling the connected user what's happened and why.
I've also just tweaked layout and order a bit for clarity and readability.

Sabersense-CC:
It's just occurred to me, the system checking whether the save file is valid and reverting to default array if it isn't, should only happen once, or at least until the next time the user switches array, at which point the system will update the save array file with a valid value.

Unless, I guess, the problem is caused by genuine file corruption that isn't the result of user error, but just a system glitch. But is that an edge case we need to specifically cover? I don't think we protect our other system save files like this.

Maybe have the system always delete the save file when reflashing to prevent a previously valid value persisiting when uploading a config with fewer arrays? Or again, is that an edge case too far?

Or maybe another way to handle this would be to have the system delete the array save file on bootup if it detects that the array value it contains is invalid. Serial monitor could still tell power users what's happened and why, but it means that thereafter, the array save file would be valid, because the system would create a new, legit one the next time the user switched arrays - or, if they never switched arrays, there would be no save file at all, rather than a lingering invalid one.

Or does that make debugging trickier? Don't forget, my prop allows people to delete all their save files just with a button press - they don't even need to remove the SD card - so maybe an auto-delete is indeed edge-case overreach.

Indeed am I now in the realms of massively overthinking this? Is the previous fix of simply falling back to the system default array and just mentioning it on Serial Monitor still the best way forward - especially as the next array change by the user should, in theory, repair the invalid save file by updating it with a valid value (system file corruption excepted maybe). I mean, I guess there is a limit to how much protection we should offer against edge cases like this - and maybe I already crossed that threshold a while ago in this instance?

Basically I'm happy to go with whatever your preference is Prof, as I think any of these are pretty easy to implement and won't carry much, if any, overhead.

PR #833, opened 2025-07-08
Add bgnidle

NoSloppy:
Needs help.
Doesn't account for FastOn() from prop_base.
bgnidle->idle plays when it shouldn't.

### DIFF
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -717,9 +717,12 @@ EFFECT(pstoff);
 #ifdef ENABLE_IDLE_SOUND
 EFFECT2(idle, idle);
 EFFECT2(bgnidle, idle);
-#endif
+EFFECT2(boot, bgnidle);
+EFFECT2(font, bgnidle);
+#else
 EFFECT(boot);
-EFFECT(font);
+EFFECT(font);      // also polyphonic
+#endif
 EFFECT(bladein);   // also polyphonic
 EFFECT(bladeout);  // also polyphonic
 EFFECT2(hum, hum);
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -714,6 +714,7 @@ EFFECT(pstoff);
 // Monophonic fonts
 #ifdef ENABLE_IDLE_SOUND
 EFFECT2(idle, idle);
+EFFECT2(bgnidle, idle);
 EFFECT2(boot, idle);
 EFFECT2(font, idle);
 #else
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -712,15 +712,14 @@ EFFECT(preon);
 EFFECT(pstoff);
 
 // Monophonic fonts
+
+// Idle effect, plays when blade is off.
 #ifdef ENABLE_IDLE_SOUND
 EFFECT2(idle, idle);
 EFFECT2(bgnidle, idle);
-EFFECT2(boot, idle);
-EFFECT2(font, idle);
-#else
-EFFECT(boot);
-EFFECT(font);      // also polyphonic
 #endif
+EFFECT(boot);
+EFFECT(font);
 EFFECT(bladein);   // also polyphonic
 EFFECT(bladeout);  // also polyphonic
 EFFECT2(hum, hum);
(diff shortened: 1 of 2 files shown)


sound/hybrid_font.h
```diff
@@ -484,12 +484,12 @@ class HybridFont : public SaberBase, public Looper {
     bool most_blades = location.on_blade(0);
 #ifdef ENABLE_IDLE_SOUND
     if (most_blades) {
-      Effect* idle = AvoidIdleSDAccess() ? nullptr : &SFX_idle;
+      Effect* bgnidle = AvoidIdleSDAccess() ? nullptr : &SFX_bgnidle;
       if (SFX_pstoff) {
         SFX_in.SetFollowing(&SFX_pstoff);
-        SFX_pstoff.SetFollowing(idle);
+        SFX_pstoff.SetFollowing(SFX_bgnidle ? bgnidle : &SFX_idle);
```
profezzorn, line 490: If I remember correctly, I fixed it so that if a "following" effect doesn't exist, it automatically goes to the next one. So it should be fine to just always link pstoff to bgnidle.

NoSloppy: Cool, yes it does work.
I also had to add boot handling and add bgnidle to StopIdleSound().

sound/hybrid_font.h
```diff
@@ -862,12 +880,18 @@ class HybridFont : public SaberBase, public Looper {
   void StopIdleSound() {
 #ifdef ENABLE_IDLE_SOUND
     RefPtr<BufferedWavPlayer> idlePlayer = GetWavPlayerPlaying(&SFX_idle);
+    RefPtr<BufferedWavPlayer> bgnIdlePlayer = GetWavPlayerPlaying(&SFX_bgnidle);
```
profezzorn, line 883: We don't need to stop both players, because there is really only one, so we can simplify as:

```
if (!idlePlayer) idlePlayer = GetWavPlayerPlaying(&SFX_bgnidle);
```

There is another problem here, which is that it's possible that bgnidle ends and idle starts right in between these two statements.  It might be enough to just swap them around, like:

```
    // We avoid a race condition by checking for bgnidle before idle.
    RefPtr<BufferedWavPlayer> idlePlayer = GetWavPlayerPlaying(&SFX_bgnidle);
    if (!idlePlayer) idlePlayer = GetWavPlayerPlaying(&SFX_idle);
```

sound/hybrid_font.h
```diff
@@ -862,12 +880,18 @@ class HybridFont : public SaberBase, public Looper {
...
       idlePlayer->FadeAndStop();
       idlePlayer.Free();
-      PVLOG_NORMAL << "**** Stopped idle wav\n";
       }
+    if (bgnIdlePlayer) {
+      bgnIdlePlayer->set_fade_time(0.5);
+      bgnIdlePlayer->FadeAndStop();
+      bgnIdlePlayer.Free();
+    }
+      PVLOG_NORMAL << "**** Stopped idle wav\n";
```
profezzorn, line 894: This should be a higher level of logging.

sound/effect.h
```diff
@@ -712,14 +712,14 @@ EFFECT(preon);
...
 #ifdef ENABLE_IDLE_SOUND
 EFFECT2(idle, idle);
-EFFECT2(boot, idle);
-EFFECT2(font, idle);
-#else
-EFFECT(boot);
-EFFECT(font);      // also polyphonic
+EFFECT2(bgnidle, idle);
 #endif
+EFFECT(boot);
```
profezzorn, line 721: Why can't we have these linked to bgnidle?

sound/hybrid_font.h
```diff
@@ -507,7 +507,11 @@ class HybridFont : public SaberBase, public Looper {
         StopIdleSound();
         break;
       case OFF_FAST:
+#ifdef ENABLE_IDLE_SOUND
+        SFX_in.SetFollowing(&SFX_bgnidle);
```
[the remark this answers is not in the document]
NoSloppy: No because the point of fast off is it bypasses postoff.

sound/effect.h
NoSloppy: Well before I know that a missing bgn sound would auto-skip to the base sound, this was needed. Reverted now because it works fine if no bgnidle sound exists.
It's great.

PR #834, opened 2025-07-08
Add fast preset switching to serial commands

NoSloppy:
The way it is before this patch, led to some confusion in testing the bgnidle stuff.
You have to physically do the button event on the saber to get a next/prev_preset_fast.
I was using Serial Monitor, so I was getting "font.wav" playing, which caused a series of unnecessary coding of a workaround, for an issue that didn't exist. (see https://github.com/profezzorn/ProffieOS/pull/833/commits/b2c351b71ab12a3334e431f6824c4d6a43167c27)
Also adding in first_preset jump as it's a convenient shortcut.
That part can go away if it's not approved. I can always just use it on my end.

### DIFF
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1578,10 +1578,7 @@ public:
       SaberBase::IsOn() ? previous_preset_fast() : previous_preset();
       return true;
     }
-    if (!strcmp(cmd, "f") || (!strcmp(cmd, "first") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
-      SaberBase::IsOn() ? first_preset_fast() : first_preset();
-      return true;
-    }
+
     if (!strcmp(cmd, "rotate")) {
       rotate_presets();
       return true;
--- a/props/prop_base.h
+++ b/props/prop_base.h
@@ -1571,11 +1571,15 @@ public:
       return true;
     }
     if (!strcmp(cmd, "n") || (!strcmp(cmd, "next") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
-      next_preset();
+      SaberBase::IsOn() ? next_preset_fast() : next_preset();
       return true;
     }
     if (!strcmp(cmd, "p") || (!strcmp(cmd, "prev") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
-      previous_preset();
+      SaberBase::IsOn() ? previous_preset_fast() : previous_preset();
+      return true;
+    }
+    if (!strcmp(cmd, "f") || (!strcmp(cmd, "first") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
+      SaberBase::IsOn() ? first_preset_fast() : first_preset();
       return true;
     }
     if (!strcmp(cmd, "rotate")) {


props/prop_base.h
```diff
@@ -1571,11 +1571,15 @@ class PropBase : CommandParser, protected Looper, protected SaberBase, public Mo
...
-      next_preset();
+      SaberBase::IsOn() ? next_preset_fast() : next_preset();
       return true;
     }
     if (!strcmp(cmd, "p") || (!strcmp(cmd, "prev") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
-      previous_preset();
+      SaberBase::IsOn() ? previous_preset_fast() : previous_preset();
+      return true;
+    }
+    if (!strcmp(cmd, "f") || (!strcmp(cmd, "first") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
```
profezzorn, line 1581: It feels like "first preset" already exists as "set_preset 0".

props/prop_base.h
```diff
@@ -1571,11 +1571,15 @@ class PropBase : CommandParser, protected Looper, protected SaberBase, public Mo
       return true;
     }
     if (!strcmp(cmd, "n") || (!strcmp(cmd, "next") && arg && (!strcmp(arg, "preset") || !strcmp(arg, "pre")))) {
-      next_preset();
+      SaberBase::IsOn() ? next_preset_fast() : next_preset();
```
profezzorn, line 1574: I'm trying to decide if I like this.
I mean, wouldn't you still want to hear the font sound when switching presets when it's on?

NoSloppy: We currently do not play font.wav when fast switching presets. 
I do not remember if that was ever a discussion, but personally I think that's appropriate. 
Announcing the font is a "when off" thing in my mind like a volume or blade length utility.
Quotes are available for any "while on" dialogue.

Anyway, this change is nothing more than aligning the behavior of driving the board with serial commands to how it actually works on saber.

props/prop_base.h
NoSloppy: Fair. I forgot about set_preset.
Removed.

props/prop_base.h
profezzorn: It makes sense that next_preset_fast / prev_preset_fast doesn't play the font name, i didn't mean that should change.  I'm just wondering if calling the "fast" functions when on is actually the right thing to do?

NoSloppy: If the real saber is on, and you do next preset, it does the fast version. Unless the prop doesn't support that, then I guess it does nothing? I might be guilty of assuming the behavior, and that's what I based the suggestion on.
I run my prop, and it does fast when saber is on, so for *me* it's mirroring the real world saber.
It's totally fine if this goes away, I just thought it was an alignment.

profezzorn: > If the real saber is on, and you do next preset, it does the fast version.

Does it? That's not a given to me. 
Does all the props do this? If so, then I guess this makes sense.
The "saber" prop doesn't really have a way to change preset while on.

I'm asking because I really don't know what is the right behavior.
Calling two different functions costs a little extra memory, but probably not enough to matter.

NoSloppy: > Does all the props do this?

No, apparently only BC and FETT263 props do.
It's fine. I'll withdraw this.

PR #836, opened 2025-07-19
Add Two-Way Array Cycling.

Sabersense-CC:
This has been touched on before, but always abandoned because it requires a define telling the system how many arrays are present. So this time I've wrapped the feature in a define so that it's there if people want it but doesn't place extra demands if they don't, with the define itself specifying the number of arrays. I've also included a couple of safeguards to prevent invalid NUM_ARRAY values.

The only thing not covered is if people define fewer arrays than are present. But that's a fairly simple thing to trace in the config once they notice that behaviour does not match expectation. 
_*** EDIT: I've now tweaked the alert to include low as well as high values. ***_

I've also tweaked the UP/DOWN angles to make them less strict, and then changed the single-preset UP/DOWN forward/backward parameter to match.

Lastly Two-Way array cycling meant NextBladeArray was no longer accurate as it would apply to previous blade arrays too. Since all it really does is play the array idents, I changed it to be a more accurately descriptive IdentBladeArray.

Hope this all OK, and sorry to keep bombarding you with PRs - at least there is only a relatively small amount of code involved!

Thanks as always.
:)

### DIFF
--- a/props/saber_sabersense_buttons.h
+++ b/props/saber_sabersense_buttons.h
@@ -346,7 +346,7 @@ GESTURE CONTROLS
 #endif
 #endif
 
-  // Check that Default Array is valid.
+  // Check that Default Array is valid at compile time.
   static_assert(
     SABERSENSE_DEFAULT_BLADE_ARRAY < NELEM(blades),
     "SABERSENSE_DEFAULT_BLADE_ARRAY must reference a valid array (note zero-based counting)."
--- a/props/saber_sabersense_buttons.h
+++ b/props/saber_sabersense_buttons.h
@@ -1,4 +1,4 @@
-/* V7/8-276.
+/* V7/8-277.
 ============================================================
 =================   SABERSENSE PROP FILE   =================
 =================            by            =================
@@ -346,7 +346,7 @@ GESTURE CONTROLS
 #endif
 #endif
 
-  //  Check user-defined array is valid at compile time.
+  // Check user-defined array is valid at compile time.
   static_assert(
     SABERSENSE_DEFAULT_BLADE_ARRAY < NELEM(blades),
     "SABERSENSE_DEFAULT_BLADE_ARRAY must reference a valid array (note zero-based counting)."
@@ -357,13 +357,12 @@ GESTURE CONTROLS
     SABERSENSE_DEFAULT_BLADE_ARRAY != 0,
     "[Sabersense] ERROR: Default array index must be 1 or higher when using Blade Detect."
   );
-  
   constexpr int min_blades_required = 3;
 #else
   constexpr int min_blades_required = 2;
 #endif
 
-  // Check that Array Selector is needed (prevents unexpected behaviour).
+  // Check that Array Selector is actually needed (prevents possible unexpected behaviour).
   static_assert(
     NELEM(blades) >= min_blades_required,
     "[Sabersense] ERROR: Array Selector is not required with only one selectable blade array. "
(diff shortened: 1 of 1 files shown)


props/saber_sabersense_buttons.h
```diff
@@ -238,6 +238,12 @@ COLOUR CHANGE FUNCTIONS WITH BLADE ON
       { 3, ... }
       etc.
 
+#define SABERSENSE_NUM_ARRAYS_TWO_WAY_SELECTOR 6
+  Allows two-way array selection, i.e. pointing up moves
+  to next array, pointing down moves to previous array.
+  Number denotes the total number of arrays present
```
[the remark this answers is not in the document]
Sabersense-CC: Well... yes, exactly... that's what I thought!
When I first tried to build the Array Selector, I tried to make it two-way, but ran into problems with reverse cycling. Whatever approach I took, I always found myself at the same impasse of needing to define the number of arrays - which is why I abandoned the idea until now. But I can't for the life of me remember the details of the problem.
However I've just re-worked it and it does indeed seem to work correctly now, just using NELEM(blades). Maybe I had a slightly different code structure before that was stopping it working, but with the latest version it seems OK, so I've added it and made it the default - no need for any extra defines, it's now just part of Array Selector.
The only question is the third static alert -should we tell the user to undefine it, or just quietly undefine it for them?

props/saber_sabersense_buttons.h
```diff
@@ -349,47 +351,66 @@ GESTURE CONTROLS
...
+
+    static void cycle_directional(bool forward) {
 #ifndef BLADE_DETECT_PIN
-        return_value = (return_value + 1) % NELEM(blades);
-#else   // Improved logic fixes early Array1 repetition when using Blade Detect.
-        return_value = 1 + return_value % (NELEM(blades) - 1);
-#endif
+      const int size = NELEM(blades);
+      return_value = (return_value + (forward ? 1 : size - 1)) % size;
+#else
```
profezzorn, line 399: The if and else do very similar things here. It seems like they should be able to share more code.
Maybe something like this:
```cpp
   int size = NELEM(blades);
   int offset = 0;
#ifdef BLADE_DETECT_PIN
   offset = 1;
   size--;
#endif
   return_value += forward ? 1 : (size - 1);
   return_value = (return_value + size - offset) % size + offset;
```

props/saber_sabersense_buttons.h
```diff
@@ -720,7 +741,7 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
   }
 #endif
 
-  void NextBladeArray() {
+  void IdentBladeArray() {
```
profezzorn, line 744: This seems like a step down in clarity.
What does "identBladeArray" mean? What does this function do now?

props/saber_sabersense_buttons.h
```diff
@@ -349,47 +351,66 @@ GESTURE CONTROLS
...
+      if (return_value < 0 || return_value >= NELEM(blades)) {
+        Serial.println("[Sabersense] ALERT: User or externally-specified array index invalid. "
+                       "Resetting to default.");
+        return_value = SABERSENSE_DEFAULT_BLADE_ARRAY;
       }
-      static void cycle() {
+      return return_value;
+    }
+
+    static void cycle_directional(bool forward) {
```
profezzorn, line 395: I think the "directional" is redundant when you have a "forward" argument.
"cycle" or "step" seems like it would be enough.

Sabersense-CC: Now changed to cycle_array.

props/saber_sabersense_buttons.h
```diff
@@ -720,7 +735,7 @@ class SabersenseButtons : public PROP_INHERIT_PREFIX PropBase {
   }
 #endif
 
-  void NextBladeArray() {
+  void IdentBladeArray() {  // Plays associated arrayx.wav to ident array landed on.
```
profezzorn, line 738: If this is what it does, how about calling it PlayArraySound() ?
Seems like if we name it well, no comment is actually required.

PR #837, opened 2025-07-24
Add EFFECT_USERx functionality to Sabersense prop...

Sabersense-CC:
...for specific blade style requirements.
It's basically a paired down, cherrypicked and modified version of Fett263's 'Special Abilities'.

### DIFF
--- a/props/saber_sabersense_buttons.h
+++ b/props/saber_sabersense_buttons.h
@@ -322,7 +322,7 @@ COLOUR CHANGE FUNCTIONS WITH BLADE ON
   being triggered by clash while holding aux.
 
 #define SABERSENSE_F263_CUSTOM_USER_EFFECT
-  Based of Fett263 'Special Abilities', this define
+  Based on Fett263 'Special Abilities', this define
   enables interaction with EFFECT_USER1 etc. in blade
   styles for things like swing colour changing.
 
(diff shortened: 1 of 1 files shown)


AusBULLy (commented):
Lines 883 and 884 are the same or is 884 supposed to be the START of the next IF statement?

fdarosa2663:
OK, missed that part. I was more thinking if users try to use the Special Abilities from my library many will not work properly without many of the other functions and effects in my prop.  EFFECT_USERX is just a simple trigger, so only a few of the actual Special Abilities in the library would run, most others use other functions on top of the trigger to operate.

When you release you probably want to clarify that many from the library will not work to avoid confusion.

I appreciate the mention, I can't say I'm fully up on the courtesy conventions myself, but I also usually include the user's name when I'm adding functionality they developed, yourself included, it seems like the best approach

props/saber_sabersense_buttons.h
```diff
@@ -843,6 +849,41 @@ bool Event2(enum BUTTON button, EVENT event, uint32_t modifiers) override {
...
+
+    case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_ON | BUTTON_AUX):
+      SaberBase::DoEffect(EFFECT_USER1, 0);
+      return true;
+
+    case EVENTID(BUTTON_NONE, EVENT_SWING, MODE_OFF | BUTTON_AUX):
+      SaberBase::DoEffect(EFFECT_USER2, 0);
+      return true;
+#endif
+#endif
```
profezzorn, line 885: Make it obvious what this #endif is for like this:

```cpp
#endif  // SABERSENSE_F263_CUSTOM_USER_EFFECT
```

PR #840, opened 2025-08-01
bug fix for 1 button Edit Settings Menu

fdarosa2663:
User reported bug in theCrucible, causes uses to be unable to exit when EDIT_SETTINGS_MENU define used.

### DIFF
--- a/props/saber_fett263_buttons.h
+++ b/props/saber_fett263_buttons.h
@@ -5421,7 +5421,11 @@ SaberFett263Buttons() : PropBase() {}
             sound_library_.SayExit();
             menu_ = false;
           } else {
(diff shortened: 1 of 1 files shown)


fdarosa2663:
thanks, good catch

profezzorn:
Why is this commit against the 7.x branch?
Normally, all changes should go to the master branch, and then be cherry-picked from there to other branches.

fdarosa2663:
Sorry, this bug is present in OS7, since OS8 is still in beta I thought I needed to submit here.
I did same update to master in other PR.

fdarosa2663:
will submit one pr to master to avoid confusion

PR #842, opened 2025-08-01
EDIT_SETTINGS_MENU bug fix for OS7 & 8

fdarosa2663:
Users get stuck in wrong menu using the "Cancel" control, was reported on theCrucible, hadn't been noticed or reported prior but should probably be applied to OS7 as a bug fix on top on including in OS8.


profezzorn:
How important / urgent is it to get this into OS7 ?

PR #844, opened 2025-08-16
Add base color error msgs

NoSloppy:
Just compile time error messages that are user friendly.

### DIFF
--- a/styles/layers.h
+++ b/styles/layers.h
@@ -11,15 +11,19 @@
 // BASE: COLOR or LAYER
 // LAYER1, LAYER2: LAYER
 // return value: COLOR or LAYER (same as BASE)
-// This style works like layers in Gimp or Photoshop: each layer adds to the stack covering the base.
-// The written order of layers in the blade style code puts the base first, with any following layers visually covering it.
+// This style works like layers in Gimp or Photoshop: each layer adds to the stack covering the BASE.
+// The written order of layers in the blade style code puts the BASE first,
+// with any following layers visually covering it. (later layers draw over earlier ones)
+// The rule for opacity is as follows:
+//   The output of Layers<> is opaque if the BASE of that Layers<> is opaque.
+//   If BASE is transparent: the overall result is transparent.
+//   No additional solid opaque colors may be stacked on top of an opaque BASE.
 //
-// When a Layers<> is used as the first template argument to StylePtr<>, its BASE color (the first layer)
-// must be a solid (opaque) color, not transparent. No additional solid colors may be stacked on top
-// of that solid base.
-// Additional layers stacked on top of a solid base color are transparent,
-// so the base color shows through until an effect wants to paint over it (e.g., a clash).
-// Layers<> can be nested inside other Layers<> and can be used anywhere a STYLE or COLOR is expected.
+// StylePtr<> expects its argument to be opaque.
+// Therefore, when using StylePtr<Layers<BASE, ... >>(), BASE MUST be an opaque color.
+//
+// Layers<> itself can still be nested anywhere a STYLE/COLOR is allowed; only the top-level StylePtr<...>()
+// requirement forces the requirement of the BASE to be opaque in that specific position.
 
 template<class BASE, class L1>
 class Compose {


styles/layers.h
```diff
@@ -73,12 +82,35 @@ class Compose<BLACK, L1> {
...
-template<class BASE, class L1, class ... REST> struct LayerSelector<BASE, L1, REST...> {
+
+// If base is exactly AlphaL<*, Int<0>> and there are 3+ args, drop it early
+template<class BASE, class L1, class L2, class... REST>
+struct LayerSelector<AlphaL<BASE, Int<0>>, L1, L2, REST...> {
+  typedef typename LayerSelector<L1, L2, REST...>::type type;
+};
+
+// Solid+solid check
+template<class BASE, class L1>
```
profezzorn, line 96: Isn't it true that no layer except the first one should be solid?

styles/layers.h
```diff
@@ -4,6 +4,15 @@
 #include "alpha.h"
 #include "../functions/int.h"
 #include "colors.h"
+#include <type_traits>
+#include <utility>
+
+// Identify “solid” colors by the return type of getColor()
+namespace layers_detail {
```
profezzorn, line 11: This seems like it could be useful in lots of places, so maybe call it color_details and put it in common/color.h instead?

styles/style_ptr.h
```diff
@@ -3,6 +3,17 @@
 
 #include "blade_style.h"
 
+template<class BASE, class L1> class Compose;  // forward-declare Layers node
+
+namespace style_base_check_detail {
+  template<class T> struct TopBase { using type = T; };
+  template<class B, class L> struct TopBase<Compose<B,L>> : TopBase<B> {};
```
profezzorn, line 10: What's the point of checking the base layer rather than the whole stack? The result should be the same.

styles/style_ptr.h
```diff
@@ -3,6 +3,17 @@
...
+
+  template<typename T> struct IsOpaqueColor : std::false_type {};
```
profezzorn, line 12: This is the same code as in layers_details, so it should definitely go in common/color.h

styles/layers.h
NoSloppy: hmm. would "“Reject solid layers after the first is found" be a better comment?
This would go away anyway with a move to color.h

NoSloppy: Comment updated.

styles/layers.h
NoSloppy: Done.

styles/style_ptr.h
NoSloppy: We only care about the first layer, right? (for this error since that's what causes a compile fail?)
It must be solid (or Int<0>). Later solids don’t fix a transparent base, so the check enforces that rule.

profezzorn: What we care about is that the result is solid.
Digging into why it's solid seems like extra work.

styles/layers.h
profezzorn: I'm saying that we don't need to make so many changes, we can just insert the assert into Compose<>
Basically, if the second argument to Compose is opaque -> error

styles/style_ptr.h
```diff
@@ -164,6 +183,18 @@ StyleAllocator StylePtr(const char* default_arguments) {
 // that you can't turn it on/off, and the battery low warning is disabled.
 template<class STYLE>
 StyleAllocator ChargingStylePtr() {
+  using _TopBaseT  = typename style_base_check_detail::TopBase<STYLE>::type;
+  using _TopBaseCol = decltype(std::declval<_TopBaseT&>().getColor(0));
+  static_assert(color_details::IsOpaqueColor<_TopBaseCol>::value,
```
profezzorn, line 188: These error messages are overkill.
A few stars at the beginning/end should be enough.

NoSloppy: yeah - calmed them down

styles/style_ptr.h
NoSloppy: The result-only check hides the error from color_details. It only inspects the final composite type, so if the base is transparent but a later solid makes the result opaque, IsOpaqueColor says “ok” and our “must be solid” message never fires.

That’s why I kept the base check: it flags the illegal transparent base immediately, at the point of mistake.

styles/layers.h
NoSloppy: done

styles/layers.h
```diff
@@ -4,21 +4,31 @@
...
+// so the base color shows through until an effect wants to paint over it (e.g., a clash).
+// Layers<> can be nested inside other Layers<> and can be used anywhere a STYLE or COLOR is expected.
+
 
 template<class BASE, class L1>
 class Compose {
 public:
+  using _BaseColorT  = decltype(std::declval<BASE&>().getColor(0));
+  using _LayerColorT = decltype(std::declval<L1&>().getColor(0));
+  static_assert(!(color_details::IsOpaqueColor<_BaseColorT>::value &&
```
profezzorn: I think it's enough to just check if _LayerColorT is opaque.
It doesn't matter if _BaseColorT is opaque or not.

styles/style_ptr.h
profezzorn: "a later solid makes the result opaque"

I'm suggesting that we make this a separate compilation error (in layers.h)

styles/layers.h
```diff
@@ -4,21 +4,28 @@
...
+// so the base color shows through until an effect wants to paint over it (e.g., a clash).
+// Layers<> can be nested inside other Layers<> and can be used anywhere a STYLE or COLOR is expected.
+
 
 template<class BASE, class L1>
 class Compose {
 public:
+static_assert(!color_details::IsOpaqueColor<decltype(std::declval<L1&>().getColor(0))>::value,
+                "\n\n"
+                "*** Layers<> ERROR: CANNOT STACK TWO SOLID COLORS.\n\n");
```
profezzorn, line 28: This error needs updating now.
Something like "Layers<> error: Only the base color may be solid" maybe?

styles/layers.h
```diff
@@ -4,21 +4,28 @@
...
+// of that solid base.
+// Additional layers stacked on top of a solid base color are transparent,
+// so the base color shows through until an effect wants to paint over it (e.g., a clash).
+// Layers<> can be nested inside other Layers<> and can be used anywhere a STYLE or COLOR is expected.
+
 
 template<class BASE, class L1>
 class Compose {
 public:
+static_assert(!color_details::IsOpaqueColor<decltype(std::declval<L1&>().getColor(0))>::value,
```
profezzorn, line 26: Should be indented two more steps.

styles/layers.h
```diff
@@ -73,12 +80,21 @@ class Compose<BLACK, L1> {
...
 };
 
-
 template<class BASE, class ... LAYERS> struct LayerSelector {};
 template<class BASE> struct LayerSelector<BASE> { typedef BASE type; };
+
+// Drop fully transparent first item (Alpha 0) when there are exactly 2 args.
 template<class BASE, class L1> struct LayerSelector<AlphaL<BASE, Int<0>>, L1> { typedef L1 type; };
+
+// Drop fully transparent first item (Alpha 0) when there are 3+ args.
```
profezzorn, line 89: I think the correct fix for this is to change line 82 from:

```
  typedef typename LayerSelector<Compose<BASE, L1>, REST...>::type type;
```

to

```
  typedef typename LayerSelector<LayerSelector<BASE, L1>::type, REST...>::type type;
```

I think that's what I originally intended anyways.
That way everything gets broken down into two-layer calls to LayerSelector, and we don't have to special-case three-layer stuff.

styles/layers.h
NoSloppy: Fixed.

styles/layers.h
NoSloppy: oops. actually deleted this time.

styles/style_ptr.h
```diff
@@ -133,14 +140,20 @@ class ChargingStyle : public Style<T> {
 // Get a pointer to class.
 template<class STYLE>
 StyleAllocator StylePtr() {
+  using _TopBaseT  = typename style_base_check_detail::TopBase<STYLE>::type;
```
profezzorn, line 143: Now that Layers<> can't become magically opaque anymore, we can remove this, right?
Edit:.. see below first

styles/style_ptr.h
```diff
@@ -133,14 +140,20 @@ class ChargingStyle : public Style<T> {
...
+  using _TopBaseCol = decltype(std::declval<_TopBaseT&>().getColor(0));
+  static_assert(color_details::IsOpaqueColor<_TopBaseCol>::value,
+                "\n\n"
+                "*** StylePtr<> error: BASE LAYER MUST BE A SOLID COLOR, NOT TRANSPARENT.\n"
```
profezzorn, line 147: This might be confusing if the argument to styleptr is not a layer.
How about we move the actual assert into style_base_check_detail, and we'll make one error message if the argument is Layers<> with a transparent base, and another error message if it's something else that is transparent?

That should also give us a little bit more code reuse since this is currently repeated in StyleChargingPtr<>

styles/layers.h
```diff
@@ -4,21 +4,28 @@
...
 // BASE: COLOR or LAYER
 // LAYER1, LAYER2: LAYER
 // return value: COLOR or LAYER (same as BASE)
-// This style works like layers in gimp or photoshop.
-// In most cases, the layers are expected to be normally transparent effects
-// that turn opaque when then want to paint an effect over the base color.
-// If the base color is opqaque, the final result of this style will also be
-// opaque. If the base color is transparent, the final result may also be transparent,
-// depending on what the layers paint on top of the base color.
+// This style works like layers in Photoshop: each layer is painted over the one beneath it.
```
profezzorn, line 14: "beneath it" can be confusing, does it refer to the order of rending or the order of writing?
Not sure what a better way to describe it would be though...

styles/style_ptr.h
NoSloppy: > another error message if it's something else that is transparent

hmm, meaning something like
`AudioFlicker<AlphaL<Blue,Int<16000>>,AlphaL<Red,Int<16000>>>` ?

How about we just leave the single message but remove "layer" from it so it just says "base must" - that covers a transparent style or a Layer.
"StylePtr<> error: BASE MUST BE A SOLID COLOR, NOT TRANSPARENT

styles/layers.h
```diff
@@ -4,21 +4,29 @@
...
 // BASE: COLOR or LAYER
 // LAYER1, LAYER2: LAYER
 // return value: COLOR or LAYER (same as BASE)
-// This style works like layers in gimp or photoshop.
-// In most cases, the layers are expected to be normally transparent effects
-// that turn opaque when then want to paint an effect over the base color.
-// If the base color is opqaque, the final result of this style will also be
-// opaque. If the base color is transparent, the final result may also be transparent,
-// depending on what the layers paint on top of the base color.
+// This style works like layers in Photoshop: each layer adds to the stack covering the base.
```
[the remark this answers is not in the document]
NoSloppy: I glazed right over that it even said Gimp.
Restored.

styles/layers.h
```diff
@@ -4,21 +4,29 @@
...
-// This style works like layers in gimp or photoshop.
-// In most cases, the layers are expected to be normally transparent effects
-// that turn opaque when then want to paint an effect over the base color.
-// If the base color is opqaque, the final result of this style will also be
-// opaque. If the base color is transparent, the final result may also be transparent,
-// depending on what the layers paint on top of the base color.
+// This style works like layers in Photoshop: each layer adds to the stack covering the base. 
+// The written order of layers in the blade style code puts the base first, with any following layers visually covering it.
+//
+// When a Layers<> is used as the first template argument to StylePtr<>, its BASE color (the first layer)
```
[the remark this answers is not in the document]
NoSloppy: ok how about now?

PR #846, opened 2025-08-21
BCprop: add button press sound

NoSloppy:
All the kids are doing it. 
But, no "release" sound, and only do press sounds when blade is off. 
That's my flavor.

### DIFF
--- a/props/saber_BC_buttons.h
+++ b/props/saber_BC_buttons.h
@@ -2120,16 +2120,15 @@ void DoSavedTwist() {
 
   RefPtr<BufferedWavPlayer> wav_player;
 
-  bool Event(enum BUTTON button, EVENT event) override {
 #ifdef BC_BUTTON_CLICKER
+  bool Event(enum BUTTON button, EVENT event) override {
     // Play clicker only when saber is OFF and a button is PRESSED
-    if (!IsOn() && event == EVENT_PRESSED &&
-        (button == BUTTON_POWER || button == BUTTON_AUX || button == BUTTON_AUX2)) {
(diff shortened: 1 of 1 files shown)


props/saber_BC_buttons.h
```diff
@@ -2117,6 +2120,17 @@ void DoSavedTwist() {
 
   RefPtr<BufferedWavPlayer> wav_player;
 
+  bool Event(enum BUTTON button, EVENT event) override {
+#ifdef BC_BUTTON_CLICKER
```
profezzorn, line 2124: I think you can move this #ifdef to surround the whole function, since you don't need it this function when BC_BUTTON_CLICKER is not defined.

props/saber_BC_buttons.h
```diff
@@ -2117,6 +2120,17 @@ void DoSavedTwist() {
...
+    // Play clicker only when saber is OFF and a button is PRESSED
+    if (!IsOn() && event == EVENT_PRESSED &&
+        (button == BUTTON_POWER || button == BUTTON_AUX || button == BUTTON_AUX2)) {
```
profezzorn, line 2127: I think this is fine, but I'm curious, why not just do it for all buttons?

props/saber_BC_buttons.h
NoSloppy: Indeed. Done.

props/saber_BC_buttons.h
NoSloppy: Ah I see. Why specify?
Done.

PR #849, opened 2025-09-06
Add voice pack version error

NoSloppy:
To more clearly specify error instead of just "error in font directory".


profezzorn:
How much space does this take?

common/errors.h
```diff
@@ -45,17 +50,38 @@ void ProffieOSErrors::font_directory_not_found() {
...
+void ProffieOSErrors::voice_pack_not_found() {
+  SaberBase::DoEffect(EFFECT_VOICE_PACK_NOT_FOUND, 0);
+  PVLOG_ERROR << "** ERROR - Voice pack not found.\n";
+#ifdef ENABLE_AUDIO
+#ifndef DISABLE_TALKIE
+  talkie.Say(talkie_voice_pack_15, 25);
+  talkie.Say(talkie_not_found_15, 15);
+  delay(2000);
+#else
+  beeper.Beep(1.0, 220.00 * 2); // A4 - Voice
```
profezzorn, line 67: It would be nice to listen to how all of these things sound.

NoSloppy: Error in voice pack version and beeps
Voice pack not found and beeps
[vp_errors.wav](https://github.com/user-attachments/files/22194485/vp_errors.wav)

NoSloppy:
> Blocking here means that errors will not show up on displays until later.

I didn't test with display yet, but wouldn't it fire first since DoEffect() is called right away? 
```
void ProffieOSErrors::error_in_voice_pack_version() {
  SaberBase::DoEffect(EFFECT_ERROR_IN_VOICE_PACK_VERSION, 0);
  ```

**Comment by profezzorn** [comment]
> I didn't test with display yet, but wouldn't it fire first since DoEffect() is called right away?

Not sure actually. OLED may work, but color displays will definitely block if the man loop isn't running.

**Comment by NoSloppy** [comment]
> make it so that other sounds are dodged instead, but that can be done separately.

ok but how then? For example, I don't know how SB_Effect() would know that talkie or beeper is busy.

**Comment by NoSloppy** [comment]
How about a delay timer? I didn't measure mem, but it likely takes a little more than just using delay, but maybe it's the proper way to do it.
Added.

**Comment by NoSloppy** [comment]
> I think remove the delays and we can check this in.

Ok removed delays in lieu of upcoming sound dodging code update.

File: props/saber_BC_buttons.h [hunk 2]
```diff
@@ -1453,11 +1453,13 @@ class SaberBCButtons : public PROP_INHERIT_PREFIX PropBase {
   SaberBCButtons() : PropBase() {}
   const char* name() override { return "SaberBCButtons"; }
 
-#ifndef MENU_SPEC_TEMPLATE
   void Setup() override{
+#ifndef MENU_SPEC_TEMPLATE
     MKSPEC<BCMenuSpec>::SoundLibrary::init();
-  }
+#else
+    SoundLibrary::init();
```
profezzorn, line 1460: Doesn't your prop use V2 sounds?

sound/sound_library.h
```diff
@@ -72,11 +72,17 @@ class SoundQueueSingleton : public SoundQueue<16>, Looper, SaberBase {
...
     }
     if (found_version < required_version_) {
-      PVLOG_ERROR << "ERROR - Sound library version " << required_version_ << " required.\n";
-      ProffieOSErrors::error_in_font_directory(); // Make new error for voice pack?
+      if (SFX_mnum) {
+        PVLOG_ERROR << "** A version " << required_version_ << " voice pack is required to use the selected prop file.\n";
+        ProffieOSErrors::error_in_voice_pack_version();
+      } else {
+        PVLOG_ERROR << "** A voice pack is required to use selected prop file.\n";
+        PVLOG_ERROR << "** Add sounds to SD card and append 'common' to preset's font search path.\n";
```
profezzorn, line 83: Two lines of errors seems like a lot. (and costs a few extra bytes of memory)
(Three if you count the one in ProffieOSErrors::voice_pack_not_fount.)
Maybe we should just add a URL to the POD where we can have a full explanation of the error and what to do about it?

sound/effect.h
```diff
@@ -610,12 +610,12 @@ class Effect {
 #endif
 
 #ifdef ENABLE_SD
+    if (!LSFS::Begin()) ProffieOSErrors::sd_card_not_found();
```
NoSloppy, line 613: The No SD Card error doesn't show because it only happens at boot (right?).
Should this go here so it gets reported at preset change too?

sound/sound_library.h
profezzorn: I don't ilke using TinyURL, since it makes it hard to know where you're going.
However, we could potentially set up redirects on the pod that have shorter paths.
Something like:
https://pod.hubbe.net/r/eifd  (r stands for redirects, eifd = error in font directory)

I'm not entirely sure if you can do redirect to anchors easily though. Might require javascript.
(Browsers tends to treat URLs and anchors as separate things.)

> Too bad serial monitor can't use hyperlinks :(

We'll have to make sure to add that for the master board. :)

sound/effect.h
profezzorn: Good question, it might be useful, might be confusing if the SD card is busy (mounted computer) though.

sound/sound_library.h
profezzorn: Did some testing, this seems to create a shortcut that works:
https://github.com/profezzorn/ProffieOSDocs/blob/main/r/eifd.md

PR #853, opened 2025-09-17
Update NO_REPEAT_RANDOM to use Sample Without Replacement (2nd Try)

pedker:
This is a second try at [this PR](https://github.com/profezzorn/ProffieOS/pull/659), attempting to address its raised issues. 

The storage used to manage the Sample-Without-Replacement algorithm is now a one-dimensional boolean array on each Effect, created statically of size `NO_REPEAT_RANDOM_BUFFER_SIZE_BITS`. Each bit of the array is used to manage an individual Sound File, and the array's size determines the maximum number of Sound Files each Effect can support using SWR. If an Effect has more Sound Files than available memory, then the Effect will default back to using the existing `NO_REPEAT_RANDOM` logic.

To meet the requirement of using less than one byte of memory per File, the value of `NO_REPEAT_RANDOM_BUFFER_SIZE_BITS` should be tuned relative to the ratio of Files to Effects used by Fonts on average. For example, setting it to 16 would only require the ratio to be 2:1, A value of 24 would require 3:1, and a value of 32 would require 4:1.

### DIFF
--- a/sound/effect.h
+++ b/sound/effect.h
@@ -5,6 +5,14 @@
 #include "../common/atomic.h"
 #include "../common/file_reader.h"
 
+#ifndef NO_REPEAT_RANDOM_BUFFER_SIZE_BITS
+#define NO_REPEAT_RANDOM_BUFFER_SIZE_BITS 24
+#endif
+
+#if NO_REPEAT_RANDOM_BUFFER_SIZE_BITS > 256
+#define NO_REPEAT_RANDOM_BUFFER_SIZE_BITS 256
+#endif
+
 class Effect;
 Effect* all_effects = NULL;
 
@@ -175,6 +183,11 @@ class Effect {
     paired_ = false;
 #ifdef KILL_OLD_PLAYERS
     killable_ = false;
+#endif
+#ifdef NO_REPEAT_RANDOM
+    for (size_t i = 0; i < NO_REPEAT_RANDOM_BUFFER_SIZE_BITS; ++i) available_[i] = available_mark_;
+    last_ = -1;
+    last_subid_ = -1;
 #endif
   }
   static int altnum(const char* s) {
@@ -379,6 +392,10 @@ class Effect {
   }
 
 #ifdef NO_REPEAT_RANDOM
+  // Storage for random sampling without replacement.
+  bool available_[NO_REPEAT_RANDOM_BUFFER_SIZE_BITS];
+  bool available_mark_ = false;
+
   int16_t last_ = -1;
   int16_t last_subid_ = -1;
 
@@ -406,6 +423,32 @@ class Effect {
 
   int random_subid(int filenum) {
     if (!sub_files_) return 0;
+#ifdef NO_REPEAT_RANDOM
+    size_t total_files = files_found() * sub_files_;
+    if (total_files > 2 && total_files <= NO_REPEAT_RANDOM_BUFFER_SIZE_BITS) {
+      const uint8_t start = filenum * files_found();
+      const uint8_t end = start + sub_files_ - 1;
+      
+      uint8_t n = 0;
+      for (uint8_t i = start; i <= end; ++i) n += available_[i] == available_mark_ && (last_ != filenum || last_subid_ != i % sub_files_);
+      if (!n) {
+        for (uint8_t i = start; i <= end; ++i) available_[i] = available_mark_;
+        n = sub_files_;
+      }
+
+      uint8_t nth_ret = RANDOMIZE(n, -1);
+      for (uint8_t i = start; i <= end; ++i) {
+        if (available_[i] != available_mark_ || (last_ == filenum && last_subid_ == i % sub_files_)) continue;
+        if (!nth_ret) {
+          available_[i] = !available_mark_;
+          last_ = i / sub_files_;
+          last_subid_ = i % sub_files_;
+          return last_subid_;
+        }
+        nth_ret--;
+      }
+    }
+#endif
     int ret = RANDOMIZE(sub_files_, last_ == filenum ? last_subid_ : -1);
 #ifdef NO_REPEAT_RANDOM
     last_subid_ = ret;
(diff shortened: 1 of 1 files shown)


sound/effect.h
```diff
@@ -175,6 +183,11 @@ class Effect {
     paired_ = false;
 #ifdef KILL_OLD_PLAYERS
     killable_ = false;
+#endif
+#ifdef NO_REPEAT_RANDOM
```
profezzorn, line 187: If we're going to have this code (which I'm not entirely convinced it is a good idea.) then we should use a new define to enable it. Maybe the NO_REPEAT_RANDOM_BUFFER_SIZE_BITS define?

sound/effect.h
```diff
@@ -379,6 +392,10 @@ class Effect {
   }
 
 #ifdef NO_REPEAT_RANDOM
+  // Storage for random sampling without replacement.
+  bool available_[NO_REPEAT_RANDOM_BUFFER_SIZE_BITS];
```
profezzorn, line 396: An array of bool does not take one bit per bool.
In most cases, a bool takes one byte.

sound/effect.h
```diff
@@ -379,6 +392,10 @@ class Effect {
...
+  bool available_mark_ = false;
```
profezzorn, line 397: This needs a better name, or a comment that explains what it is for.

sound/effect.h
```diff
@@ -430,6 +473,35 @@ class Effect {
...
       n = std::min<int>(SaberBase::sound_number, num_files - 1);
     } else {
+#ifdef NO_REPEAT_RANDOM
+      size_t total_files = files_found() * number_of_subfiles();
+      if (total_files > 2 && total_files <= NO_REPEAT_RANDOM_BUFFER_SIZE_BITS) {
+        uint8_t n = 0;
+        for (uint8_t i = 0; i < total_files; ++i) {
+          const int16_t file = sub_files_ ? i / sub_files_ : i;
+          const int16_t subfile = sub_files_ ? i % sub_files_ : 0;
+          n += available_[i] == available_mark_ && (last_ != file || (sub_files_ && last_subid_ != subfile));
```
profezzorn, line 483: Use an if statement here, it will make the code easier to read.

sound/effect.h
```diff
@@ -430,6 +473,35 @@ class Effect {
...
+        }
+        if (!n) {
+          available_mark_ = !available_mark_;
```
profezzorn, line 486: Using available_mark_ this way doesn't  save anything if you have to iterate over the whole array to check how many elements are available anyways.

sound/effect.h
pedker: Ah, crap! I incorrectly assumed that a boolean took up one bit in memory. I will have to look into the feasibility of refactoring to instead use a single integer with bitmasking.

This means that `available_` will change to be either of type `uint16_t` or `uint32_t`, and the size will no longer be configurable. The current version of the define macro will be renamed/reused as the enable define, mentioned above. 32 bits is close to too much memory, considering average file counts. 16 bits is way below the budget, but a max of 16 files per Effect may feel a bit constraining. I am leaning towards 16 bits and will procced that way unless told otherwise.

sound/effect.h
pedker: It lowers the worst case from 3N to 2N, since the alternative is to iterate over the array one more time to mark everything as available. I agree though, it's not really much of a save since it's still all O(N). The refactor will scrap this anyway, since if `available_` is becoming an integer I can just set it back to 0 or max

profezzorn: Marking everything as available only has to be done every Nth time though, so the average time spent ends up being constant.

pedker:
Successfully converted this to use a `BitSet`, though I had to add a few functions to the `BitSet` class, mainly to support some subset-based operations. Also, I set the memory size to 32 because `BitSets` denominate by increments of 32-bit ints, so setting the size to 16 would just waste the other 16 bits.

Something else to note is that I have not been able to test the `random_subid()` portion of this code, because I cannot get that function to fire under normal circumstances. It's supposed to be called by `GetFollowing()`...

common/arg_parser.h
```diff
@@ -141,13 +141,26 @@ class BitSet {
...
+    }
+  }
   size_t popcount() const {
     size_t ret = 0;
     for (size_t i = 0; i < NELEM(bits_); i++) {
       ret += __builtin_popcount(bits_[i]);
     }
     return ret;
   }
+  size_t popcount_subset(size_t first, size_t last) const {
```
profezzorn, line 157: I'm wondering if it would be easier (and faster) to do all this subset stuff with ands and ors.

Basically something like:

```cpp
 Bitset subset = ~(Bitset::All() << last) & (Bitset::All() << first);
 size_t popcount_subset = (subset & bitset).popcount();
```

sound/effect.h
```diff
@@ -430,6 +484,28 @@ class Effect {
...
       n = std::min<int>(SaberBase::sound_number, num_files - 1);
     } else {
+#ifdef NO_REPEAT_RANDOM
+#ifdef RANDOM_SAMPLE_WITHOUT_REPLACEMENT
+      const size_t total_files = files_found() * number_of_subfiles();
+      if (total_files > 2 && total_files <= SAMPLE_WITHOUT_REPLACEMENT_BUFFER_SIZE_BITS) {
+        uint8_t n = available_.popcount();
+        if (!n) {
+          available_.fill(total_files);
+          n = available_.popcount();
```
profezzorn, line 494: n = total_files; would be faster

common/arg_parser.h
```diff
@@ -141,13 +141,26 @@ class BitSet {
       bits_[i] = 0;
     }
   }
+  void fill(size_t bit = SIZE) {
```
profezzorn, line 144: This should be a static function that returns a bitset. It also needs a comment to explain what it does.

common/arg_parser.h
```diff
@@ -141,13 +141,26 @@ class BitSet {
...
+    for (size_t i = 0; i < NELEM(bits_); i++) {
```
profezzorn, line 145: The loop should only iterate over all bits_[] where it can assign 0xFFFFFFFF, then assign the final value after the loop.

pedker:
Identified a crashing issue and put up a fix. Unfortunately, before I could verify the fix and finish testing, I accidentally ripped the USB port right off of the ProffieBoard of my saber. So, for now, my ability to work on this has been crippled. I'm not sure how long it will take to remedy, as I may have to contact support, try to DIY fix/replace the board on my own, or wait until I can afford to pay for a replacement solution.

common/arg_parser.h
```diff
@@ -172,15 +197,76 @@ class BitSet {
...
+    for (size_t i = 0; i < NELEM(bits_); i++) bits_[i] = get32(i * 32 + bits);
+  }
+  BitSet<SIZE> operator>>(int bits) const {
+    BitSet<SIZE> ret;
+    for (size_t i = 0; i < NELEM(bits_); i++) ret.bits_[i] = get32(i * 32 + bits);
+    return ret;
+  }
+  void operator<<=(int bits) {
+    if (!bits) return;
+    const size_t n = NELEM(bits_);
```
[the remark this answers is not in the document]
pedker: Fair, I was going to but was afraid of moving away from using `size_t`

PR #855, opened 2025-09-23
Add set_preset_fast to serial cmds

NoSloppy:
Primarily for Forcesync app use, this bypasses font.wav from playing while on


profezzorn:
Maye we should just change set_preset?

PR #857, opened 2025-09-24
Fix errors not playing wavs

NoSloppy:
If we wait until after the call to SaberBase::Link(this); in Activate() to call the errors, it plays the wavs.

### DIFF
--- a/common/errors.h
+++ b/common/errors.h
@@ -2,6 +2,9 @@
 #ifndef COMMON_ERROR_H_DECLARED
 #define COMMON_ERROR_H_DECLARED
 
+extern bool font_dir_missing_;
+extern bool error_in_font_dir_;
+
 #ifdef ENABLE_AUDIO
 void DodgeSound(uint32_t millis);
 #else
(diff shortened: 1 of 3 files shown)


NoSloppy:
> I think the easier (and maybe better) fix is to change PropBase::chdir..... Just move the everything inside the first "ifdef ENDBLE_AUDIO" block down into the beginning of the second "ifdef ENABLE_AUDIO" block. That way Deactivate/Activate happens right next to each other, and errors should work.
> 
> Doesn't solve the overlap problem though, we might need SoundQueue for that.

Hmm. That doesn't work. 
At boot, talkie still says "error in font directory", and font.wav plays, and saber can be turned on/off ok.
Doing Next Preset (where I have a missing font) shows it's playing  e_fnt_nf.wav but it doesn't, then shows scanning the next font in the path, but freezes the board there.

I tried commenting out the stop wav playing bit too, but no change.

NoSloppy:
> Did we fix this? I don't remember?

I left it hanging in my local copy. The initial PR submission works. But then you said to try to change PropBase::chdir which I tried and replied, then you also mentioned soundQueue... I never got up to that part if I recall.
I'd love to when I get some time next week, unless you can.

profezzorn:
I figured out why it crashed, fix that and made some changes to chdir to make it work.
Please give it a try and see if it works for you too.

PR #887, opened 2026-03-16
Volume harmonization improvement

Sabersense-CC:
Harmonized button controls between 1 and 2 button volume menu control, intended for 2 button hilts in which you cannot press both buttons at the same time. Can't believe I didn't spot this one sooner - better functionality and less code! :)


profezzorn:
This PR just changes a typo, the title doesn't make much sense for that...

Sabersense-CC:
The "Comment tweak for clarity" just changes a typo, yes, but the main PR does a little bit more - it adds the one button control of:
hold-button-and-clash-to-enter-volume-menu
to two button systems to allow for 2 button hilts with rocker switch systems that only allow either/or button presses but not both at once. So there is a tangible, albeit small, material change.

PR #891, opened 2026-08-02
Better error wav use

NoSloppy:
Overlapping WAV files are unintelligible. Use SOUNDQ instead for them.
Also fixes a freezing issue caused by checking sound length before it’s loaded in PlayErrorMessage().

### DIFF
--- a/common/delay_timer.h
+++ b/common/delay_timer.h
@@ -1,0 +1,26 @@
+#ifndef DELAY_TIMER_H
+#define DELAY_TIMER_H
+// Returns a reference to the shared deadline timestamp (milliseconds).
+inline uint32_t& delay_until_ms() {
+  static uint32_t ts = 0;
+  return ts;
+}
+// Set the timer to expire `duration_ms` from now, overwriting any
+// existing deadline.  Use when only one error source is active.
+inline void StartDelayTimer(uint32_t duration_ms) {
+  delay_until_ms() = millis() + duration_ms;
+}
+// Extend the deadline by `duration_ms` from wherever it currently sits
+// (or from now if it has already expired).  Use when multiple error
+// sources may fire in quick succession so that their delays stack rather
+// than reset.
+inline void AppendToDelayTimer(uint32_t duration_ms) {
+  uint32_t now = millis();
+  if (delay_until_ms() < now) delay_until_ms() = now;
+  delay_until_ms() += duration_ms;
+}
+// Returns true while the deadline is still in the future.
+inline bool DelayTimerActive() {
+  return millis() < delay_until_ms();
+}
+#endif  // DELAY_TIMER_H
--- a/common/delay_timer.h
+++ b/common/delay_timer.h
@@ -1,15 +1,18 @@
 #ifndef DELAY_TIMER_H
 #define DELAY_TIMER_H
+
 // Returns a reference to the shared deadline timestamp (milliseconds).
 inline uint32_t& delay_until_ms() {
   static uint32_t ts = 0;
   return ts;
 }
+
 // Set the timer to expire `duration_ms` from now, overwriting any
 // existing deadline.  Use when only one error source is active.
 inline void StartDelayTimer(uint32_t duration_ms) {
   delay_until_ms() = millis() + duration_ms;
 }
+
 // Extend the deadline by `duration_ms` from wherever it currently sits
 // (or from now if it has already expired).  Use when multiple error
 // sources may fire in quick succession so that their delays stack rather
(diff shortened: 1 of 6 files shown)


sound/hybrid_font.h
```diff
@@ -612,8 +613,23 @@ class HybridFont : public SaberBase, public Looper {
       case EFFECT_FORCE: PlayCommon(&SFX_force); return;
       case EFFECT_BLAST: Play(&SFX_blaster, &SFX_blst); return;
       case EFFECT_QUOTE: PlayCommon(&SFX_quote); return;
-      case EFFECT_BOOT: PlayPolyphonic(&SFX_boot); return;
-      case EFFECT_NEWFONT: SB_NewFont(); return;
+      case EFFECT_BOOT:
```
profezzorn, line 616: Maybe we should just play BOOT on the sound queue?

sound/sound_queue.h
```diff
@@ -33,6 +33,26 @@ class SoundToPlayInCurrentDir : public SoundToPlayBase {
   const char* filename_;
 };
 
+
+// Like SoundToPlayInCurrentDir but also searches the "errors" directory
+// as a fallback.  Used by PlayErrorMessage()
+class SoundToPlayErrorFile : public SoundToPlayBase {
+public:
+  SoundToPlayErrorFile(const char* filename) : filename_(filename) {}
+  bool Play(BufferedWavPlayer* player) override {
```
profezzorn, line 41: This should still do dodging.

common/delay_timer.h
```diff
@@ -0,0 +1,29 @@
+#ifndef DELAY_TIMER_H
+#define DELAY_TIMER_H
+
+// Returns a reference to the shared deadline timestamp (milliseconds).
+inline uint32_t& delay_until_ms() {
```
profezzorn: These functions should be a class.

sound/hybrid_font.h
NoSloppy: yep.

sound/sound_queue.h
NoSloppy: what for? If it's for Talkie, that's handled in this latest update as far as delaying boot/font.wavs.

common/delay_timer.h
NoSloppy: done

NoSloppy:
Ok I tested all this with and without wavs for all error messages except for "SD card not found".
Works well. No overlap, no dodging needed, delays all seem correct. I did not try different length wav files, but that should work fine as duration is accounted for.

sound/sound_queue.h
profezzorn: Depending on the prop, there may be lots of other sounds playing.

profezzorn:
I'm trying to decide if this is a good idea or not.
Delaying playing things breaks the Event2 contract (the idea that you can get the length of the sound in Event2).

Maybe it's not a major concern since we're in an error state already.

This is the reason I originally went with dodging instead of delaying; it fits better with the existing framework.

Not saying we shouldn't do this, just saying that I need some time to think through the implications.

NoSloppy:
> I'm trying to decide if this is a good idea or not. Delaying playing things breaks the Event2 contract (the idea that you can get the length of the sound in Event2).
> 
> Maybe it's not a major concern since we're in an error state already.
> 
> This is the reason I originally went with dodging instead of delaying; it fits better with the existing framework.
> 
> Not saying we shouldn't do this, just saying that I need some time to think through the implications.

Well, the goal was intelligibility of the error messages so that a user can understand what needs to be fixed. Knowing which font they're having an issue with is helpful in my opinion, so that's why I would lean toward the way it is now, announcing the font.wav after the errors.
Maybe the fix that keeps the contract intact is to delay the whole event, not just the sound? If we defer EFFECT_BOOT/EFFECT_NEWFONT until the error queue drains, OLED display stuff and audio would stay in sync and sound_length would be correct when it fires. Would that address your concern? (I also need to drop the sound_length = 0.001 hack I'm using to suppress Talkie and use a separate flag.)
As far as Play() still dodging, maybe we just do it one level up. Rather than adding it inside Play(), PollSoundQueue() could keep re-arming a short DodgeSound() every poll while the queue is busy and set_dodge(false) on the queue's own player. That should duck _everything_ under any queued speech (errors, menu prompts, HEV, colors — the "lots of other sounds" case). Plus that should pick up any sounds that might start mid-speech (where the one-shot UnaDodger::start() can't), and it wouldn't need player->length(), which is what caused a freeze in the first place. 
Then we also put the DodgeSound() calls back in the Talkie/beeper branches of errors.h alongside the delay timer.
Does that better match what you had in mind?

NoSloppy:
I got help.

NoSloppy:
I made a Ver2 of this PR with the different approach instead. Maybe it's more to your liking.
The description there is overly thorough, so nothing to say here.

